PR #2, opened 2017-06-17
Style compiler

wodeni:
**In progress**

This branch will implement the compiler for our new Style language design. 

### Status

- [170617]: first version that compiles and check global blocks and assignment statements correctly.


src/StyAst.hs
```diff
@@ -0,0 +1,176 @@
...
+-- Style grammar (relies on the Substance grammar, specifically SubObj)
+
+data SubType
+    = Set
+    | Pt
+    | Map
+    deriving (Show)
+
+data StyProg
+    = Block Block
```
hypotext: `Program Block`?

src/StyAst.hs
```diff
@@ -0,0 +1,176 @@
...
+    deriving (Show)
+
+data Block
+    = BlockSeq [Block]
+    | GlobalBlock Stmt
+    | TypeBlock SubType Stmt
+    | ObjBlock String Stmt
+    | ConstraintBlock Stmt
```
hypotext: `ConstraintBlock [Stmt]`?

src/StyAst.hs
```diff
@@ -0,0 +1,176 @@
...
+    deriving (Show)
+
+data Stmt
+    = StmtSeq [Stmt]
```
hypotext: remove recursive reference?

hypotext: depends on whether we want arbitrary nesting

hypotext:
Merged `parse-style` branch separately. Closing this PR so we can make a new one for `style-compiler` only.

PR #3, opened 2017-06-18
Style compiler

wodeni:
**In progress**

This branch will implement the compiler for our new Style language design. 

### Status

- [170617]: first version that compiles and check global blocks and assignment statements correctly.
- [170619]: the parser now have the same functionalities as the older version
  - Rendering the continuous map example:
  170619-parser


hypotext:
Could you move penroseDB to a separate repo in the `penrose` organization? I added you as a collaborator.

PR #6, opened 2017-07-15
New Substance for functions

wodeni:
*In progress*

### TODOs

- [ ] alternative viz: Cartesian Coordinate System
   - [ ] simplified example of the new viz 
   - [ ] add dashed lines, bezier curves to the system
   - [ ] update in Style to support the new viz
- [x] wikipedia-style viz
- [x] java program to run allloy
- [x] Parser for the new Substance
- [x] Parsers for Alloy support
    - [x] Substance to Alloy translater  
    - [x] Alloy to Substance translater

### The new Substance language

- This version has backward-compatibility with the former Substance
- One more type `Value` is added, representing something like `f(a) = b`. One can also write `Value f a b`
- `Set A = { a }` would be desugared into `Set A; Point a; PointIn a a`. This might not make sense in general. We can probably make `Point` the default type if none is specified and let a `Set` contain anything.
- Comma separated declaration: `Set A, B, C` is fine
- Semantic checking: Type checking for all Substance constraints, assignment of function values, application of definitions
- Grammar for FOL: similar to Coq. I only adopted Alloy's vertical bars as separators of qualifiers
- Example program:
```
f: A -> B
Set A = { a, b, c, d }
Set B = { w, x, y, z }

Definition Surjection(Map f, Set A, Set B):
    forall b : B | exists a : A | f(a) = b
Definition Injection(Map f, Set A):
    forall a1, a2 : A | f(a1) = f(a2) implies a1 = a2
Definition Function(Map f, Set X, Set Y):
    forall x : X | exists y : Y | f(x) = y
Definition DuplicateVal(Map f, Set X):
    forall x : X | exists x1 : X | x != x1 /\ f(x) = f(x1)
Surjection(f, A, B)
Injection(f, A)
```

#### Generating instances using Alloy

- Instance generation is triggered by definition applications such as `Surjection(f, A, B)`. Therefore, all other definitions such as `Function` are ignored.
- After the compilation of the Substance program, the generated AST will be passed to a Substance-to-Alloy parser, which outputs an Alloy AST (I wrote a simplified grammar for Alloy, and a pretty printer for the AST). 
- For example, the program in the previous section will be translated to:
```
sig A {
    f : B
}
sig B {
}
one sig a extends A{ }
one sig b extends A{ }
one sig c extends A{ }
one sig d extends A{ }
one sig w extends B{ }
one sig x extends B{ }
one sig y extends B{ }
one sig z extends B{ }
fact {
  all b : B | some a : A | a.f = b
  all a1,a2 : A | a1.f = a2.f implies a1 = a2
}
pred show() { }
run show
```
- We then pass the Alloy program string to a Java program that runs Alloy Analyzer, which outputs something like this:
```
{a$0->x$0, b$0->w$0, c$0->y$0, d$0->z$0}
```
- We translate these tuples to `Value` typed objects and insert that to the list of Substance objects. So we end up with the following Substance objects:
```
Set "A"
Point "d"
Point "c"
Point "b"
Point "a"
Set "B"
Point "z"
Point "y"
Point "x"
Point "w"
Map "f" "A" "B"
Value "f" "a" "x"
Value "f" "b" "w"
Value "f" "c" "y"
Value "f" "d" "z"
```
- The Style program can choose to render `Value` or not. Here are two different styles of the same Substance:
image
image

### Examples of other definitions
- All values in the domain are mapped to the same point:
```
f: A -> B
Set A = { a, b, c, d }
Set B = { w, x, y, z }
Definition SameValue(Map g, Set S, Point p):
    forall a : S | g(a) = p
SameValue(f, A, x)
```
image

- **forall x in X, exists x’ in X : f(x) = f(x’)** (came from Katherine's comment during weekly meetings)
```
f: A -> B
Set A = { a, b, c, d }
Set B = { w, x }
Definition DuplicateValTrivial(Map f, Set X):
    forall x : X | exists x1 : X | f(x) = f(x1)
Definition DuplicateVal(Map f, Set X):
    forall x : X | exists x1 : X | x != x1 /\ f(x) = f(x1)
Definition Surjection(Map f, Set A, Set B):
    forall b : B | exists a : A | f(a) = b
DuplicateVal(f, A)
-- DuplicateValTrivial(f, A)
Surjection(f, A, B)
```
  - Seems that Alloy will just map all values to the domain to the same point the the codomain if we don't make sure the function is surjective. That's why I added `Surjection(f, A, B)`.
  - Output: 
image
  - If we change `DuplicateVal` to `DuplicateValTrivial`, we will have: 
image

### The alternative viz

- I had a couple ones in mind:
   - commutative diagram with arrow styling: see a discussion [here](https://math.stackexchange.com/questions/46678/what-are-usual-notations-for-surjective-injective-and-bijective-functions)
image
   - Slightly different ellipse-arrow style: this one emphasizes the codomain-range relationship. So this is injection, a surjection will be just two ellipses (yellow and blue ones merge). Bijection is tricky, We can use actually points and arrows to denote that. Is there any other way?
image
    - Keenan's continuous function diagram:
image
    - There are questions that I can think of:
        - x, y-axis: is this mathematical? I guess yes, but it certainly sounds more like a `View` thing here
        - bezier curve support: to draw these functions, we definitely need them, and it is possible to draw them in the case that the user provided specific points on the functions? For example, can we draw a bezier such that it passes `Point a, b`?
        - This is getting closer to the graphical calculator world. We should probably do a couple examples in TikZ/MatLab or something before we proceed.

### The Ellipse-arrow layout

- I reused the original Substance syntax, so that we can hack together the viz before having the full parser for the new Substance, which is still in progress.
- I tried to make the Style program as weight-agnostic as possible. For now, all examples (continuous map, 3 viz for subsets, and other simpler ones) work except for that tree layout
   - The tree layouts depends on the fact that `sameX` has lesser weight than `repel` because we do not want the labels of subsets to be actually right below the superset's label. So I had to decrease the weight to get the correct look. If we keep the original weight, it looks like this:
image

- Does that mean we will have to consider a good way to expose the weight param to the user soon?

### Other

- Refactoring the code base:
    - Added a list of `Floating` to each obj/constraint function, so we can potentially pass numerical arguments to them in the future
    - Used multi-parameter type classes so that `Obj'` type also implement `Located`
    - BBox retrieval: now only for labels - upon startup of the web frontend, the JS program will send back bounding boxes for all labels and `Runtime` will update them in the `Obj` list
        - Might be good to do it for all objects, so we can have them storing their bboxes, in case we want to use them in the future (a good functionality to have is containment/intersection tests implemented in terms of bboxes, as a fallback option.
    - Renamed the modules so we have `Substance` and `Style` now, instead of `StyAst` and `Compiler`

### Preview

- Wiki-style viz
170715-surjective


hypotext:
💯

PR #52, opened 2018-02-15
Refactoring `procBlock`

wodeni:
Refactored `procBlock` and some aliasing for readability of `Style`
* added underscore to all internal ids for labels
* modified tree.sty due to changes in label definition
* separated out selector logic.
* different `procBlock` cases
    * case 1: just one `global` selector -> process objs and constrs
once
    * case 2: just one selector -> the original logic
    * case 3: a list of comma separated selectors -> objs and constrs only and compute valid combinations among the selctors

### DIFF
--- a/src/Style.hs
+++ b/src/Style.hs
@@ -8,9 +8,11 @@ module Style where
 import Shapes
 import Utils
 import Control.Monad (void)
+import Data.Function (on)
 import Data.Either (partitionEithers)
 import Data.Either.Extra (fromLeft)
 import Data.Maybe (fromMaybe)
+import Data.List (nubBy, nub)
 import Text.Megaparsec
 import Text.Megaparsec.Char
 import Text.Megaparsec.Expr
@@ -166,11 +168,11 @@ assignStmt = do
 -- | modeled off of objFn
 parseComp :: Parser Expr
 parseComp = do
-  fname <- identifier
-  lparen
-  params <- expr `sepBy` comma
-  rparen
-  return (CompArgs fname params)
+    fname <- identifier
+    lparen
+    params <- expr `sepBy` comma
+    rparen
+    return (CompArgs fname params)
 
 -- | objective function call
 objFn :: Parser Stmt
@@ -364,7 +366,8 @@ matchWithAll :: StyDict -> [Selector] -> [[(VarMap, StySpec)]]
 matchWithAll dict selectors = map (matchWith dict) selectors
 
 -- | Given a selector and a Style spec that matches with it, returns a map from placeholder ids to actual matched ids.
--- TODO: the underlying logic is the same as `matched`. Think of a cleaner way to both (1) check if there is a match and (2) collect a list of `VarMap`s
+-- TODO: the underlying logic is the same as `matched`. Think of a cleaner
+-- way to both (1) check if there is a match and (2) collect a list of `VarMap`s
 getVarMap :: Selector -> StySpec -> VarMap
 getVarMap sel spec = foldl add M.empty patternNamePairs
     where
@@ -375,9 +378,9 @@ getVarMap sel spec = foldl add M.empty patternNamePairs
             RawID _    -> dict
             WildCard wc -> M.insert wc name dict
 
-
--- Gather a list of objective or constraint functions by mapping over mergedMaps
--- genFns :: (VarMap -> [t] -> Stmt -> [t]) -> VarMap -> [t]
+-- Given either 'procConstrFn' or 'procObjFn' and its arguments, returns
+-- a #map@pable function that can get @map@ped onto a list of variable maps
+genFns :: (VarMap -> [t] -> Stmt -> [t]) -> [Stmt] -> VarMap -> [t]
 genFns f stmts varmap = foldl (f varmap) [] stmts
 
 -- | 'procBlock' is called by 'getDictAndFns'. 'getDictAndFns' would fold this function on a list of blocks, a.k.a. a Style program, and accumulate objective/constraint functions, and a dictionary of geometries to be rendered.
@@ -423,25 +426,22 @@ procBlock (dict, objFns, constrFns) (selectors, stmts) =
         newConstrFns  = concatMap (genFns procConstrFn stmts) mergedMaps
     in (dict, objFns ++ newObjFns, constrFns ++ newConstrFns)
     where
-        validMap :: [(Name, Name)] -> Bool
-        validMap tuples = and . fst $ foldl
-            (\(l, m) (x, y) -> case M.lookup x m of
-                Nothing -> (True:l, M.insert x y m)
-                Just y' -> ((y == y') : l, m))
-            ([], M.empty) tuples
+        -- makes sure we don't bind same name with multiple Substance ids
+        -- in a list of comma-separated selectors
+        isOneToOne :: [VarMap] -> Bool
+        isOneToOne validMaps varmaps =
+            let flatMap      = nub $ concatMap M.toAscList varmaps
+                bijection    = bijectify flatMap
+            in  flatMap == bijection
+            where bijectify = nubBy ((==) `on` snd) . nubBy ((==) `on` fst)
 
-        noDup :: [(M.Map Name Name)] -> Bool
-        noDup varmaps = validMap $ concatMap M.toAscList varmaps
-
-        -- Combination of all selected (spec. varmap) (???)
-        allCombs :: [[VarMap]] -> [[VarMap]]
-        allCombs varmaps = filter
+        -- allCombinationsOf :: [[VarMap]] -> [[VarMap]]
+        allCombinationsOf varmaps = filter
             (\x -> length x == length varmaps) $
             cartesianProduct varmaps
-
-        mergeMaps :: [[VarMap]] -> [VarMap]
-        mergeMaps varmaps =  map M.unions (filter noDup $ allCombs varmaps)
-
+        -- mergeMaps :: [[VarMap]] -> [VarMap]
+        mergeMaps varmaps =  map M.unions (filter isOneToOne $
+            allCombinationsOf varmaps)
 
 
 -- | Called repeatedly by 'procBlock', 'procConstrFn' would look up and generate constraint functions if the input is a constraint function call. It ignores all other inputs
@@ -491,7 +491,7 @@ lookupVarMap s varMap = case M.lookup s varMap of
 -- | Resolve a Style expression, which could be operations among expressions such as a chained dot-access for an attribute through a couple of layers of indirection (TODO: hackiest part of the compiler, rewrite this)
 -- | An expression can be either a string (variable name) or float (literal)? Not sure
 procExpr :: (Autofloat a) =>
-    VarMap -> Expr -> Either String a -- TODO: revert to Either String a
+    VarMap -> Expr -> Either String a
 procExpr d (Id s)  = {-traceStack ("PROC 1 | " ++ s ++ " | " ++ show d) $ -}Left $ lookupVarMap s d
 -- FIXME: properly resolve access by doing lookups
 procExpr d (BinOp Access (Id i) (Id "label"))  = {-traceStack "PROC 2" $-} Left $ labelName $ lookupVarMap i d
@@ -503,7 +503,7 @@ procExpr v e  = error ("expr: argument unsupported! v: " ++ show v ++ " | e: " +
 -- Unsupported: Cons, and Comp seems to be (hackily) handled in procAssign
 
 -- FIXME: this (?) is incorrect, we should resolve the variables earlier
-addSpec :: VarMap -> M.Map String Expr -> Stmt -> M.Map String Expr
+addSpec :: VarMap -> Properties -> Stmt -> Properties
 addSpec _ dict (Assign s e@(Cons NoShape _)) = M.insert s (Id "None") dict
 addSpec _ dict (Assign s e@(Cons Auto _)) = M.insert s (Id "Auto") dict
 -- FIXME: wrap fromleft inside a function!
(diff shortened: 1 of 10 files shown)


src/Style.hs
```diff
@@ -283,12 +269,40 @@ attribute = identifier -- TODO: Naming convention - same as identifiers?
...
+-- Translates from a Style AST to:
+-- * a list of geometric primitives
+-- * a list of constraints
+-- * a list of objectives
+
+-- Type aliases for readability in this section
+-- | 'StyContext' maintains the current output of the translater
+type StyContext a =
+    (StyDict, -- | dictionary mapping Substance ID to Style objects
+    [(ObjFnOn a, Weight a, [Name], [a])],    -- | List of objective functions
```
hypotext: Why not use `ObjFnInfo a` and `ConstrFnInfo a`?

wodeni: Oh true. Pushing a quick fix now. Any other suggestion?

src/Style.hs
```diff
@@ -300,130 +314,145 @@ initSpec = StySpec { spType = C.PointT,
...
+            let newSpec = foldl (procAssign varmap) spec statements
+            in M.insert (spId newSpec) newSpec dct
+
+-- Case 3: multiple selectors, which in the current Style languages means
+-- we would take a subset of the cartesian product among the selectors.
+-- TEMP: the system does not support assignment statements in these blocks,
+-- because it is not clear that which Substance object the shapes associate with
+procBlock (dict, objFns, constrFns) (selectors, stmts) =
+    let mapsAndSpecs  = dict `matchWithAll` selectors
+        mergedMaps    = mergeMaps $ map (map fst) $ mapsAndSpecs
```
hypotext: $ not needed

src/Style.hs
```diff
@@ -300,130 +314,145 @@ initSpec = StySpec { spType = C.PointT,
...
+        newObjFns     = concatMap (procObjFn identityMap [])    stmts
+        newConstrFns  = concatMap (procConstrFn identityMap []) stmts
+    in (dict, objFns ++ newObjFns, constrFns ++ newConstrFns)
+
+-- Case 2: only one selector, no need to compute combinations.
+-- this function will process: (1) objectives, (2) constraints and (3)
+-- assignments
+procBlock (dict, objFns, constrFns) (selector : [], stmts) =
+    let mapsAndSpecs  = dict `matchWith` selector
+        newDict       = foldl (addShapes stmts) dict $ mapsAndSpecs
```
hypotext: $ not needed here and on 404

src/Style.hs
```diff
@@ -300,130 +314,145 @@ initSpec = StySpec { spType = C.PointT,
...
+    let keys          = M.keys dict
+        identityMap   = M.fromList $ zip keys keys
+        newObjFns     = concatMap (procObjFn identityMap [])    stmts
+        newConstrFns  = concatMap (procConstrFn identityMap []) stmts
+    in (dict, objFns ++ newObjFns, constrFns ++ newConstrFns)
+
+-- Case 2: only one selector, no need to compute combinations.
+-- this function will process: (1) objectives, (2) constraints and (3)
+-- assignments
+procBlock (dict, objFns, constrFns) (selector : [], stmts) =
```
hypotext: give dict a more descriptive name

src/Style.hs
```diff
@@ -300,130 +314,145 @@ initSpec = StySpec { spType = C.PointT,
...
-        selectedSpecs = map
-            (\sel -> let styspecs = matchingSpecs dict sel -- defined externally above procblock
-                         varmaps = map (getVarMap sel) styspecs in zip varmaps styspecs) selectors
+-- Gather a list of objective or constraint functions by mapping over mergedMaps
+-- genFns :: (VarMap -> [t] -> Stmt -> [t]) -> VarMap -> [t]
+genFns f stmts varmap = foldl (f varmap) [] stmts
 
+-- | 'procBlock' is called by 'getDictAndFns'. 'getDictAndFns' would fold this function on a list of blocks, a.k.a. a Style program, and accumulate objective/constraint functions, and a dictionary of geometries to be rendered.
+-- TODO: first split `stmts` into assigns, objs, and contrs and call procConstrFn and procObjFn with `map`?
+procBlock :: (Autofloat a) =>
```
hypotext: what does "proc" mean? consider giving all the "proc" functions more descriptive names (e.g. "blockToIR")

src/Style.hs
hypotext: looks like we’re matching the sub dict with the selector, but then adding shapes to the dict? is it possible that near the beginning of a style program, we might try to match on something that hasn’t been added to the dict yet?

src/Style.hs
```diff
@@ -300,130 +314,145 @@ initSpec = StySpec { spType = C.PointT,
...
+-- TEMP: the system does not support assignment statements in these blocks,
+-- because it is not clear that which Substance object the shapes associate with
+procBlock (dict, objFns, constrFns) (selectors, stmts) =
+    let mapsAndSpecs  = dict `matchWithAll` selectors
+        mergedMaps    = mergeMaps $ map (map fst) $ mapsAndSpecs
+        newObjFns     = concatMap (genFns procObjFn stmts)    mergedMaps
+        newConstrFns  = concatMap (genFns procConstrFn stmts) mergedMaps
+    in (dict, objFns ++ newObjFns, constrFns ++ newConstrFns)
+    where
+        validMap :: [(Name, Name)] -> Bool
```
hypotext: can you add a comment on the purpose of these functions in the where clause?

src/Style.hs
```diff
@@ -283,12 +269,40 @@ attribute = identifier -- TODO: Naming convention - same as identifiers?
...
+-- * a list of geometric primitives
+-- * a list of constraints
+-- * a list of objectives
+
+-- Type aliases for readability in this section
+-- | 'StyContext' maintains the current output of the translater
+type StyContext a =
+    (StyDict,         -- | dictionary mapping Substance ID to Style objects
+    [ObjFnInfo a],    -- | List of objective functions
+    [ConstrFnInfo a]) -- | List of constraints
```
hypotext: should we also build a list of computations in Style? right now i'm doing it in an inconsistent manner:
* since comps can only set properties of shapes, they are added to a shape's list of properties (config) in addSpec
* comps are discovered and argument names resolved in Runtime, in compsAndVars (called by getShape)

src/Style.hs
```diff
@@ -283,12 +269,40 @@ attribute = identifier -- TODO: Naming convention - same as identifiers?
...
--- Functions used by "Runtime"
+-- Functions used by "Runtime" -- Style Translator
+-- Translates from a Style AST to:
+-- * a list of geometric primitives
+-- * a list of constraints
+-- * a list of objectives
+
+-- Type aliases for readability in this section
+-- | 'StyContext' maintains the current output of the translater
+type StyContext a =
```
hypotext: this name might be confusing because i refer to a VarMap as a context (and in PL a context is typically a list of `varName : varType` of variables in the current scope)

src/Style.hs
hypotext: this new procBlock is a lot clearer than the old one! thanks1

src/Style.hs
```diff
@@ -300,130 +314,145 @@ initSpec = StySpec { spType = C.PointT,
...
 
--- Helper functions for procBlock
-
--- removeDups :: [[(VarMap, StySpec)]]
--- TODO: really a helper function. consider moving to "Utils"
-cartesianProduct :: [[a]] -> [[a]]
-cartesianProduct = foldr f [[]] where f l a = [ x:xs | x <- l, xs <- a ]
+--------------------------------------------------------------------------------
+-- Selector implementation
+-- GOAL: given a list of selectors and all Substance identifiers, output a
```
hypotext: can you give an example here of a "list of selectors"? and what are the semantics of a list of selectors? is it AND (match all of them) or OR (match at least one of them)?

src/Style.hs
```diff
@@ -283,12 +269,40 @@ attribute = identifier -- TODO: Naming convention - same as identifiers?
...
+type StyDict = M.Map Name StySpec
 
------ Parser for Style design
+-- | Style specification for a particular object declared in Substance (declarations and constraints)
+-- (TODO: maybe this is not the best model, since there is no difference between "primary" and "secondary" shapes)
+-- NOTE: for Substance constraints such as `Subset A B`, the 'spId' will be '_Subset_A_B' and the 'spArgs' will be '['A', 'B']'
+data StySpec = StySpec {
+    spType   :: C.SubType,  -- | The Substance type of the object
+    spId     :: String,     -- | The Substance ID of the object
+    spArgs   :: [String],   -- | the "arguments" following the type.  The idea is to capture @f@, @A@ and @B@ in the case of @Map f A B@, which is needed for pattern matching (TODO: Maybe not the best term here.)
```
hypotext: do we ever use spArgs? what is it used for?

src/Utils.hs
```diff
@@ -63,12 +52,40 @@ subsetSizeDiff = 10.0
 epsd :: Floating a => a -- to prevent 1/0 (infinity). put it in the denominator
 epsd = 10 ** (-10)
 
+-- TODO: check naming convention of this
+labelName :: String -> String
```
hypotext: i changed the internal naming scheme to “subObjName styObjName [label]” (e.g. "A xaxis", "A xaxis label"); see Utils in my branch)

src/Style.hs
```diff
@@ -300,130 +314,145 @@ initSpec = StySpec { spType = C.PointT,
...
+        addShapes :: [Stmt] -> StyDict -> (VarMap, StySpec) -> StyDict
+        addShapes statements dct (varmap, spec) =
+            let newSpec = foldl (procAssign varmap) spec statements
+            in M.insert (spId newSpec) newSpec dct
+
+-- Case 3: multiple selectors, which in the current Style languages means
+-- we would take a subset of the cartesian product among the selectors.
+-- TEMP: the system does not support assignment statements in these blocks,
+-- because it is not clear that which Substance object the shapes associate with
+procBlock (dict, objFns, constrFns) (selectors, stmts) =
```
hypotext: design Q: do we want to consider other ways to join/combine selectors? e.g. `Set X | Point X { … }` (matches a set or a point) if so, we'd need to come up with use cases for that, and see how the Style compiler would need to generalize

hypotext:
Looks a lot more readable, thanks! General comments:

* Since there are a lot of nested folds, I didn't follow everything that was happening. Can you summarize the algorithm in a couple of sentences? Something like "For each block, for each of its selectors build a variable mapping. Then for each constraint/objective in each block [resolve variables and generate the function]. And for each object in the block, [generate a shape specification] including each of the object's properties." (I'm probably missing layers here)
* Are there any invariants I should know about, and that should be maintained?
* Line 506 FIXME: is the variable resolution still incorrect?

src/Style.hs
wodeni: No. If you look for `initDict`, you will see we add all Substance IDs __before__ the first call of `procBlock`

src/Style.hs
wodeni: `434e2f2` gives a better implementation. Hope it's clearer

src/Style.hs
wodeni: I think things similar to `VarMap` are usually called symbol table. I always thought "context" is a set of environment variables that includes a symbol table. How about `StyEnv` for "Style environment"?

src/Style.hs
wodeni: I did try to cover it the [Style document](https://github.com/wodeni/notes-public/blob/master/penrose/style-compiler-doc.md#pattern-matching). I'll add examples to the doc for better explanation.

src/Style.hs
wodeni: > "The idea is to capture @f@, @A@ and @B@ in the case of @Map f A B@, which is needed for pattern matching (TODO: Maybe not the best term here.)"

So this `spArgs` field exists because some Substance objs are defined by an n-tuple. For instance, `Subset A B` is effectively `(A, B)` with a type `Subset`. Therefore, to pattern match on them, we need a record of all arguments in a tuple.

wodeni:
> "For each block, for each of its selectors build a variable mapping. Then for each constraint/objective in each block [resolve variables and generate the function]. And for each object in the block, [generate a shape specification] including each of the object's properties." 

That looks like all of them. I will add details of it in the Style document.

src/Style.hs
hypotext: i agree that symbol table is a better name for `VarMap`, but not sure about environment, since this type describes something that's more of a "concrete output" and not just "auxiliary information"

hypotext:
> Can you elaborate? I can't think of a particular one yet

There are some invariants in Runtime like "the shapes in the list of objects must always be kept in the same order so packing and unpacking can work correctly." Wondering if there's anything like that in Style.

> It wasn't an accurate comment. The variable resolution is not "incorrect", but was a design choice that might just be necessary. You need to keep holding the VarMap before you get down to an Id, where the resolution actually happen. Therefore, I think it will stay necessary unless we do a preprocessing pass to resolve the ids first. There could be many ways. For example, do pattern matching first and generate a new Style file with all concrete selectors (Set `A`)

Yeah, as the Style language develops we should rework the name resolution (right now I haven't done it consistently for args to computations vs. args to functions in optimization).

wodeni:
@hypotext I don't think we have invariants like that in `Style`, since it depends on the AST, which is structured. The biggest thing is probably that __a Style program is a collection of blocks__ and __a block is a collection of statements, which can be objectives, constraints, or object declarations__. If these two facts change, changes are needed to these `proc*` functions.

There are surely hardcoded components in `procExpr` (since it was hacky to begin with) and let me know if you have questions about that.

I also added a detailed explanation of how multiple selectors work in the [Style doc](https://github.com/wodeni/notes-public/blob/master/penrose/style-compiler-doc.md)

hypotext:
Thanks!

Issue #65, opened 2018-04-25
Build an explicit computational graph; infer which variables are fixed/varying/computed


hypotext:
Analyze a Substance and Style program and build an explicit computational graph so we know in what order to apply computations. (Right now we construct an implicit computational graph for the autodiff by just applying computations in order, so we are able to differentiate through the computational graph.)

Right now, since we haven't built that graph, to create the final diagram, we need to apply the computations again a couple of times, which is a hack, especially if any of the computations takes a long time to run and doesn't need to be applied multiple times. See the use of `computeN` [here](https://github.com/penrose/penrose/blob/surjection/src/Server.hs#L110).

We should probably hold off on implementing this until we have a more solid Substance and Style design and semantics.

Relevant internal docs: I worked out a lot of the theory in the [computation and optimization](https://docs.google.com/document/d/1iayNUjlFquCwmyNy2Os4SU-u8IU_gqYL2HJ72JHNbiY/edit#heading=h.z58y75c8yvy1) doc. Jenna also has several docs on a graph semantics for Substance and Style.

---

Also, infer which variables are fixed or varying or computed for each object (and in what order to apply the computations)! Right now we can't figure that out, which leads to bad hacks and bugs like manually setting parameters to fixed or varying for different shapes for each diagram (e.g. fixed arrows for surjections).

If we could automatically visualize the computational graph, that would be even better.

wodeni:
Currently, the order of which computation is applied is dictated by the order of declaration in Substance, since we access properties of objects by accessing `[(Name, Properties)]`, where `Name` is in Substance declaration order.

Ideally, we would like to analyze the style program such that the computation is ordered based on the computational graph.

hypotext:
Solved in the system rewrite. Explicitly building a computational graph is unnecessary (for now). To evaluate a computation, all its arguments are recursively evaluated in the translation and their results are cached, then the computation is applied and its result cached. This method automatically evaluates computations in the right order. (We track evaluation depth in case of cycles in the graph.)

For the optimization, the top-level evaluated expressions are all the objective functions' and constraints' arguments.

For the display, the top-level evaluated expressions are all the shapes.

Paths to varying variables are found and saved in the opt problem generation phase. Varying variables are (generally) uninitialized floats or floats initialized as `optimized` in Style (with some exceptions in the case of labels). No inference is needed.

PR #67, opened 2018-04-27
Add surjection diagram in cartesian style and clean up labeling/naming model

hypotext:
This PR adds the following:

* Surjection diagram and programs: `runpenrose snap sub/surjection_computed.sub sty/surjection_computed.sty` (see wiki for picture)

NOTE: This example only works when all the Arrow parameter annotations are set to Fix, due to [not yet having a computational graph](https://github.com/penrose/penrose/issues/65). Specifically, the axes' positions need to be fixed during the optimization so the other parameters can be optimized with respect to the right thing, and the final computed parameters can be computed with respect to the right thing. (Other than that issue, all examples work.)

* More standard, automatic model of computations that can take a list of tagged inputs `TypeIn a` and can return a tagged `TypeIn a`
* A generic `get` and `set` function for object properties
* Removes primary/secondary shape distinction in Style. Objects can have any number of primitives, which don't have to be named `shape`.
* Internally graphical primitives are called `SubObjName ShapeName` e.g. `A xaxis` or `A xaxis label`. The name delimiter has been changed to space from underscore.
* To access a label in Style, you must say `SubObjName.ShapeName.label`. Gone are the days of `SubObjName.label`. The latter is ambiguous because a Substance object may have multiple graphical primitives.
* Label text keyword is now `text = ...` not `label = ...`. Text objects are treated like labels (without their own labels).

NOT ADDED YET: 

* matching on `Surjection f` rather than `Map f`
* consistent treatment of `SubObjName` in Style as accessing all primitives vs. only the `shape` primitive vs. only one primitive (it must only have one).
* accessing properties of graphical primitives at any argument position in Style, e.g. `A.xaxis.length` ("dynamic gets")
* consistent model of name resolution / accesses / property gets


hypotext:
@wodeni Feel free to review if you'd like. I'm going to merge the branch now so I don't block Dor on his work.

Issue #53, opened 2018-02-16
Parse negative numbers in Style


hypotext:
The first step here is to replace some of the parameters in the computations/objectives with negative numbers (see the sample programs in the wiki), see what breaks, and then fix the parser.

Possibly following this guide: https://www.schoolofhaskell.com/school/to-infinity-and-beyond/pick-of-the-week/parsing-floats-with-parsec#minus

PR #68, opened 2018-04-27
Dor work

dormaayan:
Hi,
Excited to have my first pull request for penrose 🎉🎉🎉🎉
I will not do a pull request for every issue in the future (I guess it is better to do that when I finish a milestone), this is just for checking!

I solved Issue #53  and now the style parser support negative floats and integers.
I demonstrated that by adding a new substance and style programs: 

- twopoints_negative.sub

- twopoints_negative.sty

All the changes were done on utils.hs

@hypotext , @wodeni :
please review and merge

### DIFF
--- a/src/Utils.hs
+++ b/src/Utils.hs
@@ -176,11 +176,13 @@ parens = between (symbol "(") (symbol ")")
 
 -- | 'integer' parses an integer.
 integer :: Parser Integer
-integer = lexeme L.decimal
+unsignedInteger       = lexeme L.decimal
+integer = L.signed sc unsignedInteger
 
 -- | 'float' parses a floating point number.
 float :: Parser Float
-float = realToFrac <$> lexeme L.float -- TODO: parsing without sign?
+unsignedFloat = lexeme L.float
+float = L.signed sc unsignedFloat
 
 -- Reserved words
 rword :: String -> Parser ()
--- a/src/sty/twopoints_negative.sty
+++ b/src/sty/twopoints_negative.sty
@@ -1,0 +1,32 @@
+global {
+  -- objective sameX(a, b)
+}
+
+Point x {
+  shape = Dot {
+    text = Auto
+  }
+
+  objective centerLabel(x, x.shape.label)
+}
+
+Point `a` {
+  objective center(a)
+}
+
+Point `b` {
+  shape = Dot {
+    location = addVector(-100.0, -200.0, a)
+    text = Auto
+  }
+}
+
+Set X {
+  shape = Circle {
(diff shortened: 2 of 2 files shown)


src/Utils.hs
```diff
@@ -176,11 +176,13 @@ parens = between (symbol "(") (symbol ")")
 
 -- | 'integer' parses an integer.
 integer :: Parser Integer
-integer = lexeme L.decimal
+unsignedInteger       = lexeme L.decimal
```
hypotext: We try to have type signatures for all top-level functions in the codebase, so it would be good to add one for `unsignedInteger` and `unsignedFloat`

src/Utils.hs
```diff
@@ -176,11 +176,13 @@ parens = between (symbol "(") (symbol ")")
...
+integer = L.signed sc unsignedInteger
 
 -- | 'float' parses a floating point number.
 float :: Parser Float
-float = realToFrac <$> lexeme L.float -- TODO: parsing without sign?
+unsignedFloat = lexeme L.float
+float = L.signed sc unsignedFloat
```
hypotext: Using `sc` here means that `-     100.0` will parse as `-100.0`. Did you intend for that to be valid? (If so, that's fine)

dormaayan: Right, if someone wrote a float, then he wrote a float

src/Utils.hs
dormaayan: I'm not sure I understand what do you mean here

hypotext:
Thanks! Some comments:

* You can remove `twopoints_negative.sub`, as it's the same as `twopoints.sub`.
* If I merge this PR, it will also merge the commit with your work on the perpendicular mark, so you might want to move that commit to another branch first.
* Can you squash the first two commits? (The fix and the example programs)
* @wodeni is better qualified to review if this change affects the parser globally. Locally it looks fine to me.
* Please add a picture to future tests so we can check if we got the same results. Looks like it works. 
image

src/Utils.hs
hypotext: Looking at the diff, it looks like you don't have type signatures for `unsignedInteger` or `unsignedFloat`. Could you add them?

dormaayan:
Ok, now there is a single commit of all the work in the branch (Includes the removing of the duplicated substance) and my work from today is on a separate branch.
You can merge now

dormaayan:
@wodeni, @hypotext : I ran the the suite (`stack test`) and there are 9 failures. I also tried on the master and there is the same issue, so it doesn't seem to relate to my changes.
Are you aware to that? Or to open an issue?

Issue #64, opened 2018-04-25
Add "perpendicular mark" and "angle mark" as shapes


hypotext:
[Intro task for Dor]

In several of our diagram styles, we use the perpendicular mark, but we haven't added it to the system yet. We'd like to add it to already-working diagrams (e.g. diagrams in the Cartesian style) as well as upcoming diagrams (e.g. diagrams in linear algebra that have perpendicular axes and perpendicular vector projections).

Can you add this shape to our system, as well as the past diagrams that use it? The steps to do this are outlined in the wiki [here](https://github.com/penrose/penrose/wiki/Extending-the-system).

Also, in general, we're going to need the "angle mark." This is usually denoted as an arc or a pie wedge, which you can see in the complex numbers blog post and in my raytracing diagrams. Can you add this as well?

dormaayan:
## Checklist for Adding a New Shape
* In Shapes
  - [ ]  add `data Shape = …` along with its typeclass instances (e.g. `Named`, `JSON`)
  - [ ]  add a case to `data Obj = … | S Shape`
  - [ ]  add cases for that shape `S` in all of the derived Obj typeclass instances (`Located`, etc. if possible; some instances might not be possible)
  - [ ]  add a polymorphic version of that shape `data Shape' a = …` along with its typeclass instances (e.g. `Named`, `JSON`)
  - [ ]  add a case to `data Obj a = … | S’ (Shape’ a)`
  - [ ]  add typeclass instances for `Shape’`
  - [ ]  add typeclass instances for `Obj’`
  - [ ]  add property getters in `get` function
  - [ ]  add property setters in `set` function
* In Runtime
  - [ ]  add a case in `unpackObj`
  - [ ]  add a case and function in `pack’`, making sure the parameter order and number matches that in `unpackObj`
  - [ ]  add a case and function `initShape` in `getShape` to initialize it using `defaultShape`
  - [ ]  add a default shape definition `defShape`
  - [ ]  add a wrapper function `defaultShape`
  - [ ]  add a way to randomly sample the shape in `sampleCoord` (optional)
  - [ ]  add case for `addGrad` (adding gradients, converting to polymorphic floats)
  - [ ]  add case for `zeroGrad` (removing gradients, converting to `Float` type)
* In Style
  - [ ]  add a case to `StyType` (avoiding name clashes)
  - [ ]  add the name of the type as a reserved word in the `styObj` parser
* In Functions
[...]

PR #70, opened 2018-05-02
Add Angle Marks

dormaayan:
This PR solves issue #64.
I added a new shape to Style named `AngleMark`
`AngleMark` allows you specify whether this is a **right angle** or a **regular angle** by specifying the parameter `isRight = "true"\"false"`. Other parameters you might specify for the regular angle sign are the radius, the rotation of the angle and the angle itself.
I added prependicular mark to the following .sty files:

- surjection_computed.sty:
image

- cart-test-pt.sty:
image

I also added my own examples, both run with angles.sub:

- regularAngle.sty (these are 2 different angles):
image
image

- rightAngle.sty:
image

@hypotext , @wodeni 
The system pass the tests after those changes, please let me know if you have comments


src/Shapes.hs
```diff
@@ -197,6 +197,36 @@ instance Named Square where
...
+-------------------------- (Angle Mark, added by Dor)
+data AngleMark = AngleMark { xam :: Float -- Starting point for angle mark
+                     , yam :: Float
+                     , radiusam :: Float -- The radius of the angle mark
+                     , sizeam :: Float
+                     , nameam :: String
+                     , coloram :: Color
+                     , angleam :: Float -- The angle that the angle mark specifies
+                     , rotationam :: Float -- The rotation angle of amgle mark
+                     , isRightam :: String -- Is this a right angle or not
```
hypotext: This is probably better as a bool or an enum `data AngleType = Right | NotRtAngle`

dormaayan: Yes I know, I had trouble doing that at the begging but now I think I know how to solve that

src/sty/regularAngle.sty
```diff
@@ -0,0 +1,60 @@
...
+}
+
+Set `B` {
+  shape = Arrow {
+    start = origin
+    end   = vert
+    text = None
+  }
+
+    angleMark2 = AngleMark{
```
hypotext: Style style guideline: space between keyword and brace

src/sty/surjection_computed.sty
```diff
@@ -21,7 +21,12 @@ Set `codomain` {
           text = "f(x)"
     }
 
-    objective nearHead(codomain, codomain.shape.label, 0, 20) -- TODO parse negative numbers
+    angleNotation = AngleMark {
+      start = atOrigin()
```
hypotext: how are the other parameters of the arc being determined here?

hypotext:
It would also be nice to be able to choose from two arc styles "line" or "wedge," the latter of which fills in the angle with a solid color of lower opacity, as in Keenan's blog post on complex numbers.

hypotext:
Thanks for the changes so far! Let me know when the `from` and `to` API is ready, and I'll give the PR a second review. Also, it would be nice to be able to turn off the arc outline if there is a fill, but that's not super important.

BTW, Keenan requests that all fills be done with the color `#1b1f8a` at opacity 20%. :-)

Issue #63, opened 2018-04-25
Visualize more functions (injection, bijection, etc.) in the Cartesian style


hypotext:
[Intro task for Dor]

We have surjections, injections, and bijections all working in the "discrete" set theory style. But we only have surjections working in the Cartesian style (see the [example page](https://github.com/penrose/penrose/wiki/Examples)). 

Can you get injections and bijections working in the Cartesian style, and add a new function type of your own?

Note: I will merge the `surjection` branch first; then you can start working on this.

hypotext:
> As I understand it, I should define similar computing functions to those of surjection at the file: Computation.hs and then simply to use them in a dedicated style program?

Yes, you'll need to write and register computations to sample the various kinds of functions, then write new Substance and Style programs for them. You should also extend the parser to match on the type of function (surjection/injection/etc) as currently it just matches on Map (which is too general).

dormaayan:
@hypotext By extending the parser to match on the type of function do you mean to extend Substance (i.e `Surjection f` as in your comments)? Is there doc anywhere on how to extend Substance, if not I'll simply understand how to do that from the code

wodeni:
@dormaayan the doc, unfortunately, does not cover it in great details partly because the process will not be as structured as adding a shape. In general, you need to do the following to get some custom syntax you add to work (ordered in my preferred way):

- modify the AST to reflect your grammar
- add parser functions using Megaparsec (e.g. if you want to add a new type of statement, start by adding another option to `stmt` function)
- go into one of the `proc*` functions in the same file and add the semantic analysis/"code generation" logic there. Continuing the example, you would go to `procStmt` and add custom code there, depending on what you are adding. 
- Modify Runtime, Server etc to get the whole system running. 
- If you changed anything that would affect Style (for now just `SubType`), you need to make some changes to Style as well.

This is super high-level because I haven't thought too much about this specific feature yet.

dormaayan:
@hypotext : We have 2 options for substance syntax here:

1. Declare a new object like: `Surjection f A B`

2. Working on maps: `Map f A B` and in a new line: `Surjection f` (On the same way that intersect \ non intersect is working)

Personally the second syntax make more sense to me. However, I think that we can't use it until we will have the `where` syntax in style. What do you think?

hypotext:
[...]

PR #71, opened 2018-05-05
Issue #63 - More Functions

dormaayan:
This PR solves issue #63.

- Added random functions for generation of bijective and injective functions

- Added 3 new Substance objects for representation of functions: `Surjection, Injection, Bijection`

- Added tests for injection and bijection:

`./runpenrose snap sub/bijection_computed.sub sty/bijection_computed.sty`

bijection

`./runpenrose snap sub/injection_computed.sub sty/injection_computed.sty`

image

Future TODOS:

- Change `Surjection, Injection, Bijection` to work over maps (More work on Style is needed before doing so)

- Change the randomness to be on runtime and to make the diagrams to change each run, also to make the injection generation fully random

- Solve the label issue

### DIFF
--- a/src/Computation.hs
+++ b/src/Computation.hs
@@ -12,6 +12,7 @@ import System.Random.Shuffle
 import Data.List (sort)
 import Data.Dynamic
 import Data.Data
+import Data.List
 import Data.Typeable
 
 type Interval = (Float, Float)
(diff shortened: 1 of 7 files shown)


src/Computation.hs
```diff
@@ -65,13 +66,43 @@ computeSurjection g numPoints (lowerx, lowery) (topx, topy) =
...
+-- TODO pass randomness around in Runtime
+computeBijection :: Autofloat a => StdGen -> Integer -> Pt2 a -> Pt2 a -> ([Pt2 a], StdGen)
+computeBijection g numPoints (lowerx, lowery) (topx, topy) = 
+                  if numPoints < 2 then error "Bijection needs to have >= 2 points" 
+                  else let (xs_inner, g') = randomsIn g (numPoints - 2) (r2f lowerx, r2f topx)
+                           xs = lowerx : xs_inner ++ [topx] -- Include endpts so function covers domain
+                           xs_increasing = nub (sort xs)
+                           (ys_inner, g'') = randomsIn g' (numPoints - 2) (r2f lowery, r2f topy) 
+                           ys = lowery : ys_inner ++ [topy] --clude endpts so function is onto
+                           ys_increasing = nub (sort ys) in -- Random permutation. TODO return g3?
```
hypotext: Try to generate a more thorough sampling of bijections (e.g. monotonically decreasing as well)

src/Computation.hs
```diff
@@ -65,13 +66,43 @@ computeSurjection g numPoints (lowerx, lowery) (topx, topy) =
...
+-- Given a generator, number of points, and lower left and top right of bbox, return points for a injection.
+-- Points generated lie in the bbox given, whether in math space or screen space
+-- TODO pass randomness around in Runtime
+computeInjection :: Autofloat a => StdGen -> Integer -> Pt2 a -> Pt2 a -> ([Pt2 a], StdGen)
+computeInjection g numPoints (lowerx, lowery) (topx, topy) = 
+                  if numPoints < 2 then error "Injection needs to have >= 2 points" 
+                  else let (xs_inner, g') = randomsIn g (numPoints - 2) (r2f lowerx, r2f topx)
+                           xs = lowerx : xs_inner ++ [topx] -- Include endpts so function covers domain
+                           xs_increasing = nub (sort xs)
+                           (ys_inner, g'') = randomsIn g' (numPoints - 2) (r2f (lowery + (topy - lowery)/4), r2f (topy - (topy - lowery)/4))
```
hypotext: Again, try to generate a more thorough sampling of possible injective functions. This strategy will, for example, never happen to produce a bijective function.

src/Computation.hs
```diff
@@ -92,6 +123,27 @@ computeSurjectionLines g n left right bottom top =
...
+                                   -> Line' a -> Line' a -> Line' a -> Line' a -> ([Pt2 a], StdGen)
+computeBijectionLines g n left right bottom top = 
+                       let lower_left = (startx_l' left, starty_l' bottom) in
+                       let top_right = (startx_l' right, starty_l' top) in
+                       computeBijection g n lower_left top_right
+
+-- Computes the injection to lie inside a bounding box defined by the corners of a box 
+-- defined by four straight lines, assuming their lower/left coordinates come first. 
+-- Their intersections give the corners.
+computeInjectionLines :: (Autofloat a) => StdGen -> Integer
```
hypotext: Some code duplication here (the `Lines` functions). Could be refactored into a single `computeFunctionLines` function.

src/Substance.hs
```diff
@@ -265,6 +295,15 @@ checkDecls e (DeclList t ss) = e { subObjs = objs, subSymbols = syms }
 checkDecls e (MapInit t f a b) =
```
hypotext: Combine these four cases?

src/Substance.hs
```diff
@@ -437,6 +490,28 @@ objToAlloy e (LD (Map f x y)) = e { alSigs = newSigs }
            (SigDecl n l, m) = case M.lookup x $ alSigs e of
                 Nothing -> let sig = SigDecl x [] in (sig, M.insert x sig $ alSigs e)
                 Just s -> (s, alSigs e)
+
+objToAlloy e (LD (Surjection f x y)) = e { alSigs = newSigs }
```
hypotext: Why do we need to modify the Alloy code?

src/sty/bijection_computed.sty
```diff
@@ -0,0 +1,83 @@
...
+          end = toAbove()
+          text = "f(x)"
+    }
+
+    objective nearHead(codomain, codomain.shape.label, 0, 20) -- TODO parse negative numbers
+}
+
+Bijection f domain image {
+    shape = Curve {
+          path = computeBijectionLines(5, domain, image)
```
hypotext: Try testing these functions by using more points (e.g. changing 5 to 10)

src/sty/injection_computed.sty
```diff
@@ -0,0 +1,83 @@
...
+          text = "f(x)"
+    }
+
+    objective nearHead(codomain, codomain.shape.label, 0, 20) -- TODO parse negative numbers
+}
+
+Injection f domain image {
+    shape = Curve {
+          path = computeInjectionLines(5, domain, image)
+          text = "(injenction)"
```
hypotext: sp: injection

hypotext:
Thanks! These do indeed look like they satisfy the bijection/injection properties. It's nice to see more of these examples.

Overall, it would be nice to have more "random looking" / "representative" functions, more like what a human would draw. Check out the results for an image search on "surjection, injection, bijection."

Changing the rand seed to 17 and changing the number of points to 10 results in these funky looking relations (that don't seem to be functions). I guess this is happening because the runtime is not using the exact path of the curve that is connecting the points (which is not your fault).

image
image

When this PR is done, could you add the images and repro steps to the wiki as example?

The TODOs you listed are all reasonable, but you don't need to tackle them anytime soon.

src/Computation.hs
dormaayan: I think that it is better like that since the name of the function is clearer (the other option is to ask the style writer to pass a parameter which specify the desired function style)  and the compaction is insignificant

Issue #72, opened 2018-05-08
Write a DSLL Parser


dormaayan:
Full description can be found [here](https://docs.google.com/document/d/1DPpAvBKGnr96MyohJAj2M40zrppjPV7sbd25Nz-qfkg/edit)

dormaayan:
@hypotext: Did that

PR #73, opened 2018-05-10
Plugin Implementation

dormaayan:
This PR regards the modifications which are made to the system in order to support the plugin module.
This work includes:

- Implementation of the DSLL (Issue #72 )
- Implementation of the Substance Core language (Issue #74 )

Running Instructions:

- For the DSLL: `ghc Dsll.hs; ./Dsll input.dsll output`


src/dsll_progs/algebra.dsll
```diff
@@ -0,0 +1,15 @@
+constructor Vector : [VectorSpace] -> type
```
hypotext: Syntax changes look fine to me

src/dsll_progs/algebra.dsll
```diff
@@ -0,0 +1,15 @@
+constructor Vector : [VectorSpace] -> type                    
+constructor Scalar : type                                               
+constructor Map : [type, type] -> type
+constructor VectorSpace : type                                      
+
+operator AddV : forvars [V] : [VectorSpace], [Vector(V), Vector(V)] -> Vector(V)
+operator AddS : [Scalar, Scalar] -> Scalar
+operator Norm : forvars [V] : [VectorSpace], [Vector(V)] -> Scalar
+operator InnerProd : forvars [V] : [VectorSpace], [Vector(V), Vector(V)] -> Scalar
+operator Apply : fortypes [A,B] : [type,type], [Map(A, B), A] -> B
```
hypotext: There's an inconsistency between line 10 and line 12 (due to the inconsistent way I originally wrote the book):
`fortypes [A,B] : [type,type]` vs. `forvars [U,V] : [VectorSpace]`
I'd say make all types explicit for now (don't allow multiple declaration). We can add concision later.
Also, if there's a long list of quantifiers, it might be better to change the syntax to
`forvars [(v : T)]` and `fortypes [(X : Type)]` (i.e. each variable is given its type, rather than having two separate lists). I think that's more readable.

dormaayan: Yes, the syntax change you suggested is much more readable

src/Dsll.hs
```diff
@@ -0,0 +1,397 @@
...
+    t' <- typeParser
+    return Cd {nameCd = name, inputCd =  k', outputCd =  t'}
+cd2 = do
+    rword "constructor"
+    name <- identifier
+    colon
+    t' <- typeParser
+    return Cd {nameCd = name, inputCd =  [], outputCd =  t'}
+
+odParser, od1, od2, od3, od4 :: Parser Od
```
hypotext: I wonder if there's some way to combine these cases (for cd, od, and pd)--having 0 or 1 of forvals or fortypes in any combination. (You've probably already thought of this :-))

src/Dsll.hs
```diff
@@ -0,0 +1,397 @@
...
+-- --     show (Od nameOd forVarsoD typesForVarsOd forTypesOd typeForTypesOd fromOd toOd) = 
+-- --     	"(Op, " ++ nString ++ " ,ValOfType " ++ iString ++ "Output " ++ oString ++")"
+-- --         where aString = show nameOd
+-- --               bString = show inputCd
+-- --               cString = show outputCd
+-- --               dString = 
+-- --               eString = 
+
+
+data Pd = Predicate String [Var] [T] [TypeVar] [Type] [T] [Prop] Prop
```
hypotext: Probably better to have Od and Pd as record types too (looks like you're already converting them)

src/Dsll.hs
dormaayan: Sure I did, that's why I wrote the code need a major refactoring because it looks bad ;-))

src/Dsll.hs
```diff
@@ -0,0 +1,397 @@
...
+import Data.Typeable
+import Text.Megaparsec
+import Text.Megaparsec.Char
+import Text.Megaparsec.Expr
+--import Text.PrettyPrint
+--import Text.PrettyPrint.HughesPJClass hiding (colon, comma, parens, braces)
+import qualified Data.Map.Strict as M
+import qualified Text.Megaparsec.Char.Lexer as L
+--------------------------------------- DSLL AST ---------------------------------------
+type TypeVar = String
```
hypotext: I would still keep the constructors here even though there's only one branch in the ADT:
`data TypeVar = TVar String`, and analogously for Var
The reasoning is that we'll probably be doing a lot of debugging by examining the pretty-printed AST, and it's useful to see what kind of var or name something is, rather than have everything be a string
(I guess you could also put this information in the pretty-printer rather than the ADT)

src/Dsll.hs
```diff
@@ -0,0 +1,397 @@
...
+
+type Var = String
+
+type Prop = String
+
+type Type = String
+
+
+data ConstructorInvoker = ConstructorInvoker { nameCons :: String, argCons:: [Arg]}
```
hypotext: We're not currently using Typeable elsewhere in the system, so feel free to remove that from the typeclasses (unless you're planning to use it)

src/Dsll.hs
```diff
@@ -0,0 +1,397 @@
...
+    deriving (Eq, Typeable)
+instance Show ConstructorInvoker where
+    show (ConstructorInvoker nameCons argCons) = nString ++ "(" ++ aString ++ ")"
+        where nString = show nameCons
+              aString = show argCons
+
+
+data T = TTypeVar TypeVar --TODO: figure out a better name for that...
```
hypotext: Looking at the output, it seems that things like `Set` are being labeled/treated as TTypeVars, when they should really be labeled/treated as (nullary) type constructors

src/Dsll.hs
```diff
@@ -0,0 +1,397 @@
...
+import qualified Data.Map.Strict as M
+import qualified Text.Megaparsec.Char.Lexer as L
+--------------------------------------- DSLL AST ---------------------------------------
+type TypeVar = String
+
+type Var = String
+
+type Prop = String
+
+type Type = String
```
hypotext: is the only value of type Type the string `type`? then I would just have it be the nullary value constructor `TypeOfType` (because it is the type of a type)

src/Dsll.hs
```diff
@@ -0,0 +1,397 @@
...
+--import Text.PrettyPrint
+--import Text.PrettyPrint.HughesPJClass hiding (colon, comma, parens, braces)
+import qualified Data.Map.Strict as M
+import qualified Text.Megaparsec.Char.Lexer as L
+--------------------------------------- DSLL AST ---------------------------------------
+type TypeVar = String
+
+type Var = String
+
+type Prop = String
```
hypotext: likewise for Prop (same comment as for Type)

src/Dsll.hs
```diff
@@ -0,0 +1,397 @@
...
+-- --               bString = show inputCd
+-- --               cString = show outputCd
+-- --               dString = 
+-- --               eString = 
+
+
+data Pd = Predicate String [Var] [T] [TypeVar] [Type] [T] [Prop] Prop 
+    deriving (Show, Eq, Typeable)
+
+data DSLLProg = DSLLProg{ cd :: [Cd], od :: [Od], pd :: [Pd]}
```
hypotext: would give this a more descriptive name like `DSL_Definition`

hypotext:
Overall, the work looks really promising, and it's awesome that you got it working so fast!

dormaayan:
@hypotext Now you can translate to a file and have a look on the output, let me know if we think we should change things regarding the output

src/Env.hs
```diff
@@ -0,0 +1,255 @@
...
+argParser, varParser', tParser'  :: Parser Arg
+argParser = try tParser' <|> varParser'
+varParser' = do
+     i <- varParser
+     return (AVar i)
+tParser' = do
+    t <- tParser
+    return (AT t)
+
+emptyArgList :: Parser [Arg]
```
wodeni: Any particular reason why you require `()` here? This breaks the existing syntax of `Set A` and requires `Set() A` instead, which causes most of the examples to fail.

src/dsll_progs/setTheory.dsl
```diff
@@ -0,0 +1,26 @@
+tconstructor Set: type 				
+tconstructor Map([a,b] : [Set(),Set()]) : type
+tconstructor Map2(['a,'b] : [type,type]) : type
```
wodeni: Any particular reason why we need `Map2`, for maps between objects of arbitrary types??

src/Env.hs
dormaayan: Yes, since in our grammar we might give arguments for type definitions.  For example, specify a vector space as an argument to a vector. I haven’t defined a syntactic sugar for that yet but we definitely should consider that.

src/Utils.hs
```diff
@@ -114,13 +115,13 @@ uniqueShapeName subObjName styShapeName = subObjName ++ nameSep ++ styShapeName
 type Parser = Parsec Void String
 
 rws, attribs, attribVs, shapes, types :: [String] -- list of reserved words
-rws =     ["avoid", "global", "as"] ++ types ++ shapes
+rws =     ["avoid", "global", "as"] ++ types ++ shapes ++ dsll
```
wodeni: `dsll` consists of reserved words for a separate language. Maybe we need to parametrize `rword` function by different sets of reserved words.

src/dsll_progs/setTheory.dsl
dormaayan: Conceptually we might like to define map to be generic and not limit it to work only over sets . 
Because of the way the languages are built now, in a case you create a type which get a generic argument, you can not provide to it objects that has been defined before. For example, the following program will not work:
`Set() A
Set() B
Map2(A,B)`
But it will work with `Map(A,B)` which is defined specifically over sets. I don’t think that’s ideal, I discussed that with Josh before I left but we didn’t decide to change the design in that stage

src/Utils.hs
dormaayan: Agree, I’ll definietly do that. There is not any good reason to mix it up

src/Dsll.hs
```diff
@@ -131,6 +131,8 @@ cd1= do
     (y, k) <- parens ykParser
     colon
     t' <- typeParser
+    pos <- getPosition
```
wodeni: Wow didn't know getting src position is that easy! Looks great 😆

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+
+data Y = TypeVarY TypeVar
+       | VarY Var
+         deriving (Show, Eq, Typeable, Ord)
+
+data T = TTypeVar TypeVar
+       | TConstr ConstructorInvoker
+         deriving (Show, Eq, Typeable)
+
+data ConstructorInvoker = ConstructorInvoker { nameCons :: String,
```
hypotext: I would name this `TypeCtorApp` (type constructor application). Also, don't we need one for applied value constructors? (Their argument types are different)

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+              then e {errors = (errors e) ++ ("Name already exsist in the context \n")}
+              else e {names = a : (names e)}
+
+addDeclaredName :: String -> VarEnv -> VarEnv
+addDeclaredName a e = if (a `elem` (declaredNames e))
+                      then e { errors = (errors e) ++ ("Name already exsist in the context \n") }
+                      else e { declaredNames = a : (declaredNames e) }
+
+--------------------------------------- Env Data Types ---------------------------------------
+
```
hypotext: I would move all these definitions up to join the AST definitions, esp. since `VarEnv` is used before it's defined

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+import qualified Dsll as D
 import qualified Data.Map.Strict as M
 import qualified Text.Megaparsec.Char.Lexer as L
 
---------------------------------------------------------------------------------
--- Substance AST
+--------------------------------------- Substance AST ---------------------------------------
 
--- | A Substance program is a list of statements
+data ValConstructorName = ValConst String             -- “Cons”, “Times”
```
hypotext: I am confused about the relationship b/t Sub and SubCore. Which one is used? Which one is deprecated? Seems like they should be merged

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+checkT e (TConstr c)  = checkConstructorInvoker e c
+
+checkType :: VarEnv -> Type -> VarEnv
+checkType e t = e
+
+-- --TODO: fix back...
+-- checkConstructorInvoker :: VarEnv -> ConstructorInvoker -> VarEnv
+-- checkConstructorInvoker e const = e
+
+checkConstructorInvoker :: VarEnv -> ConstructorInvoker -> VarEnv
```
hypotext: this function's name would also need to be changed if we change ConstructorInvoker

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+data Ttype = Ttype { yt :: Y,
+                     kt :: K }
+             deriving (Show, Eq, Typeable)
+
+data TypeConstructor = TypeConstructor { nametc :: String,
+                                         klstc  :: [K],
+                                         typtc  :: Type }
+                       deriving (Show, Eq, Typeable)
+
+data VarConstructor = VarConstructor { namevc :: String,
```
hypotext: these are *value* constructors, not variable constructors, and I would rename everything accordingly

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+                                       ylsvc  :: [Y],
+                                       klsvc  :: [K],
+                                       tlsvc  :: [T],
+                                       tvc    :: T }
+                      deriving (Show, Eq, Typeable)
+
+data Operation = Operation { nameop :: String,
```
hypotext: would rename to Operator everywhere to be consistent w/ paper

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+                               ppred2    :: Prop }
+                  deriving (Show, Eq, Typeable)
+
+data VarEnv = VarEnv { typeConstructors :: M.Map String TypeConstructor,
+                       varConstructors  :: M.Map String VarConstructor,
+                       operations       :: M.Map String Operation,
+                       predicates       :: M.Map String PredicateEnv,
+                       typeVarMap       :: M.Map TypeVar Type,
+                       varMap           :: M.Map Var T,
+                       names            :: [String],  -- a global list which contains all the names of types in that env
```
hypotext: rename to something like "typeCtorNames"

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+                       declaredNames    :: [String],  -- a global list which contains all the names of elements declared in that env
```
hypotext: rename to something like "varNames"? I'm not clear on what this is for. maybe this should be called "typeCtorNames"? or "dsllNames"?

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+                       errors           :: String }   -- a string which accumulates all the errors founded during the run of the typechecker
```
hypotext: I would suggest accumulating `[Error]` where `Error = String` or a more detailed datatype

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+instance Eq TypeVar where
+  (TypeVar n1 _) == (TypeVar n2 _) = n1 == n2
+
+instance Ord TypeVar where
+  (TypeVar s1 _) `compare` (TypeVar s2 _) = s1 `compare` s2
+
+data Var = VarConst String
+     deriving (Show, Eq, Typeable, Ord)
+
+data Y = TypeVarY TypeVar
```
hypotext: `Y` is a confusing name. I know it's called that in the paper, but would rename to `SigNames` or `SigArgs` or something like that

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+
+-- --TODO: fix back...
+-- checkConstructorInvoker :: VarEnv -> ConstructorInvoker -> VarEnv
+-- checkConstructorInvoker e const = e
+
+checkConstructorInvoker :: VarEnv -> ConstructorInvoker -> VarEnv
+checkConstructorInvoker e const = let name = nameCons const
+                                      args = argCons const
+                                      env1 = foldl checkArg e args
+                                      kls1 = getTypesOfArgs e args
```
hypotext: name `kls1` -> `kinds`?

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+-- | Environment for the dsll semantic checker. As the 'check' function executes, it
+-- accumulate information such as symbol tables in the environment.
+
+-- | list of elements that might appear in the global context
+
+data Ttype = Ttype { yt :: Y,
+                     kt :: K }
+             deriving (Show, Eq, Typeable)
+
+data TypeConstructor = TypeConstructor { nametc :: String,
```
hypotext: suggest using underscore everywhere (`name_tc`, `name_vc`, `name_pred1`, etc)

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+
+instance Show ConstructorInvoker where
+  show (ConstructorInvoker nameCons argCons posCons) = nString ++ "(" ++ aString ++ ")"
+        where nString = show nameCons
+              aString = show argCons
+
+instance Eq ConstructorInvoker where
+  (ConstructorInvoker n1 a1 _) == (ConstructorInvoker n2 a2 _) = (n1 == n2 && a1 == a2)
+
+data Arg = AVar Var
```
hypotext: maybe soon we want to allow Arg to include anonymous values

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+checkAndGet k m pos = case M.lookup k m of
+  Nothing -> Left ("Error in " ++ (sourcePosPretty pos) ++ " : " ++ k ++ " Doesn't exist in the context \n")
+  Just v ->  Right v
+
+lookUpK :: VarEnv -> Arg -> K
+lookUpK e (AT  (TTypeVar t))  = (Ktype ((typeVarMap e) M.! t)) --(Ktype (Type {typeName = "type", typePos = typeVarPos t }))
+lookUpK e (AT  (TConstr t)) =
+        if (nameCons t) `elem` (declaredNames e)
+        then lookUpK e (AVar (VarConst (nameCons t)))
+        else (Ktype (Type { typeName = "type", typePos = constructorInvokerPos t }))
```
hypotext: does this handle the case where the type name is not legitimate -- it is neither a type variable (e.g. `X`) nor a nullary type ctor (e.g. `Set`)? it could be, say, `lsdhfsjdlfsd`, which should throw an error

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+second :: (a, b) -> b
+second (a, b) = b
+
+checkAndGet :: String -> M.Map String v -> SourcePos -> Either String v
+checkAndGet k m pos = case M.lookup k m of
+  Nothing -> Left ("Error in " ++ (sourcePosPretty pos) ++ " : " ++ k ++ " Doesn't exist in the context \n")
+  Just v ->  Right v
+
+lookUpK :: VarEnv -> Arg -> K
+lookUpK e (AT  (TTypeVar t))  = (Ktype ((typeVarMap e) M.! t)) --(Ktype (Type {typeName = "type", typePos = typeVarPos t }))
```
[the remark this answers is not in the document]
dormaayan: @hypotext , sure 
I simply hadn't a chance yet to accumulate all the errors. There is an issue for that

src/Dsll.hs
```diff
@@ -0,0 +1,322 @@
...
+  colon
+  p' <- propParser
+  return (Pd2Const (Pd2 { namePd2 = name, propsPd2 = (zip b' prop'), toPd2 = p' }))
+
+--------------------------------------- DSLL Semantic Checker ---------------------------
+
+-- | 'check' is the top-level semantic checking function. It takes a DSLL
+-- program as the input, checks the validity of the program acoording to the typechecking rules, and outputs
+-- a collection of information.
+check :: DSLLProg -> VarEnv
```
hypotext: give the various check functions more descriptive names, like `checkWithEnv_DSLL`

src/Dsll.hs
```diff
@@ -0,0 +1,322 @@
...
+                               temp = checkT localEnv res
+                               vc = VarConstructor { namevc = nameVd v, ylsvc = firsts (varsVd v),
+                                                     klsvc = seconds (varsVd v), tlsvc = seconds (typesVd v),
+                                                     tvc = toVd v }
+                               ef = addName (nameVd v) e
+                            in if ((env2 == e || env2 /= e) && (temp == e || temp /= e))
+                               then ef { varConstructors = M.insert (nameVd v) vc $ varConstructors ef }
+                               else error ("Error!") -- Does not suppose to reach here
+
+checkOperations :: VarEnv -> Od -> VarEnv
```
hypotext: don't forget to rename to operator

src/Dsll.hs
```diff
@@ -0,0 +1,322 @@
...
+  p' <- propParser
+  return (Pd2Const (Pd2 { namePd2 = name, propsPd2 = (zip b' prop'), toPd2 = p' }))
+
+--------------------------------------- DSLL Semantic Checker ---------------------------
+
+-- | 'check' is the top-level semantic checking function. It takes a DSLL
+-- program as the input, checks the validity of the program acoording to the typechecking rules, and outputs
+-- a collection of information.
+check :: DSLLProg -> VarEnv
+check p =  let env1  = foldl checkTypeConstructors initE (cd p)
```
hypotext: check if there are multiple declarations of the same name (tycon, valcon, ops, preds, or across any of those) -- and throw an error if so

src/Dsll.hs
```diff
@@ -0,0 +1,322 @@
...
+                                  varMap = M.empty, names = [], declaredNames = [], errors = ""}
+
+checkTypeConstructors :: VarEnv -> Cd -> VarEnv
+checkTypeConstructors e c = let kls  = seconds (inputCd c)
+                                env1 = foldl checkK e kls
+                                tc   = TypeConstructor { nametc = nameCd c, klstc = kls, typtc = outputCd c }
+                                ef   = addName (nameCd c) env1
+                             in ef { typeConstructors = M.insert (nameCd c) tc $ typeConstructors ef }
+
+checkVarConstructors :: VarEnv -> Vd -> VarEnv
```
hypotext: -> val constructors

src/Dsll.hs
```diff
@@ -0,0 +1,322 @@
...
+  return (Pd2Const (Pd2 { namePd2 = name, propsPd2 = (zip b' prop'), toPd2 = p' }))
+
+--------------------------------------- DSLL Semantic Checker ---------------------------
+
+-- | 'check' is the top-level semantic checking function. It takes a DSLL
+-- program as the input, checks the validity of the program acoording to the typechecking rules, and outputs
+-- a collection of information.
+check :: DSLLProg -> VarEnv
+check p =  let env1  = foldl checkTypeConstructors initE (cd p)
+               env2  = foldl checkVarConstructors env1 (vd p)
```
hypotext: do we check that all the type names used by val/op/pred declarations have already been defined by the type constructors? if not, we should

src/SubstanceCore.hs
```diff
@@ -0,0 +1,255 @@
...
+data Type = ApplyT SubType [Arg]
+          | TypeT TypeVar
+          deriving (Show, Eq, Typeable)
+
+data Arg = VarA Var
+         | TypeA Type
+         deriving (Show, Eq, Typeable)
+
+data Expr = VarE Var
+          | ApplyV Operator' [Expr]
```
hypotext: would rename to ApplyF

src/SubstanceCore.hs
```diff
@@ -0,0 +1,255 @@
...
+          | ApplyC ValConstructor [Expr]
```
hypotext: would rename to ApplyV

src/SubstanceCore.hs
```diff
@@ -0,0 +1,255 @@
...
+-- | Program is a sequence of statements
+type SubProg = [SubStmt]
+type SubObjDiv = ([SubDecl], [SubConstr])
+
+-- | Declaration of Substance objects
+data SubDecl = SubDeclConst Type Var
+               deriving (Show, Eq, Typeable)
+
+-- | Declaration of Substance constaints
+data SubConstr = SubConstrConst Predicate [Var]
```
hypotext: would remove mentions of Constr, replace with Pred or Predicate

src/SubstanceCore.hs
```diff
@@ -0,0 +1,255 @@
...
+          deriving (Show, Eq, Typeable)
+
+data SubStmt = Decl Type Var
+             | Bind Var Expr
+             | ApplyP Predicate [Var]
+             deriving (Show, Eq, Typeable)
+
+-- | Program is a sequence of statements
+type SubProg = [SubStmt]
+type SubObjDiv = ([SubDecl], [SubConstr])
```
hypotext: not sure if we need this anymore -- might be able to remove

src/SubstanceCore.hs
```diff
@@ -0,0 +1,255 @@
...
+-- Identifiers handling
+identifier :: Parser String
+identifier = (lexeme . try) (p >>= check)
+  where
+    p       = (:) <$> letterChar <*> many alphaNumChar
+    check x = if x `elem` rws
+                then fail $ "keyword " ++ show x ++ " cannot be an identifier"
+                else return x
+
+--------------------------------------- DSLL Parser -------------------------------------
```
hypotext: this is the Substance parser, right?

src/SubstanceCore.hs
```diff
@@ -0,0 +1,255 @@
...
+
+-- | 'substanceParser' is the top-level parser function. The parser contains a list of functions
+--    that parse small parts of the language. When parsing a source program, these functions are invoked in a top-down manner.
+substanceParser :: Parser [SubStmt]
```
hypotext: maybe clarify that the Substance parser does not depend on the DSLL info, just the Substance typechecker does

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
-check p = let env1 = foldl checkDecls initE p
-              env2 = foldl checkReferencess env1 p
-              in
-            env2 { subObjs = reverse $ subObjs env2 }
-          where initE = SubEnv { subObjs = [], subDefs = M.empty,  subSymbols = M.empty, subApps = [], subArgs = M.empty }
-
-applyDef (n, m) d = case M.lookup n d of
+-- AST as the input and the environment recived from the DSLL type checker, checks the validity of the program acoording to
+-- the typechecking rules, and outputs a collection of information.
+check :: SubProg -> VarEnv -> VarEnv
```
hypotext: would also name this more descriptively

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+check p varEnv = let env = foldl checkSubStmt varEnv p
+                 in if (null (errors env))
+                    then env
+                    else error ("Substance type checking failed with the following problems: \n" ++ (errors env))
+
+checkSubStmt :: VarEnv -> SubStmt -> VarEnv
```
hypotext: add a comment somewhere describing how the various env functions handle multiple declarations with same name, and multiple bindings with same name

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+
+isRecursedPredicate :: PredArg -> Predicate
+isRecursedPredicate (PP p) = p
+isRecursedPredicate (PE p) = error "Mixed predicate types!"
+
+checkRecursePred :: VarEnv -> [PredArg] -> VarEnv
+checkRecursePred varEnv args = let predArgs = map isRecursedPredicate args
+                               in foldl checkPredicate varEnv predArgs
+
+checkExpression :: VarEnv -> Expr -> (String, Maybe T)
```
hypotext: to clarify the type signature, I would do `type Error = String` and replace most of these `Strings` with errors

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+                 in if (null (errors env))
+                    then env
+                    else error ("Substance type checking failed with the following problems: \n" ++ (errors env))
+
+checkSubStmt :: VarEnv -> SubStmt -> VarEnv
+checkSubStmt varEnv  (Decl t (VarConst n)) = let env  = checkT varEnv t
+                                                 env1 = addDeclaredName n env
+                                             in env1 { varMap = M.insert (VarConst n) t $ varMap env1 }
+
+checkSubStmt varEnv  (Bind v e) = let (vstr, vt) = checkVarE varEnv v
```
hypotext: I guess we need a second environment to store the variable bindings

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+emptyArgList :: Parser [Arg]
+emptyArgList = do
+  lparen
+  rparen
+  return []
+
+----------------------------------- Typechecker aux functions ------------------------------------------
+
+firsts :: [(a, b)] -> [a]
+firsts xs = [x | (x,_) <- xs]
```
hypotext: use`map fst`

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+
+seconds :: [(a, b)] -> [b]
+seconds xs = [x | (_,x) <- xs]
```
hypotext: use `map snd`

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+
+second :: (a, b) -> b
+second (a, b) = b
```
hypotext: use `snd` everywhere

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+
+checkSubStmt varEnv  (ApplyP p) = checkPredicate varEnv p
+
+checkPredicate :: VarEnv -> Predicate -> VarEnv
+checkPredicate varEnv (Predicate (PredicateConst p) args pos) =
+               case checkAndGet p (predicates varEnv) pos of
+               Right p -> case p of
+                 Pred1 p1 -> checkVarPred varEnv args p1
+                 Pred2 p2 -> checkRecursePred varEnv args
+               Left _ -> case checkAndGet p (operations varEnv) pos of
```
hypotext: I don't understand why checkPred ignores the first error and then calls checkAndGet again with *operations*. preds and ops shouldn't be mixed, right?

src/Env.hs
```diff
@@ -0,0 +1,320 @@
...
+                             klsop  :: [K],
+                             tlsop  :: [T],
+                             top  :: T}
+                 deriving (Show, Eq, Typeable)
+
+data PredicateEnv = Pred1 Predicate1
+                  | Pred2 Predicate2
+                  deriving (Show, Eq, Typeable)
+
+data Predicate1 = Predicate1 { namepred1 :: String,
```
hypotext: confusing to use the same name twice -- name the constructor `Pred1` (and same for `Pred2`)

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+                 Right o -> checkVarOperation varEnv args o
+                 Left err -> varEnv { errors = (errors varEnv) ++ err }
+
+checkVarPred :: VarEnv -> [PredArg] -> Predicate1 -> VarEnv
+checkVarPred varEnv args (Predicate1 name yls kls tls _) =
+             let exprArgs      = map isVarPredicate args
+                 errAndTypesLs = map (checkExpression varEnv) exprArgs
+                 errls         = map (\(err1, t1) -> err1) errAndTypesLs
```
hypotext: use `firsts` everywhere (including the below functions, which have similar structure)

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+                 err           = foldl (\err1 err2 -> err1 ++ err2) "" errls
```
hypotext: replace with `concat`

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+                 argTypes      = map (\(err1, t1) -> t1) errAndTypesLs
```
hypotext: use `seconds` everywhere

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+             in if (foldl (\b at1 -> b && isJust at1) True argTypes)
```
hypotext: what is this condition trying to test? please refactor into a descriptively named function

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+                then let argTypes2 = map (\a -> KT (fromJust a)) argTypes
+                         tls2      = map (\a -> KT a) tls
+                         sigma     = subst varEnv M.empty argTypes2 tls2
+                     in if (sigma == M.empty)
+                        then varEnv { errors = (errors varEnv) ++ err } -- err should be empty str
```
hypotext: we have three of the same return value (lines 208-211). why? is there a TODO here, or can you refactor the function?

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+           in TConstr (ConstructorInvoker t argsSub pos)
+
+applySubstHelper :: M.Map Y Arg -> Arg -> Arg
+applySubstHelper sigma (AVar v) = case sigma M.! (VarY v) of
+                                  res@(AVar v2) -> res
+                                  AT t -> error "Var being mapped to a type in subst sigma, error in the TypeChecker!"
+applySubstHelper sigma (AT t) = AT (applySubst sigma t)
+
+subst :: VarEnv -> M.Map Y Arg -> [K] -> [K] -> M.Map Y Arg
+subst varEnv sigma argTypes formalTypes = let types = zip argTypes formalTypes
```
hypotext: what is a "formal type"?

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+                                              sigma2 = foldl (substHelper varEnv) sigma types
+                                          in sigma2
+
+substHelper :: VarEnv -> M.Map Y Arg -> (K, K) -> M.Map Y Arg
```
hypotext: give the 3 helpers more descriptive names, like substType and substArg

src/Substance.hs
hypotext: need to check that argTypes and formalTypes have the same length (zip throws away longer elements) -- don't forget to check argument lengths in the typechecker too

hypotext: would be super helpful to have just one or two concrete examples of a substitution as comments here
(e.g. what is unified? what doesn't match?)

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+substHelper varEnv sigma ((Ktype aT), (Ktype fT)) = sigma
```
hypotext: reorder the pattern match cases so all the failures are caught at the bottom

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+substHelper varEnv sigma ((KT (TConstr (ConstructorInvoker atc argsAT pos1))), (KT (TConstr (ConstructorInvoker ftc argsFT pos2)))) =
+                   if ((atc `elem` (declaredNames varEnv)) || (ftc `elem` (declaredNames varEnv)))
+                   then substHelper2 varEnv sigma ((AVar (VarConst atc)), (AVar (VarConst ftc)))
+                   else if (atc /= ftc) 
+                        then error ("Argument type " ++ (show atc) ++ " doesn't match expected type " ++ (show ftc))
+                        else let args   = zip argsAT argsFT
+                                 sigma2 = foldl (substHelper2 varEnv) sigma args
+                              in sigma2
+
+substHelper2 :: VarEnv -> M.Map Y Arg -> (Arg, Arg) -> M.Map Y Arg
```
hypotext: reorder the pattern match cases so all the failures are caught at the bottom

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+                   error ("Argument type " ++ (show aT) ++ " doesn't match expected type " ++ (show fT))
+substHelper varEnv sigma ((KT aT), (Ktype fT)) =
+                   error ("Argument type " ++ (show aT) ++ " doesn't match expected type " ++ (show fT))
+substHelper varEnv sigma ((KT (TTypeVar atv)), (KT (TTypeVar ftv))) =
+                   substInsert sigma (TypeVarY ftv) (AT (TTypeVar atv))
+substHelper varEnv sigma ((KT (TTypeVar atv)), (KT (TConstr (ConstructorInvoker ftc argsFT pos)))) =
+                   error ("Argument type " ++ (show atv) ++ " doesn't match expected type " ++ (show ftc))
+substHelper varEnv sigma ((KT (TConstr (ConstructorInvoker atc argsAT pos))), (KT (TTypeVar ftv))) =
+                   substInsert sigma (TypeVarY ftv) (AT (TConstr (ConstructorInvoker atc argsAT pos)))
+substHelper varEnv sigma ((KT (TConstr (ConstructorInvoker atc argsAT pos1))), (KT (TConstr (ConstructorInvoker ftc argsFT pos2)))) =
```
hypotext: this case could be refactored:
* put the simple/successful if/then cases at the top
* i'm confused about the logic here. are atc and ftc allowed to be type variables? what if one of them is a variable and the other is a type? either way, please write out the logic here first, since it could get tricky

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+--                                   tType -> Left tType
+--argTypeLookupHelper varEnv (VarY v) = case M.lookup v (varMap varEnv) of
+--                               Nothing -> Right "Argument " ++ (show v) ++ " not in environment\n"
+--                               vType -> Left vType
+
+
+-- --------------------------------------- Substance Loader --------------------------------
+-- | Load all the Substance objects for visualization at Runtime.hs
+
+data SubObjects = SubObjects { subObjs :: [SubObj] }
```
hypotext: if you're not adding more fields, remove the datatype and just use `[SubObj]` everywhere (and simplify all uses of this type below)

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
-                           case line of
-                           (LD x) -> (x : decls, constrs)
-                           (LC x) -> (decls, x : constrs)
+              where separate line (decls, constrs) =
+                             case line of
+                             (LD x) -> (x : decls, constrs)
+                             (LC x) -> (decls, x : constrs)
 
+loadObjects :: SubProg -> VarEnv -> SubObjects
+loadObjects p subEnv = let env1 = foldl (passDecls subEnv) initObjs p
```
hypotext: misleading name -- [SubObjs] is not an environment

hypotext: I see you just did a big refactor of Nimo's code here, but I'm sure it could be made much simpler still :-)
for example:
- the subEnv is only used once (in declaredNames)
- checkAndInsert and checkNameAndTyp are not used anywhere in the codebase
- not sure that Style really needs us to do this `SubObj` conversion, since it's mostly putting the same info into a different container

src/Substance.hs
```diff
@@ -4,535 +4,450 @@
...
+fixArg :: VarEnv -> Arg -> Arg
+fixArg e (AT (TConstr i)) = if (nameCons i) `elem` (declaredNames e)
+                            then AVar (VarConst (nameCons i))
+                            else AT (TConstr i)
+fixArg e a = a -- Ignore all other cases
+
+-- | 'checkReferencess' checks any statement that refers to other objects. For example,
+-- | > Subset A B
+-- | refers to identifiers @A@ and @B@. The function will perform a lookup in the symbol table, make sure the objects exist and are of desired types -- in this case, 'Set' -- and throws an error otherwise.
+passReferencess :: SubObjects -> SubStmt -> SubObjects
```
hypotext: needs a more accurate name and comment after the refactor

src/Style.hs
```diff
@@ -311,11 +313,10 @@ data StySpec = StySpec {
 type VarMap  = M.Map Name Name
 
 initSpec :: StySpec
-initSpec = StySpec { spType = C.PointT,
+initSpec = StySpec { spType = (TypeNameConst "PointT" ),
```
hypotext: i'm pretty sure PointT is no longer a type, maybe just Point?

src/Style.hs
```diff
@@ -562,23 +564,40 @@ procAssign _ spec _ = spec -- TODO: ignoring assignment for all others; what kin
...
+    where conv c = case c of
+            AVar (VarConst s) -> s
+            _ -> ""
+
+exprToString :: [C.Expr] -> [String]
+exprToString = map conv
+    where conv c = (show c)
+
+-- TODO: document how predicates are named
+getConstrTuples :: [C.SubConstr] -> [(TypeName, String, [String])]
```
hypotext: give more descriptive type synonyms for the strings

hypotext:
Thanks for writing this, @dormaayan and @jennalwise! It generally looks great. I mostly had suggestions about more descriptive naming and simplifying the code in the inline comments.

Some general notes:
- I didn't review the LA shapes and Style for this PR
- Don't forget to update the LA book
- Please follow the standard Haskell whitespace/parenthesization conventions (you can look at my commits for examples)
- Could one of you add Jenna's [documentation](https://docs.google.com/document/d/1xZwkqKDRqTzIIJ_YgS8tKGeVM6hxv3CcbE9uKM54SCs/edit) inline or on the wiki? 
- It would be nice if the typechecker functions were annotated with the relevant semantics rule(s) from the paper
- What existing tests do we have for the DSLL and Substance parser and typechecker? What tests are you planning to add?

src/Substance.hs
dormaayan: @hypotext : If you come up with a better interface for connecting Substance and Style while you implement Style let me know and I'll do the Substance work

src/Substance.hs
dormaayan: @hypotext : Yes, we probably do. As agreed, I'll do that after Style is implemented

src/Env.hs
dormaayan: But they are not part of the AST, they should remain where they are

src/Substance.hs
dormaayan: In fact might be extended in the really near future....

src/Substance.hs
dormaayan: Again might change after style is implemented

hypotext:
@dormaayan Cool, thanks for making the changes. Was there anything significant that you didn't change?

You can go ahead and merge this. Let me know if there are any nontrivial conflicts that need to be resolved.

dormaayan:
@hypotext : Just Merged, it is all on master and seems like working after resolving those conflicts

Issue #91, opened 2018-06-06
Add property access support for Style


dormaayan:
Currently, style does not allow nested access to properties of shapes, for example, the following program will not work:
`Neg v {
        shape = Arrow {
        start = v.shape.start
    }
}` will not work

the desired syntax, as already discussed, is:
`<object>.<shape-name>.<property>`

wodeni:
This change will involve `Runtime`, `Server`, and `Style`. We outline the changes needed here:

- [ ] Generalize `ObjFn` type to take in a list of `TypeIn`
  - Should we also keep the list of float `[a]`?
- [ ] In `Style`, there are several places where property access can happen
  - Objective function/constraint arguments
  - Computation arguments
  - Inside of constructors
- [ ] Runtime gets
- [ ] TODO: how about labels? `A.shape.label.text` vs `A.label.text`

dormaayan:
@wodeni: Great, Thanks! I'll merge it to my plugin branch

hypotext:
I can review the changes today. Do you mind waiting for a review before
merging?

dormaayan:
@hypotext : I already merged and resolved conflicts...

hypotext:
Ok, @wodeni could you open a PR with your branch first? I’ll review there.

@dormaayan if Nimo or I need to make any changes, I guess we can push to
Nimo’s branch and you can pull onto yours when done.

In general, for larger system refactors, I’d wait for at least one person
to review code before merging it :)

dormaayan:
@hypotext : Sounds good to me

PR #92, opened 2018-06-12
Property access

wodeni:
- Relevant issue: #91 
- I made the base branch `real-analysis` since I branched off from there, so that y'all can see the changes more clearly. 
- Now all parity tests work with property access added. Some examples of new programs we can write:
# TODOs
- [x] When to resolve objs and properties?
- [x] I deleted all the computation related lines in `lookupvarmap`
- [x] document `_get` 
- [x] run through all previous examples and parity tests
- [x] Clear up the COMBAKs

# Design decisions

- Used a built-in computation `_get` to model property access in the following places
  - RHS of assignment statements within constructors
  - Arguments to objectives and constraints
  - Arguments to computations
- Forbid nested computation: `f(f1(obj.x), f2(obj.y))` will not be allowed
- Forbid property access of a computation: `f(x).head` either isn't allowed or won't parse (?)
- Replaced `Either a String` everywhere with `TypeIn a`
- `TypeIn a` now has `TAllShapes`, `TShape`, `TProp` that models pending access to GPIs and properties
  - TODO: maybe objects should be resolved earlier, but this has been the design for a long time. I remember the main purpose is to avoid relying on the order of Style program

# Example programs and output
## Equal size sets
- run: `penrose snap sub/twosets.sub sty/same-size-circs.sty dsll/setTheory.dsl`
```
Set A, B
-------------
Set `A` {
    shape = Circle { }
    constraint contains(A, A.shape.label)
}

Set `B` {
    shape = Circle {
        radius = A.shape.radius
    }
    constraint contains(B, B.shape.label)
}
```

## `Neg`: negation of a vector in LA domain
- run `penrose snap sub/vectorsNegation.sub sty/vectors.sty  dsll/algebra.dsl`
- Substance:
```
VectorSpace U
Vector(U) A
Neg(A)
```
- Style: 
```
-- only showing the relevant block
Neg v {
    shape = Arrow {
        startx= v.shape.start
        end = reverseEndArrow(v.shape)
    }
}
```
neg

# Other notes
- No shortcut to shapes using just the Substance identifier in `procExpr`
- What does nested computation mean?
- `resolveObjs` has the synonym logic
- For now, the order of the Substance objects determines the order in which the computations are found and applied (it's the same order). Putting Substance objects out of order may mess up the computational graph


hypotext:
Are there any major design decisions, TODOs, or problems I should know about before reviewing? Also, could you link the PR to issue #91?

hypotext:
I will review this today

src/Style.hs
```diff
@@ -470,144 +469,73 @@ procObjFn :: (Autofloat a) =>
...
--- TODO: or a list of all shapes: "A xaxis", "A yaxis", ... (lookupAll does this for now)
-procExpr ctx (Id subObjPattern) = Left $ lookupVarMap subObjPattern ctx
-
--- "ID.label" is a synonym for "ID.shape.label"
-procExpr ctx r@(BinOp Access (Id subObjName) (Id "label")) = error ("cannot access label of non-shape:\n" ++ show r)
-
+procExpr :: (Autofloat a) => VarMap -> Expr -> TypeIn a
+procExpr ctx (Id subObjPattern) = TAllShapes $ lookupVarMap subObjPattern ctx
+procExpr _ (StringLit s) = TStr s
+procExpr _ (IntLit i) = TNum $ r2f i -- TODO this shouldn't flatten ints for computations
```
hypotext: we have `TInt` now, so let's use that tag

src/Style.hs
```diff
@@ -470,144 +469,73 @@ procObjFn :: (Autofloat a) =>
...
-procExpr ctx (BinOp Access (BinOp Access (Id subObjPattern) (Id styShapeName)) (Id "label")) = 
-             let subObjName = lookupVarMap subObjPattern ctx in
-             Left $ labelName $ uniqueShapeName subObjName styShapeName
-
--- in context [X ~> A], look up "X.yaxis", return "A yaxis"
-procExpr ctx (BinOp Access (Id subObjPattern) (Id styShapeName)) = 
-         let subObjName = lookupVarMap subObjPattern ctx in
-         Left $ uniqueShapeName subObjName styShapeName
-
+-- TODO: pending discussion on `A.shape.label` vs `A.label`
```
hypotext: don't forget to update the labeling comments

src/Runtime.hs
```diff
@@ -54,28 +54,23 @@ data Params = Params { weight :: Double,
...
-                     oProp :: Name, -- "radius"
-                     fnName :: Name, -- computeRadius
-                     fnParams :: [S.Expr] } -- (1.2, B)
-               deriving (Show, Typeable)
-
--- | fn name, list of args (returned by queryCondig)
-type CompInfo = (Name, [S.Expr])
+-- | object name, function name, list of args
+data ObjComp a = ObjComp {
+    oName    :: Name,      -- "A"
```
hypotext: I think this `ObjComp` thing (which I originally put in for property-setting like `prop = f(x)`) is better as a suspended `set` (just like `A.shape.prop` has become a suspended `get`). That will unify our approaches

src/Shapes.hs
```diff
@@ -780,16 +673,25 @@ data TypeIn a = TNum a
...
               | TStr String
               | TInt Integer
+              | TFloat Float
               | TPt (Pt2 a)
               | TPath [Pt2 a]
-              | TArrow SolidArrow
               | TColor Color
-              | TStyle String -- dotted, etc.
+              | TStyle String -- | dotted, etc.
+              | TAllShapes String -- | Substance ID
```
hypotext: kind of a confusing name, can we call this something like `TSubObjShapes`?

src/Functions.hs
```diff
@@ -7,13 +7,15 @@ import Utils
...
 import qualified Data.Map.Strict as M
 
-type ObjFnOn a = [Obj' a] -> [a] -> a
-type ConstrFnOn a = [Obj' a] -> [a] -> a
-type ObjFn = forall a. (Autofloat a) => [Obj' a] -> [a] -> a
-type ConstrFn = forall a. (Autofloat a) => [Obj' a] -> [a] -> a
-type ObjFnInfo a = (ObjFnOn a, Weight a, [Name], [a])
-type ConstrFnInfo a = (ConstrFnOn a, Weight a, [Name], [a])
-type Weight a = a
+type ObjFnOn    a = [Obj' a] -> [TypeIn a] -> a
```
hypotext: are we still using the On types? can we remove them?

src/Functions.hs
```diff
@@ -7,13 +7,15 @@ import Utils
...
+type ConstrFnOn a = [Obj' a] -> [TypeIn a] -> a
+type ObjFn    = forall a. (Autofloat a) => [Obj' a] -> [TypeIn a] -> a
```
hypotext: I'm concerned about whether objfns get arguments in the right order as specified in Style. Can we document how the ordering is preserved or not preserved?

src/Computation.hs
```diff
@@ -224,6 +227,56 @@ regionX' :: CompFn a
...
+
+endArrow' :: CompFn a
+endArrow' [] [A' a] = TPt((endx' a),(endy' a))
+
+reverseEndArrow' :: CompFn a
+reverseEndArrow' [] [A' a] = let sx = (startx' a)
+                                 sy  = (starty' a)
+                                 ex = (endx' a)
+                                 ey = (endy' a)
+                              in TPt(sx - (ex - sx), sy - (ey - sy))
```
hypotext: rename function to something generic like `negV`
can simplify last expr to just `(-ex, -ey)`

src/Computation.hs
```diff
@@ -224,6 +227,56 @@ regionX' :: CompFn a
 regionX' [] [LN' lineLeft, LN' lineRight] = TNum $ regionX lineLeft lineRight
 regionX' v o = error' "regionX" v o
 
+-- returns the middle of the square as a starting position
+midSquare' :: CompFn a
+midSquare' [] [S' b] = TPt((xs' b),(ys' b))
```
hypotext: maintain spacing everywhere and avoid unnecessary parens: TPt (xs' b, ys' b)

src/Functions.hs
```diff
@@ -193,7 +184,7 @@ _centerMap a s1@[x1, y1] s2@[x2, y2] [o1, o2] =
     (fromx - sx)^2 + (fromy - sy)^2 + (tox - ex)^2 + (toy - ey)^2
 
 centerLine :: ObjFn
-centerLine [LN' l, P' p1, P' p2] _ = 
+centerLine [LN' l, P' p1, P' p2] _ =
```
hypotext: if a function's expecting no params, better to use [] instead of _ so an error happens

src/Runtime.hs
```diff
@@ -275,28 +270,32 @@ computeOn objDict comp =
...
+-- TODO: complete documentation
+resolveObjs :: (Autofloat a) => M.Map Name (Obj' a) -> TypeIn a
+                             -> Either (TypeIn a) [Obj' a]
+resolveObjs objs e = case e of
+    -- TODO: type synonyms logic here
+    TAllShapes v   -> case lookupAll v objs of
+                      [] -> error ("id '" ++ v ++ "' /and subobjects do(es) not exist in obj dict")
+                      xs -> Right xs
+    TShape     v   -> case M.lookup v objs of
+                      Nothing -> error ("id '" ++ v ++ "' does not exist in obj dict: " ++ show e)
```
hypotext: I prefer to put "Nothing" cases after the "Just" cases (but I recognize that's inconsistent with the list treatment above)

hypotext:
Looks good overall, thanks for doing this! 

Some notes:
* Can you briefly fill in the design decisions section above?
* I'm trying to repro the examples on `master` and am not able to do either of the two examples because of repo organization. Can we restore the linear-algebra folder? and where did the same-size-circs style program go?
* I'd like to see more working examples, especially for tricky cases like properties being accidentally set to themselves, properties being mutually set to each other, or lots of layers of indirection.
* I guess you'll make these changes on `master`?

src/Functions.hs
wodeni: Yes, since the `Info` types use them and they are actually used by Runtime

src/Style.hs
wodeni: WIll propagate that change in Computations.  BTW what do you mean by `-- TODO this shouldn't flatten ints for computations`?

src/Shapes.hs
wodeni: Now that I think about it, I don't see any motivation for having this shortcut to access all objects. All other parts of this feature guarantee correct ordering, but if you have something like `func(A, B.circ)`, there's no way to tell how many arguments this function is taking without looking at the Style block(s) for `A`. Do you think we should still keep this feature?

src/Functions.hs
wodeni: Noted in documentation of `applyAndSet`

src/Computation.hs
wodeni: Again, we should all use Hlint 😄

wodeni:
# Example: properties as arguments
- run: `penrose snap sub/twosets.sub sty/half-size-circs.sty dsll/setTheory.dsl`
- Style file:
```
Set `A` {
    shape = Circle { }
    constraint contains(A, A.shape.label)
}
Set `B` {
    shape = Circle { }
    constraint contains(B, B.shape.label)
    objective equal(A.shape.x, B.shape.x)
    objective ratioOf(A.shape.radius, B.shape.radius, 0.5)
}
```
- `equal` is a generic function penalizing the difference of two `TNum`s
- `ratioOf` is similar but multipies the second number by the third argument
- As a result, the circles always have the same x coordinates and the radio of their radii is 1 : 2
- Screenshot:
half-circ

wodeni:
By modifying `same-size-circs.sty`, I tested for two tricky cases
- properties being accidentally set to themselves: no error happens if ```Set `A` { shape = Circle { radius = A.shape.radius } }```. There is no effect because we don't perform `get` until the Style program is completely analyzed
- properties being mutually set to each other: for the same reason as the previous case,  the properties will just be equal and there will not be any scoping issue

wodeni:
Filled in the major design decisions. Maybe I should switch the base branch to `master` and see if I can merge

hypotext:
Looks like I accidentally started a formal review earlier. I just ended it (those comments are old)

src/Shapes.hs
hypotext: I agree with the ordering/argument problems. Let's remove the "GPI synonym feature" for now.

hypotext:
Thanks for addressing the comments! Most things look good. The design decisions section is especially helpful. I edited your PR doc slightly.

Two things:
* I'm trying to repro, so I pulled your changes, but `same-size-circs.sty` isn't in my `sty` folder (but my code contains the rest of your changes). Do you happen to know why that's happening?
* Do you plan to do the ObjComp refactoring? It would be much cleaner as a runtime `_set`, and to change `ObjComp` into a generally suspended computation. (I realize this was my original design, so I can fix it later if you prefer, it just might have to wait a few days)

src/Runtime.hs
```diff
@@ -430,22 +431,25 @@ getShape (oName, (objType, config)) =
...
-                             res             -> Just res
+compsAndVars :: (Autofloat a) =>  Name -> S.Properties a -> ([ObjComp a], S.Properties a)
+compsAndVars n props =
+    let (props_comps, props_nocomps) = M.partition isComp props
+        comps = mapVals $ M.mapWithKey packComp props_comps
+    in
+    (comps, props_nocomps)
+    where
+        packComp :: Property -> TypeIn a -> ObjComp a
+        packComp prop (TCall f args) = ObjComp { oName = n, oProp = prop, fnName = f, fnParams = args }
```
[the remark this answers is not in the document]
wodeni: I don't really see how to do it in a cleaner way, as I thought `ObjComp` _was_ a general suspended computation...

wodeni:
`same-size-circs.sty` is definitely tracked in the most recent commit 549864f. Make sure you are on the correct branch? Otherwise, I'm not sure what happened 😕

Issue #31, opened 2017-11-05
Deprecate gloss


hypotext:
Better to just support one frontend.

Issue #10, opened 2017-10-31
gloss freezes with sub/nested.sub and sty/venn_comp.sty


hypotext:
The initial exterior point weight seems very high. But the error doesn't happen in snap.

wodeni:
No longer using gloss now! (#93)

PR #93, opened 2018-06-12
Deprecate gloss

wodeni:
- Time to say bye to Gloss 😢 
- The first argument of `penrose` binary is now removed
- `Color` is replaced with a simplified version of Gloss' implementation
- `black` and `white` are the only built-in colors for now
- Seems that the long compilation time is __not__ caused by Gloss. Maybe `Typeable` (@hypotext let me know why we need it, if you remember anything about it)?? 
- Relevant issues:  #10 #31


hypotext:
Looks good overall! Are we sure that @lilyshellhammer won't be using gloss in the future? The graphics API design doc mentioned possibly using gloss to access pixels of an image.

I wouldn't be surprised if `Typeable` was causing long build times. We can remove it too. (I originally added it because I wanted to do reflection over datatype definitions for dynamically applying computations. We're no longer taking that approach.)

wodeni:
Checked with @lilyshellhammer. There's no particular reason to use `gloss` for accessing pixel values in images, so it's okay to deprecate it. Now is a good time to merge.

Issue #59, opened 2018-03-29
Supporting LaTeX labels and `Label` statements


wodeni:
# Proposed solution

* Allow LaTeX strings to be rendered in labels, e.g. `Label X $\sum_i^n\frac{x^2}{10}$
![image](https://user-images.githubusercontent.com/11740102/41427712-c4836992-6fd5-11e8-8f46-caa84c1d37e6.png)
* A labeling language, so Substance variables can be named in ASCII and given TeX labels, e.g.
```
Set A, B
f: A -> B
AutoLabel A, B
Label f $\sum_i^n \frac{x^2}{10}$
```
* We support three kinds of statements in our labeling DSL: 
  * `Nolabel xs`: forbids Substance objects from having labels. `xs` is a comma separated list of Substance identifiers.
  * `Label x s`: normal declaration of labels. Gives a Substance object with identifier `x` a label, where the content of the label is specified by `s`. `s` is a TeX expression in math environment (`"$f$"`) (TODO: do we want to have labels that are outside of math env as well, say `"hello world"`?)
  * `Autolabel xs`: automatically generates labels for all identifiers in `xs`. By default, a Substance id `a` will be translated into TeX math environment as `$a$`. 
    * Special keyword `All` can replace `xs`, which just refers to all Substance ids. 
* The semantics of label statements are determined by the ordering of them in the source file. 
  * If we have the following, `A` will end up not having any label. 
  ```
  Label A "AA"
  AutoLabel A
  NoLabel A
  ```
  
* In the system, we would add the two kinds of label declarations (manual vs. auto labeling) to the options of Substance statements.
```haskell
data SubStmt = {- all other Substance stmts -} | LabelDecl String String | AutoLabel [String]
```
*  __As a result, label statements can be interleaved with other Substance statements__. Thus, the following style is also supported:
```
Set A, B
AutoLabel A, B

f: A -> B
Label f $\sum_i^n \frac{x^2}{10}$
```
* In Style, a labeled Substance object, `X`,  will have a Graphical Primitive Instance (GPI) `X.text`, which is of type `Text`. If more than one shape is related to `X` and we want to assign labels to these objects, we can do:
```
Set A {
  component_1 = Circle {
      text = concat(A.text, "_1")
  }
  component_2 = Square {
      text = concat(A.text, "_2")
  }
}
```
* Do we want to have labels that are outside of math env as well, say `"hello world"`? If everything is by default in math env, then the following will be rendered as the label:
[...]

PR #94, opened 2018-06-20
JS refactoring, LaTeX label, and labeling DSL

wodeni:
# Description
Related issues: #59 
This PR includes changes to achieve: 
* JS refactoring using module pattern
* support for LaTeX rendering on the front end
   * Allow LaTeX strings to be rendered in labels, e.g. `Label X $\sum_i^n\frac{x^2}{10}$
image
* implementation of the label DSL
  * A labeling language, so Substance variables can be named in ASCII and given TeX labels, e.g.
```
Set A, B
f: A -> B
AutoLabel A, B
Label f $\sum_i^n \frac{x^2}{10}$
```

## Label DSL spec

* We support three kinds of statements in our labeling DSL: 
  * `Nolabel xs`: forbids Substance objects from having labels. `xs` is a comma separated list of Substance identifiers.
  * `Label x s`: normal declaration of labels. Gives a Substance object with identifier `x` a label, where the content of the label is specified by `s`. `s` is a TeX expression in math environment (`"$f$"`) 
  * `Autolabel xs`: automatically generates labels for all identifiers in `xs`. By default, a Substance id `a` will be translated into TeX math environment as `$a$`. 
    * Special keyword `All` can replace `xs`, which just refers to all Substance ids. 
* The semantics of label statements is determined by the ordering of them in the source file. 
  * If we have the following, `A` will end up not having any label. 
  ```
  Label A "AA"
  AutoLabel A
  NoLabel A
  ```
  
* In the system, we would add the two kinds of label declarations (manual vs. auto labeling) to the options of Substance statements.
```haskell
data SubStmt = {- all other Substance stmts -} | LabelDecl String String | AutoLabel [String]
```
*  __As a result, label statements can be interleaved with other Substance statements__. Thus, the following style is also supported:
```
Set A, B
AutoLabel A, B

f: A -> B
Label f $\sum_i^n \frac{x^2}{10}$
```
* In Style, a labeled Substance object, `X`,  will have a Graphical Primitive Instance (GPI) `X.text`, which is of type `Text`. If more than one shape is related to `X` and we want to assign labels to these objects, we can do:
```
Set A {
  component_1 = Circle {
      text = concat(A.text, "_1")
  }
  component_2 = Square {
      text = concat(A.text, "_2")
  }
}
```
* Do we want to have labels that are outside of math env as well, say `"hello world"`? If everything is by default in math env, then the following will be rendered as the label:
image
  * @hypotext: all label texts are double-quoted, in which TeX expressions are enclosed by dollar signs (`Label A "hello $f$"`). --> __Currently this is NOT supported and label text has to be enclosed by either quotes or dollar signs. The two cannot be mixed__ 
  * Current solution: `\textrm` works in MathJax!
* Label statements can be interleaved with other Substance statements?

- On the highest level, all labels are converted to SVG strings by MathJax __the first time the scene is rendered__ and will then be kept inside of a global dictionary called `label_svgs` indexed by `uniqueShapeName`
- Everytime `Render.scene` is called, `_renderLabel` will parse the strings into SVG fragments using Snap and place them onto canvas
- MathJax has an asynchronous API and requires queuing its "Queue" with typesetting requests. Therefore, I used Promises, `await` and `async` for all related functions (see `tex2svg` and `_collectLabels` for details)

# Examples
- `penrose sub/twosets.sub  sty/venn.sty  dsll/setTheory.dsl`
latex
- `penrose sub/continuousmap.sub  sty/continuousmap.sty  dsll/setTheory.dsl`
image
- `penrose sub/surjection_computed.sub  sty/surjection_computed.sty dsll/setTheory.dsl`
image

# Checklist

- [x] Minimum working example with MathJax
- [x] Refactor JS front-end code for better modularity and readability
- [x] Parsing and rendering basic labels (e.g.: `f`, `g`...) 
- [x] Supporting more complex Latex statements
- [x] Create labeling statements (`Label f $f$`)
- [x] Adding syntactic sugar for auto labels ( `AutoLabel f, g, h`)
- [x] I ran [Haddock](https://github.com/penrose/penrose/wiki/Writing-and-generating-documentation#generating-html-documentation-site) and there were no errors when generating the HTML site

# Open questions
- Gaining access to Bezier curves
- To solve this, we first have to know the best way to represent Bezier curves in our Runtime? Currently, we only represent them as a list of point, which could be insufficient for optimization purposes 
- We still need to parse the SVGs of TeX labels back into our Runtime representation in order to optimize them. Here's an SVG parser library I found: [svg-parser](https://github.com/elaye/svg-parser).

* In Style, how to design a mechanism for “don’t make a labeling objective in Style if A does not have a label”? Below are possible solutions:
  * `if x.labeled`, where `labeled` is a built-in field in a Substance object
  ```
  Set A {
    if A.labeled: labelFn = encourage -- some objective
  }
  ```
  * A built-in unary predicate `Labeled` in the core Substance language, so that we can write selectors for them:
  ```
  Set A 
  where Labeled(A) {    labelFn = encourage -- some objective }
  ```


hypotext:
@wodeni Can you add the PR description to the wiki? Thanks

Issue #85, opened 2018-06-05
Better Error Messages From the Type Checker


dormaayan:
Refers to both Substance and the DSLL Type Checkers
- [ ] Make sure that we get all the errors together at the end of the type-checking
- [ ] Make sure that all errors are informative with line numbers and column numbers
- [ ] Create a list of errors instead of a string of all the errors in the environment

wodeni:
Addressed in the ported version (#447, #441).

PR #98, opened 2018-06-27
Dor revisions implementation

dormaayan:
**This PR is not ready for review currently**
#Description

Related issues: # (issue)

# Checklist

# Open questions

### DIFF
--- a/src/Dsll.hs
+++ b/src/Dsll.hs
@@ -58,6 +58,16 @@ instance Show Vd where
               cString = show typesVd
               dString = show toVd
 
+-- | Subtype declarations
+data Std = Std {subType :: T,
+                superType :: T}
+          deriving (Eq, Typeable)
+
+instance Show Std where
+   show (Std subTypeStd superTypeStd) = aString ++ "Subtype of" ++ bString
+      where aString = show subTypeStd
+            bString = show superTypeStd
+
 -- | predicates
 data Od = Od { nameOd  :: String,
                varsOd  :: [(Y, K)],
@@ -107,15 +117,17 @@ instance Show Pd2 where
 
 -- | DSLL program composed out of type constructors followed by ar declarations, operations and predicates
 data DSLLProg = DSLLProg { cd :: [Cd],
+                           std :: [Std],
                            vd :: [Vd],
                            od :: [Od],
                            pd :: [Pd] }
                 deriving (Eq, Typeable)
 
 instance Show DSLLProg where
-    show (DSLLProg cd vd od pd) = "types = " ++ cdString ++ "\n \n" ++ "vars = " ++
+    show (DSLLProg cd std vd od pd) = "types = " ++ cdString ++ "\n \n" ++ "subtypes = " ++ stdString ++ "\n \n" ++ "vars = " ++
         vdString ++ "\n \n" ++ "operations = " ++ odString ++ "\n \n" ++ "predicates = " ++ pdString ++ "\n \n"
         where cdString = show cd
+              stdString = show std
               vdString = show vd
               odString = show od
               pdString = show pd
@@ -133,10 +145,11 @@ dsllParser = between scn eof dsllProg -- Parse all the statemnts between the spa
 dsllProg :: Parser DSLLProg
 dsllProg = do
     listCd <- cdParser `sepEndBy` newline'
+    listStd <- stdParser `sepEndBy` newline'
     listVd <- vdParser `sepEndBy` newline'
     listOd <- odParser `sepEndBy` newline'
     listPd <- pdParser `sepEndBy` newline'
-    return DSLLProg { cd = listCd, vd = listVd, od = listOd, pd = listPd }
+    return DSLLProg { cd = listCd, std = listStd, vd = listVd, od = listOd, pd = listPd }
 
 -- | type constructor parser
 cdParser, cd1, cd2 :: Parser Cd
@@ -155,6 +168,14 @@ cd2 = do
     colon
     t' <- typeParser
     return Cd { nameCd = name, inputCd = [], outputCd = t' }
+
+-- | sub type declarations parser
+stdParser :: Parser Std
+stdParser = do
+  subtype <- tParser
+  rword "<:"
+  supertype <- tParser
+  return Std { subType = subtype, superType = supertype}
 
 -- | parser for the (y,k) list
 ykParser :: Parser ([Y], [K])
@@ -218,16 +239,18 @@ pd2 = do
 -- program as the input, checks the validity of the program acoording to the typechecking rules, and outputs
 -- a collection of information.
 check :: DSLLProg -> VarEnv
-check p =  let env1  = foldl checkTypeConstructors initE (cd p)
-               env2  = foldl checkValConstructors env1 (vd p)
-               env3  = foldl checkOperators env2 (od p)
-               env4  = foldl checkPredicates env3 (pd p)
-           in if (null (errors env4))
-              then env4
-              else error ("DSLL type checking failed with the following problems: \n" ++ (errors env4))
+check p = let  env1  = foldl checkTypeConstructors initE (cd p)
+               env2  = foldl checkSubTypes env1 (std p)
+               env3  = computeSubTypes env2
+               env4  = foldl checkValConstructors env3 (vd p)
+               env5  = foldl checkOperators env4 (od p)
+               env6  = foldl checkPredicates env5 (pd p)
+           in if (null (errors env6))
+              then env6
+              else error ("DSLL type checking failed with the following problems: \n" ++ (errors env6))
            where initE = VarEnv { typeConstructors = M.empty, valConstructors = M.empty,
                                   operators = M.empty, predicates = M.empty, typeVarMap = M.empty,
-                                  varMap = M.empty, typeCtorNames = [], declaredNames = [], errors = ""}
+                                  varMap = M.empty, subTypes = [], typeCtorNames = [], declaredNames = [], errors = ""}
 
 checkTypeConstructors :: VarEnv -> Cd -> VarEnv
 checkTypeConstructors e c = let kinds  = seconds (inputCd c)
@@ -235,6 +258,19 @@ checkTypeConstructors e c = let kinds  = seconds (inputCd c)
                                 tc   = TypeConstructor { nametc = nameCd c, kindstc = kinds, typtc = outputCd c }
                                 ef   = addName (nameCd c) env1
                              in ef { typeConstructors = M.insert (nameCd c) tc $ typeConstructors ef }
+
+
+checkSubTypes :: VarEnv -> Std -> VarEnv
+checkSubTypes e s = let env1 = checkDeclaredType e (subType s)
+                        env2 = checkDeclaredType env1 (superType s)
+                        env3 = env2{subTypes = ((subType s),(superType s)) : (subTypes env2)}
+                        in env3
+
+computeSubTypes :: VarEnv -> VarEnv
+computeSubTypes e = let env1 = e { subTypes = (transitiveClosure (subTypes e))}
+                    in if (isClosureNotCyclic (subTypes env1)) then
+                      env1
+                    else env1 { errors = (errors env1) ++ "Cyclic Subtyping Relation! \n"}
 
 checkValConstructors :: VarEnv -> Vd -> VarEnv
 checkValConstructors e v = let kinds = seconds (varsVd v)
@@ -259,7 +295,7 @@ checkOperators e v = let kinds = seconds (varsOd v)
                          args = seconds (typesOd v)
                          res = toOd v
                          env2 = foldl checkT localEnv args
-                         temp = checkT localEnv res
+                         temp = checkT env2 res
                          op = Operator { nameop = nameOd v, ylsop = firsts (varsOd v),
                                            kindsop = seconds (varsOd v), tlsop = seconds (typesOd v), top = toOd v }
                          ef = addName (nameOd v) e
@@ -300,6 +336,9 @@ parseDsll dsllFile dsllIn =
               print prog
               divLine
               let env = check prog
+              divLine
+              putStrLn "DSLL Env: \n"
+              print env
               return env
 
 -- --------------------------------------- Test Driver -------------------------------------
--- a/src/Env.hs
+++ b/src/Env.hs
@@ -135,7 +135,7 @@ tParser, tTypeCtorAppParser, typeVarParser' :: Parser T
 tParser = try tTypeCtorAppParser <|> typeVarParser'
 tTypeCtorAppParser = do
     i         <- identifier
-    arguments <- option [] $ parens (argParser `sepBy1` comma) 
+    arguments <- option [] $ parens (argParser `sepBy1` comma)
     --try (parens (argParser `sepBy1` comma)) <|> emptyArgList --option [] $ parens (argParser `sepBy1` comma) --try (parens (argParser `sepBy1` comma)) <|> emptyArgList
     pos <- getPosition
     return (TConstr (TypeCtorApp { nameCons = i, argCons = arguments, constructorInvokerPos = pos }))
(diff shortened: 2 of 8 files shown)


hypotext:
Thanks, I will review it soon!

@dormaayan Could you standardize the error-handling in Substance and DSLL? In particular:
* removing all uses of the runtime `error` function, returning a string for the error instead
* returning a list of error strings instead of one big string

This is useful for the Style compiler, as I don't want the Substance typechecker to be throwing runtime errors while compiling.

dormaayan:
@hypotext , There is already an issue for that: (#85). If it's urgent I can do that soon

hypotext:
Yes, I'd say removing runtime errors is urgent (and it's an easy fix) because it allows me to write test cases easily. You don't have to do the line/col number fixes now.

src/Dsll.hs
```diff
@@ -58,6 +66,16 @@ instance Show Vd where
               cString = show typesVd
               dString = show toVd
 
+-- | Subtype declarations
+data Std = Std {subType :: T,
```
hypotext: `Std` is a confusing name. I would rename it `SubtypeDecl`
(Same for `Od` -> `PredDecl`, etc.)

hypotext: minor nit: standardize record whitespace (spaces around brackets, lining up type decls)

src/Dsll.hs
```diff
@@ -104,42 +122,24 @@ instance Show Pd2 where
               bString = show propsPd2
               cString = show toPd2
 
-
--- | DSLL program composed out of type constructors followed by ar declarations, operations and predicates
-data DSLLProg = DSLLProg { cd :: [Cd],
```
hypotext: I would still store DSLL programs in this data structure (easier to analyze). I'd just loosen up the parsing order.

src/Dsll.hs
```diff
@@ -148,13 +148,21 @@ cd1 = do
...
-    return Cd { nameCd = name, inputCd = [], outputCd = t' }
+    return (CdStmt Cd { nameCd = name, inputCd = [], outputCd = t' })
+
+-- | sub type declarations parser
+stdParser :: Parser DsllStmt
+stdParser = do
+  subtype <- tParser
+  rword "<:"
+  supertype <- tParser
+  return (StdStmt (Std { subType = subtype, superType = supertype}))
```
hypotext: you can also write `return $ StdStmt $ Std { subType = subtype, superType = supertype}`. often application is easier to read and write than a lot of parens

src/Dsll.hs
dormaayan: Where do we need to analyse it (further than the typechceker which does not use it at all)?

src/Dsll.hs
```diff
@@ -203,90 +211,111 @@ pd1 = do
...
-                                     then ef { predicates = M.insert (namePd1 v) pd1 $ predicates ef }
-                                     else error ("Error!")  -- Does not suppose to reach here
-
-checkPredicates e (Pd2Const v) = let pd = Pred2 $ Prd2 { namepred2 = namePd2 v, plspred2 = seconds (propsPd2 v),
-                                                               ppred2 = toPd2 v }
-                                     ef = addName (namePd2 v) e
-                                 in ef { predicates = M.insert (namePd2 v) pd $ predicates ef }
+checkDsllStmt e (StdStmt s) = let env1 = checkDeclaredType e (subType s)
+                                  env2 = checkDeclaredType env1 (superType s)
+                                  env3 = env2{subTypes = (subType s,superType s) : subTypes env2}
```
hypotext: add spacing around records

src/Dsll.hs
```diff
@@ -203,90 +211,111 @@ pd1 = do
...
+checkDsllStmt e (PdStmt (Pd2Const v)) = let pd = Pred2 $ Prd2 { namepred2 = namePd2 v, plspred2 = seconds (propsPd2 v),
+                                                                ppred2 = toPd2 v }
+                                            ef = addName (namePd2 v) e
+                                         in ef { predicates = M.insert (namePd2 v) pd $ predicates ef }
+
+computeSubTypes :: VarEnv -> VarEnv
+computeSubTypes e = let env1 = e { subTypes = transitiveClosure (subTypes e)}
+                    in if isClosureNotCyclic (subTypes env1) then
+                      env1
+                    else env1 { errors = errors env1 ++ "Cyclic Subtyping Relation! \n"}
```
hypotext: after storing a list of errors, you can remove the newlines from error strs

dormaayan: I keep them since I have one big string of all the errors, seperated by newlines

src/Env.hs
```diff
@@ -170,6 +153,20 @@ emptyArgList = do
 
 ----------------------------------- Typechecker aux functions ------------------------------------------
 
+-- | Compute the transitive closure of list of pairs
```
hypotext: I would have written this using an infinite list created by applying `iterate` function with `closureAccum`, dropping elements a fixpoint is reached. This has the advantage that you don't have to manually write a recursion (which is slower).

src/Env.hs
```diff
@@ -170,6 +153,20 @@ emptyArgList = do
...
+--   Useful for subtyping and equality aubtyping checkings
+transitiveClosure :: Eq a => [(a, a)] -> [(a, a)]
+transitiveClosure closure
+  | closure == closureAccum = closure
+  | otherwise                  = transitiveClosure closureAccum
+  where closureAccum =
+          nub $ closure ++ [(a, c) | (a, b) <- closure, (b', c) <- closure, b == b']
+
+-- | Return whether a closure is cyclic (a,b) and (b,a) appears in the closure
```
hypotext: If a closure is cyclic, then don't you just have to check for the presence of `(a, a)` in the closure?

src/Env.hs
hypotext: Also, is there any condition under which this function doesn't terminate? If so, would be good to document it (or test to avoid it)

src/Env.hs
```diff
@@ -302,14 +305,21 @@ checkTypeCtorApp e const = let name = nameCons const
...
-                                                                          " != " ++ (show kinds2) ++ "\n") }
+                                                      then env1 { errors = errors env1
+                                                                          ++ ("Args do not match: " ++ show kinds1 ++
+                                                                          " != " ++ show kinds2 ++ "\n") }
                                                       else env1
-                                      Left err -> env1 { errors = (errors env1) ++ err }
+                                      Left err -> env1 { errors = errors env1 ++ err }
+
+checkDeclaredType :: VarEnv -> T -> VarEnv
+checkDeclaredType e (TConstr t) = if nameCons t `elem` typeCtorNames e
```
hypotext: fix whitespace in this case (line up if statements, etc)

src/Substance.hs
```diff
@@ -137,26 +148,30 @@ predicateParser = do
   return Predicate { predicateName = n, predicateArgs = args, predicatePos = pos }
 
 subStmt, decl, bind, applyP, labelDecl, autoLabel, noLabel :: Parser SubStmt
-subStmt = labelDecl <|> autoLabel <|> noLabel <|> try bind <|> try decl <|> try applyP
-decl = do
-  t' <- tParser
-  v' <- varParser
-  return (Decl t' v')
+subStmt = labelDecl <|> autoLabel <|> noLabel <|> try equalE <|> try equalQ <|> try bind <|> try decl <|> try applyP
```
hypotext: for readability, i'd put each case on a separate line

src/Dsll.hs
hypotext: I'd prefer if the `VarEnv` stored a list of strings. That would be cleaner since you can choose later how you want to display the errors

src/Substance.hs
```diff
@@ -304,11 +328,9 @@ checkVarE varEnv v = case M.lookup v (varMap varEnv) of
 checkFunc :: VarEnv -> Func -> (String, Maybe T)
 checkFunc varEnv (Func f args) = let vcEnv = M.lookup f (valConstructors varEnv)
                                      fEnv  = M.lookup f (operators varEnv)
-                                 in if (isNothing(vcEnv) && isNothing(fEnv))
-                                    then ("Function or Val Constructor " ++ (show f) ++ " not in environment\n", Nothing)
-                                    else if (isJust(vcEnv))
-                                         then checkVarConsInEnv varEnv (Func f args) (fromJust vcEnv)
-                                    else checkFuncInEnv varEnv (Func f args) (fromJust fEnv)
+                                 in if isNothing vcEnv && isNothing fEnv
```
hypotext: I would clarify the logic here as follows:
```
case (vcEnv, fEnv) of
(Nothing, Nothing) -> -- error here
(Just vcEnv', Nothing) -> -- do the value ctor check
(Nothing, Just fEnv') -> -- do the function check
(Just _, Just _) -> -- error: should not be both value ctor and function
```

src/Substance.hs
```diff
@@ -378,34 +400,28 @@ applySubstHelper sigma (AT t) = AT (applySubst sigma t)
 subst :: VarEnv -> M.Map Y Arg -> [K] -> [K] -> M.Map Y Arg
 subst varEnv sigma argTypes formalTypes = let types = zip argTypes formalTypes
                                               sigma2 = foldl (substHelper varEnv) sigma types
-                                          in if ((length argTypes) /= (length formalTypes))
-                                            then error ("Arguments list lengths are not equal")
+                                          in if length argTypes /= length formalTypes
+                                            then error "Arguments list lengths are not equal"
```
hypotext: Give more descriptive error message: "There were $NUM1 arguments to the constructor $CTOR, but it expects $NUM2 arguments"

src/Substance.hs
```diff
@@ -190,7 +205,16 @@ checkSubStmt varEnv (Bind v e) = let (vstr, vt) = checkVarE varEnv v
                                                    ++ show et
                                                    ++ " assigned to variable of type " ++ show vt ++ "\n"}
                                      else varEnv { errors = errors varEnv ++ vstr ++ estr }
-
+checkSubStmt varEnv (EqualE expr1 expr2) = let (estr1, et1) = checkExpression varEnv expr1
+                                               (estr2, et2) = checkExpression varEnv expr2
+                                            in if isJust et1 && isJust et2 && et1 /= et2
```
hypotext: standardize indentation here with let, if, then, else

src/Substance.hs
```diff
@@ -441,6 +457,50 @@ substInsert sigma y arg = case M.lookup y sigma of
...
+----------------------------------------- Binding & Equality Environment ------------------
+-- | Definition of the suBSTANCE environemt + helper functions.
+--   Contains binding information and equality of expressions and predicates
+--   In order to calculate all the equalities, we cmpute the closure of the
+--   equlities in Substance + symetry
+
+data SubEnv = SubEnv { exprEqualities :: [(Expr,Expr)],
+                       predEqualities :: [(Predicate,Predicate)],
+                       bindings       :: M.Map Var Expr,
+                       subPreds :: [Predicate]
```
hypotext: standardize record spacing

src/Substance.hs
```diff
@@ -441,6 +457,50 @@ substInsert sigma y arg = case M.lookup y sigma of
...
+-- | The top levek function for computing the Substance environement
+--   Important: this function assuemes it runs after the typechecker and that
+--              the program is well-formed (as well as the DSLL)
+loadSubEnv :: SubProg -> SubEnv
+loadSubEnv p = let subEnv1 = foldl loadStatements initE p
+                   subEnv2 = computeEqualityClosure subEnv1
+               in subEnv2
+              where initE = SubEnv {exprEqualities = [], predEqualities = [], bindings = M.empty, subPreds = []}
+
+loadStatements :: SubEnv -> SubStmt -> SubEnv
```
hypotext: this is better named as `loadStatement`

hypotext: you might want to comment that the order in all the lists is reversed (in case anyone is relying on order of Substance statements... hopefully not)

src/Substance.hs
```diff
@@ -540,17 +600,24 @@ subSeparate = foldr separate ([], [])
...
 parseSubstance subFile subIn varEnv =
                case runParser substanceParser subFile subIn of
                Left err -> error (parseErrorPretty err)
                Right xs -> do
+                   divLine
                    putStrLn "Substance AST: \n"
                    mapM_ print xs
+                   let subTypeEnv = check xs varEnv
+                       c          = loadObjects xs subTypeEnv
+                       subSubEnv   = loadSubEnv xs
```
hypotext: maybe rename `subSubEnv` to `subDynEnv`? it behaves a little more like a dynamic/evaluation environment (even though it's not exactly that). `subEqEnv` (for equality) could also work

src/Utils.hs
```diff
@@ -134,17 +134,22 @@ attribs = ["shape", "color", "label", "scale", "position"]
 attribVs = shapes
 shapes =  ["Auto", "None", "Circle", "Box", "SolidArrow", "SolidDot", "HollowDot", "Cross"]
 labelrws = ["Label", "AutoLabel", "NoLabel"]
-dsll = ["tconstructor","vconstructor","operator","forvars","fortypes","predicate", "Prop", "type"]
+dsll = ["tconstructor","vconstructor","operator","forvars","fortypes","predicate", "Prop", "type", "<:", ":=", "<->"]
```
hypotext: we don't have `:=` in the DSLL; it got replaced by "->".
e.g. `notation "f.domain" -> "Dom(f)"`

src/dsll/algebra.dsl
```diff
@@ -1,16 +1,18 @@
 tconstructor VectorSpace: type
 tconstructor Vector(v : VectorSpace) : type
 tconstructor Scalar : type
+tconstructor ZeroScalar : type
```
hypotext: why are we distinguishing ZeroScalars as a type?
I would add a predicate `IsZero(c : Scalar) : Prop` if you need that

src/paper-eval/linear-algebra-maps.sub
```diff
@@ -0,0 +1,92 @@
...
+LinearMap g : V → W
+LinearMap h : W → X
+Vector u1, u2, u3, u4, u5 ∈ U
+Vector v1, v2, v3 ∈ V
+u3 := u1 + u2
+v1 := f(u1)
+v2 := f(u2)
+v3 := f(u3) = v1 + v2
+-- the "equivalence" equal sign is tentative
+Scalar a := det(u1,u3) u4 := a * u2
```
hypotext: watch out for newlines here and below, e.g. before `u4`

src/paper-eval/linear-algebra-maps.sub
```diff
@@ -0,0 +1,92 @@
...
+In(v1,V)
+In(v2,V)
+In(v3,V)
+
+u3 := Add(u1,u2)
+v1 := Apply(f,u1)
+v2 := Apply(f,u2)
+v3 := Apply(f,u3)
+
+v3 = Add(v1,v2)
```
hypotext: you could also write
`Add(Apply(f(u1)), Apply(f(u2))) = Apply(f, Add(u1, u2))`
(i guess it's my responsibility to make sure that all versions work in Style)

src/paper-eval/linear-algebra-maps.sub
```diff
@@ -0,0 +1,92 @@
...
+Vector w1 ∈ W := g(v1) Vector x1 ∈ X := h(w1)
+
+Autolabel All
+Label u4 "$u_4 = au_2/2$
+Label u5 "$u_5 = -u_4$"
+Label c "$|v_1| \langle v_1, v_3 \rangle$" Label w1 "$g(v_1)$"
+Label x1 "$(h \circ g)(v_1)$"
+
+*/
+VectorSpace U
```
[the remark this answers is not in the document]
hypotext: can you put the desugared and sugared version in separate files?

src/paper-eval/linear-algebra.dsl
```diff
@@ -0,0 +1,16 @@
+tconstructor Scalar : type
+tconstructor VectorSpace : type
+tconstructor Vector : type
+tconstructor LinearMap : type
+
+operator Add (v1 : Vector, v2 : Vector) : Vector
```
hypotext: now we want functions to have lowercase names (it's OK for predicates to have uppercase names, i think)

src/paper-eval/real-analysis.dsl
```diff
@@ -0,0 +1,73 @@
...
+predicate OpenEnd (p : Real, I : Interval) : Prop
+
+-- Prelude exports
+--value RR : Reals
+
+-- notation is ASCII only
+-- kw NEW := OLD
+--notation "f.domain" -> "Dom(f)"
+-- TODO: flip rest
+--notation "Cod(f)" := "f.codomain"
```
hypotext: I forgot to update these to use `->`

src/sub/vectors.sub
```diff
@@ -1,5 +1,22 @@
 VectorSpace U
 Vector(U) A
 Vector(U) B
+Vector(U) C
+
+Scalar C1
+ZeroScalar C2
```
hypotext: why `ZeroScalar`?

src/Dsll.hs
hypotext: ok, I admit we aren't analyzing it anywhere, but I think it's neater to have all decls of the same type grouped together in the datatype :-)

hypotext:
Looks good overall! Also thanks for using `hlint`. The code is much more readable.

Can you add the names of the test files you're using, and/or steps for reproducing the tests, to the PR description? e.g. I would write, in my PR, `runpenrose sub/linearMap.sub sty/vectors-new.sty dsll/algebra.dsl`

src/sub/vectors.sub
dormaayan: Oh, that is just for my checks of subtyping, we do not need that in the real book....

Issue #97, opened 2018-06-26
Implementation of the new Style language


wodeni:
Implement a new design of the Style language, which has language features such as a new selector grammar and redesigned Style semantics.

## High-level Steps

- "Service" language grammar
- Full Style parser
- Selector statics 
- Matching dynamics
- Translation dynamics
- Block statics
- Selector specificity
- Computational graph building, checking, and inference
- Tests for the new features
- Better error messages similar to that of the DSLL
- Substance - Style interface refactoring

## Canonical names for commonly used variables
- <img width =20 src="https://user-images.githubusercontent.com/11740102/41912416-b1b2b240-791d-11e8-85fa-6a7f7f5d0d27.png"/>: `Env`, typing context
-  <img width =20 src="https://user-images.githubusercontent.com/11740102/41912465-e03e773e-791d-11e8-9c77-14cf904c7f9c.png"/>: local/selector environment
- <img width =20 src="https://user-images.githubusercontent.com/11740102/41912645-01f9e25a-791e-11e8-932e-4cd2a7d08edb.png"/>: Substance bindings
- <img width =20 src="https://user-images.githubusercontent.com/11740102/41912782-14b0e1d2-791e-11e8-930c-a56f75d25693.png"/>: Subtitution
- <img width=20 src="https://user-images.githubusercontent.com/11740102/41913652-e6dab782-791e-11e8-9630-018908471ffd.png"/>: Translation

hypotext:
First draft of new Style compiler was finished in August 2018. Will open new issues for new language features if needed.

PR #99, opened 2018-06-28
New Style Language

wodeni:
# Description

Related issues: #97 

Implement a new design of the Style language, which has language features such as a new selector grammar and redesigned Style semantics.

# Current TODOs (7/17)

**bolded** = highest priority

**Port**
- [x] **Shape definitions**
  - [x] Circle
  - [x] Arrow
  - [x] Curve (CubicBezier?)
  - [x] Line
  - [x] Square
  - [x] Arc
  - [x] Rectangle
  - [x] Parallelogram
  - [x] Text
  - [x] Image
  - [x] Ellipse
  - [x] Point (AnchorPoint?)
  - [x] Their samplers
- [ ] Shapes' derived properties
- [ ] **Shape rendering code**
  - [x] Circle
  - [x] Arrow
  - [x] Curve (CubicBezier?)
  - [x] Line
  - [x] Square
  - [x] Rectangle
  - [ ] Parallelogram (needs @dormaayan to clean up code first)
  - [x] Text
  - [x] Image
  - [x] Ellipse
  - [ ] AngleMark  (needs @dormaayan to clean up code first)
  - [x] Point (AnchorPoint?)
- [x] **Objective functions** 
- [x] **Constraints**
- [x] **Computations**
- [x] **Test**
  - [ ] Shapes
  - [ ] Objectives
  - [ ] Constraints
  - [ ] Computations

Compiler + front-end
- [x] **Labels and text**
- [ ] Layering
- [x] Pass frontend info back to optimizer

Style compiler
- [ ] **Implement lists and accesses, as well as tuples (Pt2)**
- [ ] **Raytracing semantics**
- [ ] Test matching more thoroughly
- [ ] Generate names for anonymous fields (don't belong to a specific Sub object). Does the parser need to change?
  - [ ] Fix programs that use cartesian selectors (e.g. venn.sty, `Set x; Set y`) so they are less brittle WRT the arbitrarily-chosen name
- [ ] Namespaces
- [ ] Add/use beta environment
- [ ] Matching WRT equality and implication
- [ ] Matching anon exprs (make anonymous variables for unification for Substance program)
- [ ] Figure out if we need to check substitution validity
- [ ] Figure out what `T x; T y` should match (all 4 combinations or not?)
- [ ] (*ShapeDef* related) use `invoke*` functions defined in `ShapeDef` to type check the objectives/constraints/comps. 

Optimization
- [x] **Get "contains" working**
- [x] Sizefuncs on everything, or canvas bbox contains everything

Shapes
- [ ] Derived properties
- [ ] Parse `.shapes` file
- [ ] Splice in rendering code
- [ ] Composite shapes

End-to-end
- [x] **Repro wiki examples**
- [ ] **Get LA example working**
- [ ] **Get RA example working**

Documentation/cleanup
- [ ] **Document the three representations (translation, varying state, shapes) and information flow between them**
- [ ] Document sampling/setting logic
- [ ] Clean up/standardize error messages and warnings
- [ ] Clean up / split up NewStyle
- [ ] Document NewStyle, remove old code and TODOs, and add type signatures
- [ ] Remove old Runtime, Style, Shapes, Functions, Computation files and their tests
- [ ] Write actual tests for new Style, Optimizer, Shapes
- [ ] Document ordering
- [ ] Write up how the compiler works in the design paper

# High-level steps
- [x] "Service" language grammar
- [x] Full Style parser
- [x] Selector statics 
- [x] Matching dynamics
- [x] Translation dynamics
- [ ] Block statics
- [ ] Selector specificity
- [ ] Computational graph building, checking, and inference
- [ ] Tests for the new features
- [ ] Better error messages similar to that of the DSLL
- [x] Substance - Style interface refactoring

# Implementation notes and TODOs
## Style parser
- TODOs
  - [x] Parsing selectors
  - [x] Parsing expressions
  - [x] Parsing blocks
  - [x] Parsing bindings
  - [x] Parsing lists
  - [x] Parsing Layer stmts
- Error handling can be done with `SourcePos` type returned by `getPosition` from Megaparsec. The question is how to do it. Do we want to change the AST types to records instead of sum types and add the positions as fields? Might be messy to write
- Permutation parsers are magical!

# Checklist

# Open questions


wodeni:
Seems like Travis is failing due to a bug in `happy`, which `pretty-show` depends on. Will look into this later

hypotext:
Basic translation works now. Still need to link to Runtime and fix rough edges around namespaces and equality.

Repro: `runpenrose paper-eval/linear-algebra-maps.sub paper-eval/linear-algebra.sty paper-eval/linear-algebra.dsl`

Result: [gist](https://gist.github.com/hypotext/ffa9bf53d3a1c0850eeea82d6e7eceb2)

wodeni:
I added a mechanism for sampling properties of shapes more systematically by modifying the type of `PropertiesDef a` to contain a sampler function, which can be invoked with a random generator to generate a sampled value for a property.

An example sampler function is the following:
```Haskell
sampleFloatIn :: (Autofloat a) => FloatInterval -> SampledValue a
sampleFloatIn interval g =
    let (n, g') = randomR interval g in (FloatV $ r2f n, g)
```

Inside of our shapedef, it will be used as:

```Haskell
circType = ("Circle", M.fromList
    [
        ("x", (FloatT, sampleFloatIn (0.0, canvasWidth))),
        ("y", (FloatT, sampleFloatIn (0.0, canvasHeight))),
        ("r", (FloatT, sampleFloatIn (0.0, canvasWidth / 6))),
        ("stroke-width", (IntT, constValue $ IntV 0)),
        ("name", (StrT, constValue $ StrV "defaultCircle")),
        ("style", (StrT, constValue $ StrV "filled")),
        ("color", (ColorT, constValue $ ColorV black))
    ])
```

My rationale for the design is the following (copying from the commit comments):
- In general, I don't think there is a need for default values (vs.
sampled values). A constant value is just a sampled value drawn from a
sample space with only one element.
- Therefore, I think shape def should include samplers in place of their
default values and we can sample them in a principled manner.
- The question now is how to not sasmple anything manually in Translation. If translation needs to have a sampled property value for one of the varying params, it should look for a readily genereated value from an instance of the corresponding shapedef, NOT doing the sampling by itself.
- In a long run, we would expose some syntax for the shapedef writers to
specify how to sample their values, which is not limited to only
numbers but all the available types in Style.

hypotext:
This mechanism looks good to me in general. Three minor comments:

- does sampleFloatIn take a “g”? That’s not in its type.
- is circType storing all the resulting “g”s from sampling? That seems
unnecessary to me.
- we can revert strokeWidth to being a float sampled from, say, 0-3. The
bounds depend on the radius but we can ignore that for now

wodeni:
This typedef should explain it a bit better:
```
type SampledValue a = StdGen -> (Value a, StdGen)
```

src/Server.hs
```diff
@@ -173,11 +173,12 @@ updateShapes newShapes conn s =
     let polyShapes = toPolymorphics newShapes
         uninitVals = map NS.toTagExpr $ NS.shapes2vals polyShapes $ NS.uninitializedPaths s
         trans' = NS.insertPaths (NS.uninitializedPaths s) uninitVals (NS.transr s)
```
hypotext: At this point, are the only uninitialized paths in the translation the label dimensions? (Or anything else that has to be figured out renderer-side)

hypotext:
Moved to #121 on the `new-style-fast` branch.

PR #133, opened 2018-09-04
Merge Syntactic Sugar to master

dormaayan:
# Description

Related issues: #112 

This PR adds syntactic sugar support for Penrose.

1. In order to define new syntax extension, you need to add a `StmtNotation` in the dsll file. For example, a syntax extension for vectors additions might be:
`StmtNotation "v1 + v2" -> "AddV(v1,v2)"`
Note that `v1` and `v2` are automatically recognized as patterns since they are not a special sign, they are not part of any DSLL definition, and they appear in both sides of the `StmtNotation`
In order to write a multi-line desugaring, you should use `;` for example:
`StmtNotation "LinearMap f : U → V" -> "LinearMap f;From(f,U,V)"`

2. I created a Lexer using Alex, which gets as an input a sugared Substance program and creates a list of tokens of the Substance program.

3. I use the lexer in order to tokenize the sugared Substance program, the lexer, and the 

All the LA examples in the branch are sugared and it seems to work fine.  You may also find a syntactic sugar definitions for the RA and set theory domains which work fine at the language level.

# Checklist

# Open questions for future work

- We do not support nested expressions, for example, we are not able to write the following:
 `A := B + C + D` and get a desugared version of it automatically. However, we can define at the DSLL an explicit statement notation for it.

- We do not support recursive structures other than the one described above. Syntactic sugars for lists, tuples, etc. should be defined at `Substance.hs`

- There are other relevant works on syntactic sugars and syntax extensions. The most relevant are [coq](https://coq.inria.fr/refman/user-extensions/syntax-extensions.html) and the book The language of Mathematics


hypotext:
Sounds fine to me. I won't be pushing much to that branch over the weekend.

dormaayan:
@hypotext , you can review and merge. Let me know if you face any special problems with using it for the RA domain but it seems fine to me so far

src/Dsll.hs
```diff
@@ -118,25 +125,51 @@ data Pd2 = Pd2 { namePd2 :: String,
...
+    show (En fromEn toEn associativityEn precedenceEn) =
+       "(ExprNotation: from: " ++ a ++ " to: " ++ b ++ " associativity: "
+        ++ c ++ " precedence: " ++ d
+         where a = show fromEn
+               b = show toEn
+               c = show associativityEn
+               d = show precedenceEn
+----------------------------------------- DSLL Parser --------------------------
+
+-- | 'DSLLParser' is the top-level parser function. The parser contains a list
```
wodeni: Some of the comments seem to be a little out of date here.

src/Dsll.hs
```diff
@@ -118,25 +125,51 @@ data Pd2 = Pd2 { namePd2 :: String,
...
---   operations followed by a collection of predicates
+-- |'dsllProg' parses the entire actual DSLL program which is a collection of
+-- constructors followed by a collection of operations followed by a collection
+-- of predicates
 dsllProgParser :: Parser [DsllStmt]
 dsllProgParser = dsllStmt `sepEndBy` newline'
 
 dsllStmt :: Parser DsllStmt
-dsllStmt = try cdParser <|> try vdParser <|> try odParser <|> try subtypeDeclParser <|> try pdParser
+dsllStmt = try snParser <|> try enParser <|> try cdParser <|> try vdParser
```
wodeni: We have `tryChoice` in Utils.hs now, which is a neater way to write this

hypotext:
@dormaayan In the README, what do you mean by this sentence? 

> I use the lexer in order to tokenize the sugared Substance program, the lexer, and the

hypotext:
I'd like to build this branch so I can repro it, but it looks like it's going to unregister and reinstall all my dependencies. Here's the output from two `stack` commands: https://gist.github.com/hypotext/791a43bd29df002c326c8dbf1eb44c19

Is there some way to prevent this rebuilding? Or is it inherent due to some difference in the dependencies between my currently-built `real-analysis` branch and this branch? Is this going to cause problems when merged to master?

dormaayan:
@hypotext to the best of my knowledge this is because I use Alex and added things we didn't use before. That is of course not the case for the other branch which contains much more trivial changes.

src/SubstanceTokenizer.x
```diff
@@ -0,0 +1,87 @@
+{
+module SubstanceTokenizer where
+}
+%wrapper "basic"
+$digit = 0-9
+$alpha = [a-zA-Z]
+$symbol = [\=\+\-\*\/\:\∈\←\→\↑\↓\↔\↕\↖\↗\↘\↙\↚\↛\↮\⟵\⟶\⟷\<\>\|\;\~\`\!\#\%\&\*\±\§\?\$\<\>\⊆\∪\∩\`\∫\[\]\^]
```
wodeni: Seems like this is an arbitrary list of symbols. Anyway to include all Unicode chars?

hypotext:
@dormaayan Could you add a section documenting the design, implementation, and use of the syntactic sugar in the design paper? (With any grammars/semantics necessary.) 

Doesn't have to be super long/detailed, and you should feel free to take material from the existing Google docs. The idea is just to make sure that we have a complete spec of the language in one place. Thanks!

hypotext:
@dormaayan Quick reminder about the documentation—it's very important to have a complete language spec written somewhere outside of the implementation. So it would be great if you could document the syntactic sugar, value deconstructors, and Prelude in the design paper. See also my comment [here](https://github.com/penrose/penrose/pull/137#issuecomment-439182722) for more details. Thanks!

dormaayan:
@hypotext, Merged this PR to master and checked the integrations with the features from #137 
Don't worry, I'll add full documentation to the design paper in the next few days as I told you already 😀

hypotext:
Awesome, thanks! 🙂

PR #152, opened 2018-11-27
Addition of Docker container

vsoch:
# Description

This is an early pull request to do the following:

 - Add a set of Dockerfiles in the docker folder, with instructions to the developer [here](https://penrose.github.io/penrose.github.io/penrose-development/docker-local/) and the user [here](https://penrose.github.io/user-docs/hello-world/) for getting started with it.
 - add Continuous Integration via CircleCI to build and deploy a Docker container. This will need to be hooked up to build properly, and importantly, we should have a "penrose" organization. I can make this if others want to share their Docker Hub usernames to be given access. 
 - the domains are re-organized under a "domains" folder. Having them under a common root with the top level as the URI seems to be a good strategy for revealing them to the user (and in the future) installing additional ones.

A few things to talk about / still need work:

**A Better Base Docker Image**

(priority lower)

I opened an [issue here](https://github.com/commercialhaskell/stack/issues/4419#issuecomment-442019053) and aside from caching on CircleCI, I don't see good ways to make the build any faster. This means slower development, and we definitely don't want the user / developer to need to do this. This was the rationale that I made the second Dockerfile.dev - it uses the first as a base image and only re-adds the static content (happens very quickly). If, however, in developing a new domain the entire compile is needed again, this is going to hugely de-incentivize working on it locally, but only given that haskell re-compiling is needed (so maybe not so bad)

**Trio of Style, Sub, Dsl Files**

(priority higher)

I'd like to talk about how the style, *.sub, and DSL files are linked together. For example, how would a user know which triads work together?  I did a lot of trial and error to find a trio, and I think we need to have these trios defined something. The somewhere I think should be an API that is an endpoint to serve json that the client can parse, each domain entry pointing to a github repository for the domain (e.g., penrose/domain-linear-algebra) serves it's own metadata, which is validated by (`penrose/domains`) and then fed into the program from `https://penrose.github.io/domains/library.json` (or similar). How well does Haskell parse json? For example, I noticed that the filename is read as a string here: https://github.com/penrose/penrose/blob/master/src/ShadowMain.hs#L43 and is there any reason we can't provide the web application the URL to that same file? Is there any reason we can't have an opening portal that parses the API and then let's the user interactively select the trio?

Given we have this modular approach, we wouldn't need to store the domains with penrose (e.g., penrose/domains can be removed here). We would, however, want the plugins to be versioned so they know what version of penrose they can be matched with.

**GHC Compiler Version**

Is there a reason for the GHC 8.2.1 compiler? It's one of the few that *isn't* provided as a base container [here](https://hub.docker.com/r/fpco/stack-build/tags/). I got around this by adding a flag `--system-ghc` but this would not be a good idea if you have a strong reason for a specific version. Can we not use one that has long term support? (lts) 

 - domains should be under a common root, "domains" for easy install and listing
 - a main Dockerfile should do most of compiling, and a development (Dockerfile.dev) will allow for quick changes to static files.

# Checklist

Just wondering - where is this site hosted? How do I run this for a modified ghc?

I'm not privy to old ones, or what a warning looks like, so I'll need help here!

A comment here, the tests pass with an amendment:

```bash
 stack test --system-ghc
...
  Style tests
    Unit tests
      Parse cart-test-pt.sty:               OK
      Parse continuousmap.sty:              OK
      Parse surjection_computed.sty:        OK
      Parse tree.sty:                       OK
      Parse twopoints.sty:                  OK
      Parse venn.sty:                       OK
      Parse venn_comp.sty:                  OK
      Parse venn_comp_simple.sty:           OK

All 29 tests passed (0.02s)
               
Completed 15 action(s).
```

I don't think this is applicable for me, I didn't edit Haskell? (other than one line)

# Open questions

Oups, see my section at the top! In summary we should talk about:
 
 - ghc version
 - domains 
 - setting up CircleCI / DockerHub for penrose organization. If I have admin on penrose I can set up Circle here, or point someone to docs for how to do it. It would set up the equivalent pipeline we have [here](https://circleci.com/workflow-run/706393e2-0556-4aac-a5cf-a18b8f6ec81d).

And I want to say thank you to @dandavison [here](https://github.com/dandavison/penrose/pull/1) because I was scratching my head over exposing websockets in the Docker container and his help saved the day! :D) 

I have a few meetings this afternoon and then need to switch my brain into GoLang mode, so please take your time to think and review!


vsoch:
Interesting! It looks like Github isn't recognizing the move in folder (and rename) of the domains to here https://github.com/vsoch/penrose/tree/master/src/domains as a conflict. See if you can resolve in the interface or if we need to try something else.

docker/entrypoint.sh
```diff
@@ -0,0 +1,103 @@
...
+                           as input to the penrose executable
+
+         Commands:
+
+                help: show help and exit
+                web: start penrose frontend (requires mapping of ports)         
+
+         Examples:
+             docker run -p 8000:8000 -p 9060:9060 <container> web --custom math.sub math.sty math.det
+             docker run -p 8000:8000 -p 9060:9060 <container> web linear-algebra
```
dandavison: Looks like the websockets port numbers should be 9160 in this help text.

PR #159, opened 2018-11-29
Adding Docker Container (to master branch without conflict folders)

vsoch:
This is a migration from #152 , and the main difference is that it's from a master where the files have been removed (not sure how to resolve the other merge conflict). I had intended to set up circle ci to push to penroseorg/penrose, but I don't have admin to set up circleci for that, so for now this won't work. I can update the PR when the circleci setup is done, either by me if given some admin, or someone else I am glad to help! Here is the content of the other PR, which I will close in favor of this one.

This is a pull request to do the following:

 - Add a set of Dockerfiles in the [docker folder](https://github.com/vsoch/penrose/blob/master/docker), with instructions to the developer [here](https://penrose.github.io/penrose.github.io/penrose-development/docker-local/) and the user [here](https://penrose.github.io/user-docs/hello-world/) for getting started with it.
 - add [Continuous Integration via CircleCI](https://github.com/vsoch/penrose/blob/master/.circleci/config.yml) to build and deploy a Docker container. This will need to be hooked up to build properly, and importantly, we should have a "penrose" organization. I can make this if others want to share their Docker Hub usernames to be given access. 
 - the domains are re-organized under a [domains folder](https://github.com/vsoch/penrose/tree/master/src/domains). Having them under a common root with the top level as the URI seems to be a good strategy for revealing them to the user (and in the future) installing additional ones.

A few things to talk about / still need work:

**A Better Base Docker Image**

(priority lower)

I opened an [issue here](https://github.com/commercialhaskell/stack/issues/4419#issuecomment-442019053) and aside from caching on CircleCI, I don't see good ways to make the build any faster. This means slower development, and we definitely don't want the user / developer to need to do this. This was the rationale that I made the second Dockerfile.dev - it uses the first as a base image and only re-adds the static content (happens very quickly). If, however, in developing a new domain the entire compile is needed again, this is going to hugely de-incentivize working on it locally, but only given that haskell re-compiling is needed (so maybe not so bad)

**Trio of Style, Sub, Dsl Files**

(priority higher)

I'd like to talk about how the style, *.sub, and DSL files are linked together. For example, how would a user know which triads work together?  I did a lot of trial and error to find a trio, and I think we need to have these trios defined something. The somewhere I think should be an API that is an endpoint to serve json that the client can parse, each domain entry pointing to a github repository for the domain (e.g., penrose/domain-linear-algebra) serves it's own metadata, which is validated by (`penrose/domains`) and then fed into the program from `https://penrose.github.io/domains/library.json` (or similar). How well does Haskell parse json? For example, I noticed that the filename is read as a string here: https://github.com/penrose/penrose/blob/master/src/ShadowMain.hs#L43 and is there any reason we can't provide the web application the URL to that same file? Is there any reason we can't have an opening portal that parses the API and then let's the user interactively select the trio?

Given we have this modular approach, we wouldn't need to store the domains with penrose (e.g., penrose/domains can be removed here). We would, however, want the plugins to be versioned so they know what version of penrose they can be matched with.

**GHC Compiler Version**

Is there a reason for the GHC 8.2.1 compiler? It's one of the few that *isn't* provided as a base container [here](https://hub.docker.com/r/fpco/stack-build/tags/). I got around this by adding a flag `--system-ghc` but this would not be a good idea if you have a strong reason for a specific version. Can we not use one that has long term support? (lts) 

 - domains should be under a common root, "domains" for easy install and listing
 - a main Dockerfile should do most of compiling, and a development (Dockerfile.dev) will allow for quick changes to static files.

# Checklist

Just wondering - where is this site hosted? How do I run this for a modified ghc?

I'm not privy to old ones, or what a warning looks like, so I'll need help here!

A comment here, the tests pass with an amendment:

```bash
 stack test --system-ghc
...
  Style tests
    Unit tests
      Parse cart-test-pt.sty:               OK
      Parse continuousmap.sty:              OK
      Parse surjection_computed.sty:        OK
      Parse tree.sty:                       OK
      Parse twopoints.sty:                  OK
      Parse venn.sty:                       OK
      Parse venn_comp.sty:                  OK
      Parse venn_comp_simple.sty:           OK

All 29 tests passed (0.02s)
               
Completed 15 action(s).
```

I don't think this is applicable for me, I didn't edit Haskell? (other than one line)

# Open questions

Oups, see my section at the top! In summary we should talk about:
 
 - ghc version
 - domains 
 - setting up CircleCI / DockerHub for penrose organization. If I have admin on penrose I can set up Circle here, or point someone to docs for how to do it. It would set up the equivalent pipeline we have [here](https://circleci.com/workflow-run/706393e2-0556-4aac-a5cf-a18b8f6ec81d).

And I want to say thank you to @dandavison [here](https://github.com/dandavison/penrose/pull/1) because I was scratching my head over exposing websockets in the Docker container and his help saved the day! :D) 

I have a few meetings this afternoon and then need to switch my brain into GoLang mode, so please take your time to think and review!


docker/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM fpco/stack-build:lts-9.10
```
qrilka: @vsoch this could be easily simplified by switching to newer LTS e.g. https://github.com/qrilka/penrose/commit/2faf1ec4736c5dffcdfcba41191466a12ab2da6a builds fine for me
With that change you could use `fpco/stack-build:lts-11.22` and don't spend time on installing GHC, alex and happy and `--system-ghc` won't be needed (BTW there's `stack setup` to get proper GHC with Stack)
Hope this helps

vsoch: Ah this is so helpful! Let me give this a try locally, and I'll update the PR. I really appreciate your testing this out and advice on this!

vsoch:
:green_apple: :green_book: :green_heart:

vsoch:
oh geez! I totally forgot about finals! I'm sending my best study vibes. If I was there I'd make you guys some good study snacks... I had a lot of fun doing this when I was a TA... TBA for next year :L)  

Let's hold off here for sure, and for now, here is a preview... "Pizza Cookie"

image

and "Candy Tumors" that got baked into cookies, and machine learning learned with them 

image

Good luck! I'll post a note in the slack because I feel really badly for forgetting this.

dormaayan:
@vsoch , we can discussion the syntactic sugar things on our Tuesday meeting.

vsoch:
Would it be okay for me to merge this into `master-without-domains` so I can then figure out how to update to the current master? I'm having trouble trying to do any more work having it so separated from master and I want to get this work integrated with the syntactic sugar additions (and other changes to master). Otherwise I think I'd just abandon the PR because it's too far off from what is currently there, which is unfortunate because I put a lot of time into it.

vsoch:
okay I don't think anyone is working / relying on `master-without-domains` branch, and given that it's the holiday and we had good discussion, I'm going to say it's okay to merge this so I can keep making progress. It shouldn't have any impact on what others are working on, and the next PR I'll do will be some update to the "real" master branch (which I of course would not merge without approval by all).

It looks like there was a `--force` used at some point in the git history which is why I was having trouble with this PR to the original master, I'll see what I can do for the next attempt to resolve this!

hypotext:
You should feel free to merge into a branches that look inactive!

PR #178, opened 2019-02-21
Extend Substance to call external libraries (plugins)

hypotext:
# Description

This PR extends Substance, Style, and ShadowMain to support plugins, which are external programs that can process a Substance program to instantiate it with further Substance objects.

See plugin documentation [here](https://github.com/penrose/penrose/wiki/Adding-and-using-Substance-plugins).

Related issues: #180 

See plugin documentation.

Try any of the three plugins in the `plugins` directory, making sure the Style program imports the plugin, and you should see the generated Substance programs in the compiler output, as well as any corresponding visual representations in the diagram. 

Here's one example command: `runpenrose mesh-domain/Figure1.sub mesh-domain/SimplicialComplex.sty mesh-domain/mesh.dsl`

You can also see surjection/bijection/injection examples that use the Alloy plugin in issue #180.

# Checklist

# Open questions

* Currently the JSON export doesn't handle all of Substance. Should we extend it? For example:
  * What to do about subtyping?
  * How to convert Substance forms that are currently not handles? (e.g. binding a variable, deconstructor, predicate inequality)
  * Converting parametrized types?
* Have the external program return a JSON instead of a Substance program for symmetry?
* Extend Style so it can access values returned by a plugin


hypotext:
@wodeni I'm considering moving the mesh domain to a new branch off of `plugin`, force pushing on `plugin` to remove the mesh commits, then merging `plugin` to master. How does that sound?

wodeni:
> @wodeni I'm considering moving the mesh domain to a new branch off of `plugin`, force pushing on `plugin` to remove the mesh commits, then merging `plugin` to master. How does that sound?

8ecae11 is the last commit from `plugin` on my `raytracing` branch. As long as the history before that is preserved, I guess I'm okay with it

maxkrieger:
@wodeni and I started thinking about how this will work with concurrent users. Reading/writing to one file concurrently obviously doesn't work without unique names. We aren't sure, however, what the alternative is. Is stdin/stdout unique to each spawned process? If not, do we have to resort to IPC?

wodeni:
> @wodeni and I started thinking about how this will work with concurrent users. Reading/writing to one file concurrently obviously doesn't work without unique names. We aren't sure, however, what the alternative is. Is stdin/stdout unique to each spawned process? If not, do we have to resort to IPC?

* https://stackoverflow.com/questions/41438174/how-multiple-process-are-able-to-use-stdin-stdout-simultaneously Turns out stdin and stdout are unique to each process and bidirectional streams can be created using them.
* A quick-and-dirty solution we discussed: just name of tmp files differently using the uuid of each client and store the files in `tmp`
* Other IPC options: https://en.wikipedia.org/wiki/Inter-process_communication

hypotext:
@wodeni I just documented this PR, rolled back to 8ecae11, and merged it to master. If y'all want to extend this to work with concurrent users, feel free.

PR #187, opened 2019-03-05
Add mesh / simplicial complex domain

hypotext:
This PR adds initial support for the mesh domain (in the simplicial complex style) to Penrose, as well as plugin accessors for the plugin and Style, and a heuristic to improve Style matching speed.

To repro, follow the setup directions in `plugins/mesh-plugin`, then run this command:

`runpenrose mesh-domain/Figure1.sub mesh-domain/SimplicialComplex.sty mesh-domain/mesh.dsl`

For the newest examples, see [this wiki page](https://github.com/penrose/penrose/wiki/The-mesh-domain).

For documentation on plugins and plugin accessors, see [the wiki page](https://github.com/penrose/penrose/wiki/Adding-and-using-Substance-plugins).

For documentation on how the plugin actually works, see the top of `plugins/mesh-plugin/mesh-plugin.js`.

# Major future work

* test a more thorough selection of Substance programs to make sure the plugin can handle all of it
* repro diagrams in DDG course notes, e.g. style that shows the action of a function on a different mesh
* improve label placement #196 
* link with Scotty3D and visualize local mesh operations like EdgeFlip, EdgeSplit, etc.

# Minor improvements needed 

* make Style program more generic: all scale factors and offsets should be calculated as a function of canvas size
* make the random meshes look better (sample more uniformly and push vertices away, or do K random flips on an obj file and perform a Tutte embedding)
* improve Style matching speed; currently the Face selector generates hundreds of thousands of matches
* generalize plugin accessors so plugins can pass values that are not just floats

# Old examples

Example of closure of a star at a vertex, and star at a vertex:

image

More recent example using plugin accessors and showing the link at a vertex: (not sure if meshes are laid out correctly)

image

Using a more complicated mesh + plugin accessors + optimization to place vertices in the bbox:

image

With style closer to spec

image

Spec diagram [here](https://github.com/penrose/penrose/blob/plugin/src/mesh-domain/SimplicialComplexStyle.pdf)


hypotext:
@wodeni I'd like to implement the [plugin accessors](https://docs.google.com/document/d/1LCFEYfIkhlZcWKP-T3Fx7MeyBkip6PoKgvXlTjd6biU/edit#) soon. Do you think you could add the form to the Style parser soon? From the doc:

`expr ::= … | L[expr1][expr2]`
where `L` has type “string that is a plugin name," and `expr1`, `expr2` have type string and can be evaluated at Style compile-time. I can deal with the typechecking.

PR #190, opened 2019-03-12
Add Circleci Support

maxkrieger:
This adds circleci support to our repository. There are two features:

* **Automated building and testing**: we can create binaries and test our builds automatically.

* **Continuous tagged deployment**: If you tag your commit (preferably before pushing) following the format `vx.x` where `x` is an integer, Circleci will deploy the tagged commit to our AWS cluster.

### DIFF
--- a/Dockerfile.circleci
+++ b/Dockerfile.circleci
@@ -1,0 +1,12 @@
+FROM ubuntu
+
+RUN mkdir -p /opt/penrose
+COPY . /opt/penrose/
+WORKDIR /opt/penrose
+RUN apt-get update && apt-get install -y \
+  ca-certificates \
+  libgmp-dev
+
+ENTRYPOINT ["./penrose"]
+EXPOSE 9160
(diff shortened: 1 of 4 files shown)


Dockerfile.circleci
```diff
@@ -0,0 +1,12 @@
...
+RUN mkdir -p /opt/penrose
+COPY . /opt/penrose/
+WORKDIR /opt/penrose
+RUN apt-get update && apt-get install -y \
+  ca-certificates \
+  libgmp-dev
+
+ENTRYPOINT ["./penrose"]
+EXPOSE 9160
+CMD ["--domain=0.0.0.0","src/sty/venn.sty","src/set-theory-domain/setTheory.dsl"]
```
wodeni: We should change the CLI to use the remote version of the files instead in the near future

maxkrieger:
Maybe allow omitting the domain files as arguments altogether? We can put
the burden on the frontend to send them over.

Dockerfile.circleci
maxkrieger: Maybe allow omitting the domain files as arguments altogether? We can put the burden on the frontend to send them over.

Issue #207, opened 2019-04-15
All the graphics


hypotext:
Goal: have great-looking reasonably-complex figures that are laid out reasonably fast and automatically. 

Timeline: May 10 (TODO: break this down)

Port
- [x] Circle
- [x] Polygon
- [x] Rectangle
- [x] Square
- [x] Line
- [x] Arrow (deprecated)
- [x] Curve
  - [x] with arrowheads
  - [x] with thickness
- [x] Image
- [x] AnchorPoint (deprecated)
- [x] Parallelogram
- [x] Ellipse
- [ ] Text
  - [x] Bbox
  - [ ] Multiple polygons
- [ ] (Brace)
- [ ] (Arc)
- [ ] All of the above, but accounting for fill, holes, and multiple parts

Katherine and Rain

- [x] debug two-circle containment with second order opt?
- [x] model a richer polygon type (how to represent curves and polylines)
- [x] discuss polygonization of curves, arrows, and text
- [ ] test all the energies 
  - [ ]  representative: boundary intersect, containment, disjoint
  - [ ] rest: insideTangent, outsideTangent
  - [ ] composite energies
  - [ ] on all the polygons (convex, nonconvex, skinny polygons, holes, size difference?, different DOF, many segments)
- [ ] design realistic optimization test cases (diagrams + larger polygon tests)
- [ ] collab on improving the optimizer (diagram-specific opt strategies #120)
- [ ] port any analytic objectives/constraints needed
- [ ] finish the port and test it
  - [ ] rename new shapes, remove old shapes,  backwards compatibility / parity with old style programs, merge PR

Katherine

- [x] write up steps for adding a new transformed+polygonized shape (+ link diff)
- [x] Revise line search to use hessian
- [x] Take opt parameters at runtime
- [x] low-res poly curves
- [x] finish parallelograms
- [ ] write backend part for polygonizing text
- [x] write up discussion about polygonizing text
- [x] use new polygon type in code
- [ ] start porting our old examples
- [ ] profile the code and see what the bottleneck is (hessian? sampling?)
- [ ] work on top-down goals for graphics/optimization (realistic diagram goals for paper)
- [ ] Test how well newton's method works on a variety of shapes and energies
- [ ] store the default polygon, default bbox, default samples, as well as the three transformed things
  - [ ] don't polygonize curves on every step
- [ ] figure out what to do about constraints
- [ ] soup up optimizer
- [ ] Refactor optimizer code
[...]

PR #255, opened 2019-09-03
New arrow style

wodeni:
# Description

Related issues: #207 TOG-paper request

This PR adds another style for arrows, which is modeled after arrow #2 of Illustrator.

- Started using `auto-start-reverse` to avoid manual rotation of marker paths
- As a result, unified `Start/EndArrowhead` into just `Arrowhead` with auto-orientation
- Added another arrowhead style
- Arrow styles can now be added to a dictionary located in `Util.tsx` called `arrows`, and the `Arrowhead` component takes in an extra property `style: string` that specifies which arrowhead to use. The arrowheads are indexed by string values. The current convention is `arrowhead-[0-9]+`. The default value is `arrowhead-2`, which is the new style. 
- The arrowhead style property is currently not propagated to the Style level
  
image
image

# Checklist

# Open questions

The current implementation will not place arrowheads exactly at the end of the path, but rather placing the reference center (`refX, refY`) at the end of the path. This is a general problem with SVG markers, and here is a proposed implementation: https://www.w3.org/TR/svg-markers/#MarkerKnockout. 

This proposal doesn't seem to be implemented by any vendor yet.

### DIFF
--- a/opt-config/default.json
+++ b/opt-config/default.json
@@ -1,0 +1,1 @@
+{ "optMethod": "LBFGS" }
--- a/react-renderer/src/Util.tsx
+++ b/react-renderer/src/Util.tsx
@@ -3,48 +3,98 @@ import * as React from "react";
 declare const MathJax: any;
 import memoize from "fast-memoize";
 
+// const arrows = [
+//   {
+//     start: ,
+//     end:
+//   },
+//   {
+//     start:,
+//     end:
+//   }
+// ]
+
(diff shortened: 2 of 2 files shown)


hypotext:
I just re-added the opt config because the system fails at runtime otherwise (maybe it wasn't checked in during the big cleanup). Otherwise, I can reproduce, so it looks great. Thanks for adding the new arrowhead!

PR #267, opened 2019-10-01
Add hyperbolic geometry as a visualization style

hypotext:
# Description

This PR implements the teaser figure. Specifically, it adds almost-final Substance, Style, and Domain programs for the domain of (neutral) geometry in with three styles: hyperbolic, spherical, and Euclidean. This PR also adds minor system changes to make the Substance program look more natural.

Related issues: https://github.com/penrose/TOG-2019/issues/9 https://github.com/penrose/TOG-2019/issues/10 https://github.com/penrose/TOG-2019/issues/5

Basically, the system does a good amount of actual spherical or hyperbolic geometry in math space, converts math objects to 2D objects in screen space, and optimizes labels around those objects in screen space. For specifics:

- Spherical implementation documented on [wiki](https://github.com/penrose/penrose/wiki/The-geometry-domain)
- Hyperbolic implementation documented in [doc](https://docs.google.com/document/d/18H3ieQK3B3SlnwMDxDJUMEbzbCGfMa_E6A-mmUqEclg/edit)

[Hyperbolic diagram critique](https://github.com/penrose/TOG-2019/issues/5)

`runpenrose examples/geometry-domain/teaser.sub examples/geometry-domain/euclidean.sty examples/geometry-domain/geometry.dsl`

[Spherical diagram critique](https://github.com/penrose/TOG-2019/issues/9)

`runpenrose examples/geometry-domain/teaser.sub examples/geometry-domain/spherical.sty examples/geometry-domain/geometry.dsl`

[Euclidean diagram critique](https://github.com/penrose/TOG-2019/issues/10)

`runpenrose examples/geometry-domain/teaser.sub examples/geometry-domain/hyperbolic.sty examples/geometry-domain/geometry.dsl`

# Checklist

# Open questions

- [ ] Improve label placement (currently ad-hoc, slow, sometimes fails)
- [ ] Make hyperbolic layout faster
- [ ] Make sure segments/rays do not intersect each other
- [ ] Sphere visibility checks/shading

---

## Old working notes

Try this out in the IDE: https://use.penrose.ink/?branch=paper-example

`runpenrose examples/geometry-domain/teaser.sub examples/geometry-domain/hyperbolic.sty examples/geometry-domain/geometry.dsl`

image

Hyperbolic circles and angles (with different `p`)

`runpenrose examples/geometry-domain/simple.sub examples/geometry-domain/hyperbolic.sty examples/geometry-domain/geometry.dsl`

image

image

Older examples of teaser figure:

image

Example showing a hyperbolic triangle with three perpendicular bisectors:

`runpenrose examples/geometry-domain/simple.sub examples/geometry-domain/hyperbolic.sty examples/geometry-domain/geometry.dsl`

image

image

image

TODO:
- [x] improve explanation/caption in paper
- [x] angle mark
- [x] angle bisector
- [x] perpendicular mark
- [x] change range of `(x,y)` sampling so the triangles don't all look ideal
- [x] generate several samples
- [x] start diagram review with branch in IDE
- [x] add to paper teaser figure
- [x] clean up code
- [ ] document how domain works

Later
- [ ] circle in Substance?
- [ ] calculate ray length as hyperbolic distance properly

[Implementation notes](https://docs.google.com/document/d/18H3ieQK3B3SlnwMDxDJUMEbzbCGfMa_E6A-mmUqEclg/edit)

Other people's examples of right angles

**These are NOT Penrose diagrams**

https://math.stackexchange.com/questions/1250164/why-do-lines-in-the-poincare-model-meet-the-infinite-edge-at-right-angles
image

https://en.wikipedia.org/wiki/Hypercycle_(geometry)#/media/File:Hypercycle_(vector_format).svg
image

https://inspirehep.net/record/1642261/plots
image

Hyperbolic squares

https://mathcs.clarku.edu/~djoyce/poincare/poincare.html

image


hypotext:
@wodeni Could you give a quick once-over to the Substance parser and tokenizer changes in 125b4d5? The goal was to enable a Substance program like this: https://github.com/penrose/TOG-2019/issues/5#issuecomment-541485788

I did some light testing and it seems to desugar and parse declare-and-bind statements correctly, just not sure if it might break parsing for other Substance programs.

hypotext:
@wodeni When you get a chance, can you also give a once-over to the minor changes I just made to get bigger text in https://github.com/penrose/penrose/pull/267/commits/102997badcc3a09fba486073b4417f6335d806e0? The idea is that Style writers usually don't want to change the aspect ratio of a label. So I added a scaling property to text (that the user can set in Style), and the system calculates the final bbox of a piece of text (for optimization) as a computed property. A good place to start is by reading the `Shapes` diff which has a summary of the changes.

I made this change to address comments in this issue: https://github.com/penrose/TOG-2019/issues/5 Though is it me or does it still not scale quite uniformly?

The approach seems to work with your sets-3D example as well. I would also suggest to port that example to scale the text (and optimize the scaled text), rather than setting its absolute width and height, because if text has non-square aspect ratios then the labels will look weird.

examples/elementLibrary.json
```diff
@@ -16,5 +16,11 @@
     "label": "real analysis",
     "icon": "https://gist.githubusercontent.com/maxkrieger/65b0f7e2a151bb95e2913fda6ad7eaaa/raw/d5e80084cb5d907d3a2a5375c76d5c51ce8c30da/RealAnalysis.svg?sanitize=true",
     "uri": "real-analysis-domain/real-analysis.dsl"
+  },
+  {
+    "value": 3,
```
wodeni: Might need to resolve conflicts with other branches later

react-renderer/src/Label.tsx
```diff
@@ -6,36 +6,39 @@ import { IGPIPropsDraggable } from "./types";
 const styleLabel = (
   label: HTMLElement,
   color: string,
-  w: string,
-  h: string
+  finalW: string,
```
wodeni: Are we keeping this change? I thought master deals with it properly now.

react-renderer/src/layers/BoundingBoxLayer.tsx
```diff
@@ -14,14 +14,15 @@ class BoundingBoxLayer extends React.Component<ILayerProps> {
               [shape.x.contents, shape.y.contents],
               canvasSize
             );
-            const {w, h} = shape;
+            // const {w, h} = shape;
+            const {finalW, finalH} = shape;
```
wodeni: Similar to my comment `Label.tsx`: consider reverting these changes

src/Penrose/Optimizer.hs
```diff
@@ -79,8 +79,8 @@ useAutodiff = True
 constT :: Floating a => a
 constT = 0.001
 
--- debugOpt = True
-debugOpt = False
+debugOpt = True
```
wodeni: Revert to `False` for deployment: Console IO affects performance quite negatively

src/Penrose/Substance.hs
```diff
@@ -83,6 +83,10 @@ instance Show Predicate where
         where nString = show predicateName
               aString = show predicateArgs
 
+-- | Special construct added for "declare and bind" statements, which get parsed into a declare and then a bind in the Substance core language
```
wodeni: I would generally avoid nesting the AST like this for (1) readability and (2) easier post-processing. In addition, I thought only `Expr`s are allowed on the right-hand side, so It would be cleaner to add `DeclBind T Var Expr` to the `SubStmt` type.

src/Penrose/Substance.hs
```diff
@@ -152,12 +156,27 @@ convertDeclList t vars var = vars ++ [Decl t var]
...
+subStmt' = tryChoice [declBind,
+                      singleSubStmt]
+
+declBind :: Parser SubStmt'
+declBind = do t <- tParser
+              var <- varParser
+              rword ":="
+              DeclBind t var <$> exprParser
+
+singleSubStmt :: Parser SubStmt'
```
wodeni: Surprised that this compiled? The `do` is probably redundant here. If you agree with the AST change, we don't even have to write these functions altogether.

src/Penrose/Substance.hs
```diff
@@ -716,14 +735,19 @@ subSeparate = foldr separate ([], [])
                              (LD x) -> (x : decls, constrs)
                              (LC x) -> (decls, x : constrs)
 
+-- | Convert a "declare and bind" to a "declare, then bind" as separate statements
+toSubCore :: [SubStmt'] -> [SubStmt]
+toSubCore stmts = concatMap toCoreStmt stmts
```
wodeni: This can be done without the extra `SubStmt'` type by pattern matching on `DeclBind` only.

src/Penrose/GenOptProblem.hs
```diff
@@ -776,13 +776,15 @@ evalEnergyOn s vstate =
...
     in applyCombined penaltyWeight fnsE
 
-evalEnergy :: (Autofloat a) => State -> a    
-evalEnergy s = 
-    let varyMap = mkVaryMap (varyingPaths s) (map r2f $ varyingState s) 
-        fns = objFns s ++ constrFns s
-        (fnsE, transE, rng') = evalFns evalIterRange fns (castTranslation $ transr s) varyMap (rng s)
-        penaltyWeight = r2f $ weight $ paramsr s
-    in applyCombined penaltyWeight fnsE
+-- TODO: This doesn't seem to be used anywhere. Why is it here?
```
wodeni: This is meant to be an API call exposed to our future benchmarking frontend, so I'd say keep it.

src/Penrose/Substance.hs
hypotext: In principle I would actually prefer both `DeclBind` and `DeclList` to be separate syntactic forms from `SubStmt` because they are convenience statements that aren't part of "Substance Core" (i.e. can already be expressed by combinations of Substance Core statements). Pragmatically, if you still think `DeclBind` should go in `SubStmt`, then we can move it there because it makes the code simpler. What do you think?

wodeni: Unless we plan to do anything other than deconstruction to the non-core statements, I think it would be better to keep it simple and avoid the unnecessary code. In general, it's a bit easier to debug the parser if the AST structure is simpler. If such changes add any useful information, I would leave it in (e.g. making pretty printing easier). In this case, I don't think we are gaining anything more than an extra layer to unwrap later.

hypotext: Another potential issue with putting those extra forms into Substance Core is that if we are analyzing the Substance Core AST but always compiling the "extra" Substance forms into other Substance forms, then we will have two pattern-match branches that are always unused in the SubStmt type, which can lead to runtime errors.

hypotext: Again, I think the `DeclList` and `DeclBind` are convenience forms that are mainly for the user and parser, and should be kept separate from the core grammar of the language. I see that `refineDeclList` is basically doing the same thing with turning the `DeclList` convenience form into a list of declarations and so I would rather move `DeclList` to `SubStmt'` as well.

So, I think I am not personally going to revert these Substance changes. However, if it's important to you as someone who spends more time working with the parser (e.g. if the different Substance form affects the IDE typeahead), then you should go ahead and revert the changes and I won't complain further about it 🙂

hypotext:
@wodeni Thanks for the review! I resolved most of your comments (though see above for discussion of the Substance grammar). I also cleaned up the code and documented the PR. I think this PR is ready to merge. Let me know if you want to make Substance changes first.

wodeni:
> @wodeni Thanks for the review! I resolved most of your comments (though see above for discussion of the Substance grammar). I also cleaned up the code and documented the PR. I think this PR is ready to merge. Let me know if you want to make Substance changes first.

Looks great! Thanks for addressing the comments. Now that we treat `DeclList` and `DeclBind` as the same kind, I would agree with your reasoning there. I do agree that we do not want two unused branches when processing the "core AST". I'll merge `DeclList` into the `SubStmt'` type and merge the PR after.

hypotext:
Awesome, thanks @wodeni! I think that change will make things cleaner (and set us up for more Substance sugar if needed). Feel free to merge after you've made the changes and tested them.

Issue #269, opened 2019-10-03
Add 3D style for set theory


hypotext:
From @keenancrane:

In our meeting today I raised the question of how to cook up some examples where people won't say, "What's the point? I could just do that by hand."

One way to construct this kind of example is to make it more complicated, i.e., more pieces that have to be arranged in an intricate way.  That we should be able to do by writing some more elaborate Substance programs.

Another axis that we haven't really explored—but is probably worth doing at least one example for the paper—is amping up the aesthetics in a way that is superficial at a programmatic level, but a pain to execute by hand.  For example, I took Nimo's latest subset diagram, and added some nice shading, styling, etc.:

![image](https://user-images.githubusercontent.com/2762356/66143117-a4aadf80-e5d4-11e9-974f-910b9e8e0e51.png)

In Illustrator, this was a pain in the butt and took me ~20 minutes.  In Penrose, this is a simple matter of drawing a pre-baked semi-transparent SVG file on top of the existing circles.

![image](https://user-images.githubusercontent.com/2762356/66143131-aa082a00-e5d4-11e9-8107-b05509ab40ba.png)

Seeing this kind of example in the paper could go a long way toward giving people the impression that you can do some really sophisticated, nice-looking stuff in Penrose—and not just something you can easily do in tools like TikZ, Inkscape, etc.

P.S. The other tweak I made was making the text size a fraction of the disk size.  This is trivial in Style, but painful by hand.  In general this sounds like a good mantra when thinking about things to do with examples for the paper:

_"Trivial in Penrose; painful by hand"_

wodeni:
Resolved in #273.

Issue #60, opened 2018-04-05
Enable Penrose to generate 3D diagrams


hypotext:
This issue has two goals:

* Improve and showcase the generality of our system's rendering frontend.
* Generate more beautiful and useful visualizations for particular domains that are well-suited to 3D diagrams.

Some first steps:

* To ramp up, first add a 3D Style and renderer for existing domains we've implemented, e.g. set theory and linear algebra. Step number one is to create a version of `venn.sty` where the circles are now semi-transparent spheres.
* The Substance programs should largely stay the same. The Style programs will change to use 3D types/functions/objectives and to specify a renderer. For example, any sampling mechanisms that used types like a 2D vector or matrix should be generalized.

More specifically:

* Add host language datatypes that support optimizing in 3D
  * Possibly split out and generalize rendering functions for each shape
* Write objectives and constraints for optimizing in 3D
* Write styles for 3D (e.g. bold outlines, soft shadows, contour lines, 3D surfaces, lighting)
* Output the optimized world as a 3D scene in some format for some renderer. Perhaps output COLLADA (an XML-based schema for 3D scenes), POVRAY, or the [pbrt input format](http://www.pbrt.org/fileformat-v3.html)
* Optimize and project labels onto the rendered image

Later:

* Add new Substance domains that benefit from 3D diagrams and scenes, such as geometry, topology, meshes, raytracing, and optimization, and repeat the steps above
* Write documentation, examples, and tests. Especially document how to generally extend the system with a new renderer (3D or not).
* Possible research questions: optimizing curves on surfaces, creating illustrative rendering styles (à la [Blender Freestyle](https://docs.blender.org/manual/en/dev/render/freestyle/introduction.html#what-is-freestyle)), sampling 3D diagrams

Stretch goal:
* Output to VR?! Generate something like the [Venn Diagram Museum](http://elevr.com/venn-diagram-museum/)

To look into:
* What concrete examples/classes of well-known, compelling, and useful 3D diagrams could we aim to create?
* What renderers could we use, and with what output formats? What are the pros and cons of each?
* What other related work is there? 
* What research questions might we explore?

More internal discussion, and links to related work, can be found [here](https://docs.google.com/document/d/1ym9rg3vD6lG0wQKCJK1bq0bp3gZFdNGD_FRgoJ--Xss/edit).

PR #273, opened 2019-10-09
3D Venn style for sets

wodeni:
# Description

Related issues: #269 #60 

Using the strategy described in #269, created a "3D" style for Venn diagrams for sets.

* Fixed some bugs in the backend and frontend for pending properties: the backend finds pending paths _after_ `initShape`, because that's the function that populate all the `Pending` expressions in the translation. The frontend now only updates label dimensions when they are not set explicitly in Style. 
* Label dimensions are now changeable by setting `width` and `height` of the `<svg>` element.

`penrose set-theory-domain/tree.sub set-theory-domain/venn-3d.sty set-theory-domain/setTheory.dsl`
* Without variable label sizes:
3d-sets
image (1)

* With variable label sizes:
3d-sets-2
image

# Checklist

# Open questions


wodeni:
Added the example to the IDE and tested other existing examples. Ready to merge

hypotext:
Was able to repro and the code generally looks good. Thanks!

> the backend finds pending paths after initShape, because that's the function that populate all the Pending expressions in the translation.

That definitely seems like a bug. I'm surprised that it didn't cause problems earlier.

Notes
* The examples take a little longer to optimize than with the "flat" style. Wondering if it's due to the label sizes change.
* Also takes unusually long to resample.

wodeni:
Make sure you have your opt-flag on (i.e. no `--fast`). The label dimensions might be the cause. The resample time should be the same. If it's still longer with the same binary, we can look into that.

hypotext:
Yeah, I just rebuilt on master with `stack install --profile` and it still optimizes and samples slower than the "larger venn" example in the IDE. Maybe it's the profile flag? Does it feel a little slower on your machine too?

Issue #315, opened 2020-01-18
Use MathJax 3


maxkrieger:
Eliminates the need for global scoping, weird configs, etc. See https://github.com/mathjax/MathJax-demos-node/blob/master/direct/tex2svg for usage example.

PR #327, opened 2020-03-17
[mathjax-3] Change Mathjax runtime to 3.0 (#315)

maxkrieger:
(#315) This PR upgrades MathJax to the npm-packaged 3.0 version. Eliminates the need for global variables, and is way faster (it's not even async anymore, it's synchronous).

The biggest change is that I've lifted the `collectLabels` related functions out into their own file `react-renderer/src/utils/CollectLabels.tsx` from the existing `Util.tsx` file.

There is a strange issue, at least with the 3d style. It seems the dimensions aren't being understood by the server:
image

@wodeni any immediate ideas why? ^


maxkrieger:
It looks like the issue is independent of this PR. It's worth cross checking on other examples before merging though.

PR #333, opened 2020-03-24
Fix continuous map example.

joshpoll:
# Description

f^{-1} should be in the domain, not the codomain. Per Munkres: "A function f \colon X \to Y is said to be continuous if for each open subset V of Y, the set f^{-1}(V) is an open subset of X."

Before:
continuous map before

After:
continuous map after

# Checklist

- [x] I ran [Haddock](https://github.com/penrose/penrose/wiki/Writing-and-generating-documentation#generating-html-documentation-site) and there were no errors when generating the HTML site

### DIFF
--- a/examples/set-theory-domain/continuousmap.sub
+++ b/examples/set-theory-domain/continuousmap.sub
@@ -2,6 +2,7 @@ AutoLabel All
 
 Set A
 Set U
+Label U $f^{-1}(V)$
 Set Rn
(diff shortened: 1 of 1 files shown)


wodeni:
Thanks for revisiting this and actually throwing in the fix 😄! Merging it to `master`.

hypotext:
Thanks @joshpoll!

PR #336, opened 2020-04-03
Porting runtime to the web

wodeni:
# Description

Related issues: none

This PR will eventually contain a web runtime for Penrose. Short-term goals include:
* [x] TypeScript types for various data structures for easier pattern-matching in the evaluator
* [x] Evaluator: given a serialized `State`, decode it into some data structure with type, and evaluate the translation in the state. Top level function `evalTranslation(s : State) : Shapes`
* [x] A subset of computations as arbitrary TS functions

Longer-term goals:
- [x] Optimizer: progress below. Close to done
- [x] A subset of constraints in TensorFlow.js compatible format
- [x]  If needed, port the computations to the TensorFlow format as well

See additional documentation on this [wiki page](https://github.com/penrose/penrose/wiki/Contributing-to-the-web-runtime).

* This branch has more npm dependencies, so you'll have to run `npm install`. 
* If you run into `fsevents` related errors, try `brew install watchman`, which did the trick for me on Mac OS X Catalina

* TS definitions from Haskell: `stack install && tsdef > types.d.ts`
* Tests for evaluator: `npm test` in `react-renderer`

# Checklist

# Open questions

## Backend runtime related

* While the design of `ArgVal` seems convenient for differentiating between `GPI` and `Val`, I think it's cleaner if we just treat it as an expression since the GPI values can just be treated as yet another data structure similar to lists. Obviously a few things to consider:
```
-- Synonyms of GPIs: can a GPI be reused? Is it copied as reference or value?
global { defaultRect = Rectangle { ... } }
Set A { A.shape = global.defaultRect }
Set B { B.shape = global.defaultRect }

-- As discussed a long time before, GPIs as outputs of computation functions
Set A { 
  A.shape = computeRect(...) -- returns Rectangle. Does `computeRect` have to compute all properties?
  A.shape.x = 3.0 -- Does this take precedance over the previous computation?
```


wodeni:
In the case of `BindingForm`, I think it's just a bit of redundancy in the original Haskell types, and we should fix that soon. For the `*Fn`s, they are necessary because the string is the actual name of the function, e.g. `repel`, `center`, ...

wodeni:
Some more types here. After manually encoding `Data.Map.Strict.Map a b` as `Map<a, b>`. Only `Map` and `Record` allow non-string keys, and I'm using `Map` now for benefits noted in an [SO post](https://stackoverflow.com/questions/18541940/map-vs-object-in-javascript)
```
type Translation<T> = ITrans<T>;

interface ITrans<T> {
  trMap: Map<Name, Map<string, FieldExpr<T>>>;
  warnings: string[];
}

type FieldExpr<T> = IFExpr<T> | IFGPI<T>;

interface IFExpr<T> {
  tag: "FExpr";
  contents: TagExpr<T>;
}

interface IFGPI<T> {
  tag: "FGPI";
  contents: [string, Map<string, TagExpr<T>>];
}

type TagExpr<T> = IOptEval<T> | IDone<T> | IPending<T>;

interface IOptEval<T> {
  tag: "OptEval";
  contents: Expr;
}

interface IDone<T> {
  tag: "Done";
  contents: Value<T>;
}

interface IPending<T> {
  tag: "Pending";
  contents: Value<T>;
}
```

wodeni:
Turns out that decoding serialized JSON `state` requires quite a bit of manual effort (after double-checking with @maxkrieger it seems to be necessary). Some tests in https://github.com/penrose/penrose/blob/95edd0e6d85796f3f7dd3b519e5183360f8eb02a/react-renderer/src/__tests__/Evaluator.test.ts.

I ended up using typedoc over tsdoc for it's automation:
image

wodeni:
# First working example of the evaluator

The evaluator now works on `venn-3d` with the backend optimizer:
* I disabled the backend `evalTranslation` altogether
* Now every time the frontend receives a packet, I will decode it into our TS type `IState`
* I wrote a TS version of `evalTranslation` and call it on the new `IState` object, which evaluates and returns all shapes
* This `IState` type is now enforced throughout the frontend renderer, instead of an `any` type JSON schema as before. As a result, most of the hacky JSON lookup and manipulation are now replaced by typed functions. 
* Technically we can throw away `shapesr` (fully evaluated list of shapes stored in the backend), but I decided to keep it since somehow the resampling logic uses them. 
* However, since we need the backend to deserialize the `State` that the frontend sends, I had to encode the `IState` object back to the backend schema. There are some minor schema changes that only exists in the frontend, so this is necessary.
* __TL;DR: the frontend now evaluates the shapes by itself  (`decode -> eval -> encode` per step)__  

Screen Recording 2020-04-24 at 05 20 52

## Debugging

Turns out that moving to TS lets you use all the nice tools like VSCode + Chrome debugger and all the Chrome profiling features (see next section).
However, you will need to be careful about how objects behave in TS. I encounter two major things:
* Dynamically adding props to a typed object cannot be checked. If you insert a property at the wrong place and want to access it later, you will find the old value instead.
* The spread syntax for constructing objects `{...obj, prop1: newprop}` requires the spread to be __before__ the new props. 

## Performance

I was a little curious about how well this new evaluator performs and the additional cost of encoding and decoding. So I ran the Chrome profiler and the React profiler. Here's the data for the former:
https://gist.github.com/wodeni/00dbcc34918118b31257d096d88a4f56

image
image

Turns out that `evalTranslation` and `decodeState` are reasonably fast. I'm sure we'll find this tool to be much more usable than the Haskell counterpart.

maxkrieger:
Beautiful!! Now get some sleep man 😂

hypotext:
Amazing!! Excited to integrate with the new optimizer. 

Some questions that come to mind: 

* How close are we to totally deprecating the backend evaluator for all examples?
* Does this evaluator solve our performance problem with `computedProperties` in the former backend evaluator?
* How does the wall-clock time to convergence change with the two different evaluators?
* Can we document clearly what data is immutable or mutable in the frontend (i.e. what the design decisions were)? Can we enforce guarantees on mutability, immutability, staleness, updates etc.? 
* Will the types in the evaluator need to be tf.js variables, and if so, how does that affect the modularity and performance?

No need to answer all of these; we can discuss on Wed.

Other thoughts:

* I'm particularly excited to use Chrome debugger/profiling on our new system, and would appreciate any recs for resources here.
* At some point I would also love to have a "grand tour" of the frontend, since I'm not clear on how the evaluator integrates with all the other types / architecture in general.

wodeni:
> * How close are we to totally deprecating the backend evaluator for all examples?
We will need all computation functions. I'll pick two more examples to repro for debugging purposes, but we should do more together!

> * Does this evaluator solve our performance problem with `computedProperties` in the former backend evaluator?
The evaluation strategy is still the same, but we don't have to worry about polymorphism if that was problematic at all. Currently I disabled computed props in the backend because the evaluator seems to compute them at every step (which makes sense). If we want them back in the frontend, we will have to port them as well.

> * How does the wall-clock time to convergence change with the two different evaluators?
Visually faster for my sleepy eyes last night, but hard to tell without bigger examples/better measurements.

> * Can we document clearly what data is immutable or mutable in the frontend (i.e. what the design decisions were)? Can we enforce guarantees on mutability, immutability, staleness, updates etc.?
This is something I would like to do more systematically, perhaps with our new typedoc documentation standard. Right now they are all in docstrings for each of the exported functions I wrote.

> * Will the types in the evaluator need to be tf.js variables, and if so, how does that affect the modularity and performance? How are tf.js vars modeled in Typescript?
As you know, those types are generic, so theoretically possible. I think it might be better if we try to limit TFjs types only in the (currently nonexistent) optimizer module and see how it works. I see propagating those types to the evaluator as the fallback option, but obviously I don't know much about TFjs atm. 

> * I'm particularly excited to use Chrome debugger/profiling on our new system, and would appreciate any recs for resources here.
I just started using them. Feel free to ping me if you wanna try them out! You can import the file I shared in chrome and repro the profiling results.

wodeni:
# Reproducing 3 more examples

I repro'ed 3 more examples: `venn`, `tree`, and `twoVectors`. Results below:
Screen Recording 2020-04-26 at 00 05 19
Screen Recording 2020-04-25 at 23 58 13
Screen Recording 2020-04-26 at 01 27 43

* Performance seems pretty good. I haven't done serious comparisons yet and the computations are pretty small. 
* I ported several functions like `hsva`, `sin` and `cos`. They all worked as intended except for `sampleColor`. My current strategy for the random generator is to seed a global generator with the backend seed every time I decode a `State` from the backend. I do get consistent samples, but since we are not caching results of computations now, you can't get consistent colors when two properties refer to the same color, e.g.
```
A.shape.color = sampleColor() // computed once
B.shape.color = A.shape.color // call sampleColor() again because there's no caching!
```
* This should be easy to fix, but I'd like to optimize the evaluator once everything becomes a bit more stable.
* New computation functions are here:
https://github.com/penrose/penrose/blob/62bd76577fddae597ac1d791e76f2c16171ae353/react-renderer/src/Evaluator.ts#L52

wodeni:
# First working version of the frontend optimizer

Pretty excited to see all-frontend-optimized Penrose diagrams:

Screen Recording 2020-04-28 at 06 55 21

* I tested the new optimizer with a simplified `venn` Style program and the good o `tree.sub` Substance program. This Style program does not contain any computation and only has `min/maxSize`, `contains` (with padding), and `disjoint`. Here's an example of contains:
```ts
{ // the constraint dict
  contains: (
    [t1, s1]: [string, any],
    [t2, s2]: [string, any],
    offset: Tensor
  ) => {
    if (t1 === "Circle" && t2 === "Circle") {
      const d = dist(center(s1), center(s2));
      const o = offset
        ? s1.r.contents.sub(s2.r.contents).sub(offset)
        : s1.r.contents.sub(s2.r.contents);
      return d.sub(o);
    } else throw new Error(`${[t1, t2]} not supported for contains`);
  },
}
```
* The new optimizer does require all numbers that are involved (more on this later) in computations and objectives to be TF types like `Scalar` or `Tensor`. As a result, `contains` is written in a very different way from regular JS functions. `dist` and `center` are also TF functions:
```ts
export const center = (props: any): Tensor =>
  tf.stack([props.x.contents, props.y.contents]);
export const dist = (p1: Tensor, p2: Tensor) => p1.sub(p2).norm(); 
```

## tfjs gotchas
* (with @hypotext) we found that `tf.grads` requires __all parameters to the energy function to be "used" at least once__, which means that `(x, y, z) => x.sub(y)` will not be accepted by tfjs. This is due to a hidden setting in the tfjs engine, which prohibits variables without gradients by default. [Source](https://github.com/tensorflow/tfjs/blob/b72893a610091e70aba180f0fc26fe72d4e6be30/tfjs-core/src/engine.ts#L972)
* This turns out to be problematic for Penrose because all degrees of freedom in a Penrose diagram are marked as `Varying`, which gets translated to tfjs `Variable`. However, it's up to the users to decide which degrees of freedom to use in their Style program. There's no way for the backend to easily work out the variables actually got used in the constraints (except for tracing all computations and objectives). 
* To work around this, we "fake" an energy function that uses all varying variables in our system by adding and subtracting a dummy value like below. This does not change the value nor gradients of the overall energy, but pretend to involve all variables.
```ts
    const dummyVal = stack(varyingValuesTF).sum();
    return overallEng.sub(dummyVal).add(dummyVal);
```
* After this, I found that all my variables have zero gradients, which turned out to be an issue with `VaryMap`, which is a map from paths to varying values and their corresponding tfjs variables. This map is indexed by a JS object, and the lookup fails because JS apparently has problems comparing objects. The important thing is: tfjs identify their variables by an `id` in the `Variable` object. As a result, __once you convert numbers to tfjs variables, you have to use the same ones throughout. __ I was able to fix the zero gradients by stringifying the keys to `VaryMap`.
* The current optimizer is a wrapper around an existing tfjs optimizer (`adam`), which turns out to be quite slow with our current (naive) implementation. Maybe this is not the right opt method to use. I'll wait a bit to hear more from @hypotext about porting LBFGS to tfjs.

## Frontend changes

* As we move closer to a functional web runtime, I'd like to think about how the renderer and IDE would change. 
* @maxkrieger I had a pretty novice react problem: now that the frontend has to determine whether to repeatedly call `step` depending on the value of `autostep`, we need to determine the timing of this action. I tried to do this in `componentDidUpdate`, but rendering freezes all the way until convergence, which is not expected. What's the right way to sequentially update a react state? I tried `setState` callbacks, various `async` annotations, and wrapping `setState` with `Promise` without any success. That's why I'm clicking through the optimization in the demo. Let me know!

react-renderer/src/Evaluator.ts
```diff
@@ -7,11 +7,12 @@ import {
   mapKeys,
   concat,
   PropertyPath,
+  zip,
 } from "lodash";
 import { mapValues } from "lodash";
 import { dist, randFloat } from "./Util";
 import seedrandom from "seedrandom";
-import { Tensor, Variable, scalar, pad2d } from "@tensorflow/tfjs-node";
+import { Tensor, Variable, scalar, pad2d } from "@tensorflow/tfjs";
```
hypotext, line 15: Is there a clear difference between the tfjs node version vs pure js version? 

Regarding backends, if possible we should use the [react native](https://github.com/tensorflow/tfjs/tree/master/tfjs-react-native) version of tfjs, since it supports model inference and training via GPU/WebGL, which is the fastest option. Does our current frontend work with react native?

There is also an alpha wasm backend [here](https://github.com/tensorflow/tfjs/tree/master/tfjs-backend-wasm) that only supports inference (so we can't use it, since we want "training").

In terms of speed, the wasm page has useful benchmarks for specific models. Basically plain JS is the slowest. Quoting:

* "For this model, the WASM backend is between ~3X-11.5X faster than the plain JS backend, and ~5.3-7.7X slower than the WebGL backend."
* "You should always try to use the WASM backend over the plain JS backend since it is strictly faster on all devices, across all model sizes. Compared to the WebGL backend, the WASM backend has better numerical stability, and wider device support. Performance-wise, our benchmarks show that:
  * "For medium-sized models (~100-500M multiply-adds), the WASM backend is several times slower than the WebGL backend."
  * "For lite models (~20-60M multiply-adds), the WASM backend has comparable performance to the WebGL backend"

maxkrieger: @hypotext React native is a runtime for react that runs on mobile devices, so I assume tfjs-rn is based on native code and hooks into mobile-specific chips. I would expect hardware acceleration to still be available with the tfjs we use now, check this page: https://www.tensorflow.org/js/guide/platform_environment#backends

react-renderer/src/Evaluator.ts
```diff
@@ -285,7 +289,7 @@ export const evalExpr = (
   }
 };
 
-const toTF = (v: Value<number>): Value<Tensor> => {
+const differentiableValue = (v: Value<number>): Value<Tensor> => {
```
hypotext, line 292: Should the return type be `Value<Variable>`?

react-renderer/src/Evaluator.ts
hypotext: That's a better source, thanks @maxkrieger. Yes, let's just make sure we use the WebGL backend, and do the memory management that that requires (I'm happy to work on this)

react-renderer/src/Evaluator.ts
```diff
@@ -339,10 +343,12 @@ export const resolvePath = (
...
               ], // TODO: check if this is true
             };
-            varyingVal = varyingMap?.get(propPath);
+            varyingVal = varyingMap?.get(JSON.stringify(propPath));
             if (varyingVal) {
-              return { tag: "FloatV", contents: varyingVal[1] };
-            } else return tf ? toTF(p.contents) : p.contents;
+              return { tag: "FloatV", contents: varyingVal };
+            } else {
+              return autodiff ? differentiableValue(p.contents) : p.contents;
```
hypotext, line 350: I like this approach. I wonder if tf.js might be somehow faster at evaluating the forward pass though? (i.e. maybe everything can be a tensor but not necessarily a variable) It also feels inefficient to pack one number per tensor. Could be performance opt for later

maxkrieger:
@wodeni `componentDidUpdate` is indeed a risky place to set state in in because it fires whenever state is set! So if you setState inside it, you'll get an infinite loop when unguarded. The react lifecycle, I bet, won't perform a render until it has a reliable state. [This page](https://reactjs.org/docs/react-component.html) is pretty comprehensive when it comes to gotchas for the lifecycle.

If I understand the issue correctly, you're storing the opt state in react state and want to listen for when a step completes to compute the next step. So this doesn't need to be different from client-server eventing?

It sounds like the tfjs side of things should be async, and right now it's synchronous and you're relying on react to listen for the synch stuff to catch up. Does tfjs have an async way to perform a step? And in the "callback", that's where you can see the state and decide whether to perform a recursive call.

maxkrieger:
Tldr: separate react from tf as much as possible by treating tf as a "server" that asynchronously dispatches updates to the client. Neither React _nor_ tf need to be the source of truth for opt. Let them ping back and forth reactively ;)

react-renderer/src/Optimizer.ts
```diff
@@ -114,53 +107,69 @@ export const evalEnergyOn = (state: State) => {
...
-    ) as VaryMap
+      xs.map((x) => tfStr(x))
+    ) as [Path, number][]
   );
   const newState = {
     ...state,
     translation: trans,
     varyingValues: xs.map(tfStr), // TODO: need this?
   };
+  if (tfStr(energy) > 10) {
```
hypotext, line 166: I know this is a heuristic, but the right thing to check is the norm of the gradient being above some threshold, as in the usual convergence check

react-renderer/src/Optimizer.ts
```diff
@@ -71,9 +65,9 @@ export const constrDict = {
   },
 };
 
-export const center = (props: any): Tensor1D =>
-  tf.stack([props.x.contents, props.y.contents]) as Tensor1D; // HACK: need to annotate the types of x and y to be Tensor
-export const dist = (p1: Tensor1D, p2: Tensor1D) => p1.sub(p2).norm(); // NOTE: or tf.squaredDifference
+export const center = (props: any): Tensor =>
```
hypotext, line 68: What's the relationship between Tensor and Variable types in typescript? Is it doing "permissive subtyping" (any variable can be used as a tensor) or would it ever coerce the variable type either way?

maxkrieger:
I suggest finding some react projects that integrate tfjs to get some ideas on how to manage the state weirdness. Also check this section, we want to make sure things are as lazy/thunkable as possible for maximum performance:

https://www.tensorflow.org/js/guide/platform_environment#avoid_blocking_the_ui_thread

> This makes it important to use the asynchronous x.data() and x.array() methods over their synchronous counterparts x.dataSync() and x.arraySync() to avoid blocking the UI thread while the computation completes.

react-renderer/src/Optimizer.ts
```diff
@@ -114,53 +107,69 @@ export const evalEnergyOn = (state: State) => {
...
-    //     .mul(0)
-    // );
-
-    // return objEng.add(
-    //   constrEng.mul(tf.scalar(constraintWeight * state.paramsr.weight))
-    // );
+    // NOTE: the current version of tfjs requires all input variables to have gradients (i.e. actually involved when computing the overall energy). See https://github.com/tensorflow/tfjs-core/blob/8c2d9e05643988fa7f4575c30a5ad3e732d189b2/tfjs-core/src/engine.ts#L726
+    // HACK: therefore, we try to work around it by using all varying values without affecting the value and gradients of the energy function
+    const dummyVal = stack(varyingValuesTF).sum();
+    return overallEng.sub(dummyVal).add(dummyVal);
```
hypotext, line 129: I know this is an interim hack until we can fork tfjs, but these lines might make gradient computation slower by adding extra operations that involve all variables. Not sure how much slower since I don't know how the AD is implemented.

maxkrieger: Btw, is there any floating point danger here, where the add/sub will actually create a tiny gradient? Could that be avoided by adding zero instead, multiplying by 1, etc?

hypotext: Yeah, my suggestion is to do something like `overallEng + sum(vector_of_zeros `dot` varyingValuesTF)` (maybe with a `stack` in there). Multiplying by zero is a very safe thing to do with floats!

hypotext:
So awesome to see the new Penrose engine working! ⚙️ Thanks for the writeup!

Some more comments:

* The schema for each shape is still defined in the backend, right? When is a good time to move that to the frontend?
* Is there some way to get rid of the `.contents` nesting to make things more concise?
* Interesting that the `contains` constraint works without the full EP method increasing the penalty, but just a constant penalty. Is that right?

On Wednesday it will be great if we can do some profiling together.

maxkrieger:
> * Is there some way to get rid of the `.contents` nesting to make things more concise?

There absolutely is, I believe @wodeni is keeping it that way for now to be backwards-compatible (this is an artifact of aeson).

react-renderer/src/Evaluator.ts
wodeni: I’m aware of these backend options and saw some resource management tricks. You should feel free to try them out! My purpose on this branch is to get the optimizer working with the evaluator, so performance wasn’t my main concern

react-renderer/src/Evaluator.ts
wodeni: Yeah, Tensor is fine for typechecking and doesn’t really do anything to the actual values. I’ll change it

react-renderer/src/Evaluator.ts
wodeni: I wanted to minimize the use of tensors and intentionally left the evaluator as close to the plain JS version as possible, but I’m happy to chat more about opt a bit later

react-renderer/src/Optimizer.ts
wodeni: Variable extends Tensor, so the latter thing you said

react-renderer/src/Optimizer.ts
wodeni: Will do!

wodeni:
> Tldr: separate react from tf as much as possible by treating tf as a "server" that asynchronously dispatches updates to the client. Neither React _nor_ tf need to be the source of truth for opt. Let them ping back and forth reactively ;)

Yeah I would agree. Let me know if you come across a good way to do that. I'm not sure if this is something that the tfjs API supports.

wodeni:
> * The schema for each shape is still defined in the backend, right? When is a good time to move that to the frontend?
Yep, it's a bit more involved because of property sampling and rng management, which I haven't worked on.

> * Interesting that the `contains` constraint works without the full EP method increasing the penalty, but just a constant penalty. Is that right?

Correct, just a content penalty.

react-renderer/src/Optimizer.ts
wodeni: Need to be careful about `dataSync` tho. I wonder if there's a way to do convergence check without syncing with the CPU at all

maxkrieger:
> > Tldr: separate react from tf as much as possible by treating tf as a "server" that asynchronously dispatches updates to the client. Neither React _nor_ tf need to be the source of truth for opt. Let them ping back and forth reactively ;)
> 
> Yeah I would agree. Let me know if you come across a good way to do that. I'm not sure if this is something that the tfjs API supports.

Either way, the immediate fix is to wrap the tfjs synch call in a callback/promise function, and use it as I suggest:

> the React state should be a _side effect_, a preview, of an asynchronous recursive callstack

So, make a separate component method that doesn't rely on the component state to send to the tfjs, but rather its own param that gets piped through at each recursive call

wodeni:
# No GPUs, Please!

Had a discussion with @maxkrieger and @hypotext and we randomly tried to set the tfjs backend to `cpu` instead of `webgl`, and... things sped up 50x (don't actually know the number, but very visible). 
__Default backend (with GPU)__
Screen Recording 2020-04-30 at 07 17 33
__CPU-only backend__
Screen Recording 2020-04-30 at 07 20 34

## Penrose objectives vs. typical loss functions

## Other improvements

* I moved the constraints and objectives to `Constraints` module for better readability, and added a few for the venn example. 
* With one additional objective (`sameCenter` for labels and circles), the optimizer takes forever to converge. This is due to the lack of dynamic weights in the current version. So we will need to port:
  * Weight changes
  * Efficient test on the normal of gradients for EP convergence test 
* Label collection is back (still hacky but to fix with @maxkrieger ).
* Made some more React changes attempting to get our "animation" back without success, but I managed to force it with `setTimeout(1)`
* I played around with different learning rates, which does affect the time to convergence and steps taken. Would be nice to have a nice module in which we manage these things. 

Here's the same example, but with more constraints and labels:

Screen Recording 2020-04-30 at 07 11 26

react-renderer/src/Evaluator.ts
wodeni: Turns out that not all values are `Variable`s. So I'll keep this for now.

hypotext:
Thanks for starting the EP template @wodeni. Quick Qs: Did you already test it? Are there specific problems / weird behavior that I should make sure to eliminate in my implementation, or things that changed in the web version? If not, I'll just do a 1:1 port of what's in `stepShapes`.

wodeni:
> Thanks for starting the EP template @wodeni. Quick Qs: Did you already test it? Are there specific problems / weird behavior that I should make sure to eliminate in my implementation, or things that changed in the web version? If not, I'll just do a 1:1 port of what's in `stepShapes`.

I put the EP version of step into stepEP, which is not tested. Just keep in mind how tfjs mutates the varying state and decide how you want to keep track of them. Since there’s no line search, “steps” might have a different meaning in the web version. I’m not sure if it’s something we should port as well. The parameters to Adam are also something I’m not sure about.

Finally, as usual, always look out for performance traps like converting things to tensors. Hopefully we’ll learn more as we do more!

hypotext:
Thanks @wodeni. Can I remove commit 4ff0e43 from this branch? Looks like you moved it to `opt-webworker` already. And I'm getting a runtime error with webworkers on this branch on startup:

Screenshot 2020-05-05 14 15 15

wodeni:
> Thanks @wodeni. Can I remove commit [4ff0e43](https://github.com/penrose/penrose/commit/4ff0e43549a7db22dfd33d55218296fd54aea138) from this branch? Looks like you moved it to `opt-webworker` already. And I'm getting a runtime error with webworkers on this branch on startup:
> 
> Screenshot 2020-05-05 14 15 15

Yeah, this is unintended. Sorry about the confusion!

hypotext:
web port with our own gradient descent (not tf.js) w/ hand tuned timesteps:

`t=0.002`
web-gd

`t=0.1` (clearly t is too big)
web-tree

hypotext:
@maxkrieger I'd be happy to repro/review the inspector work after the README is ready!

hypotext:
Took a quick look at the inspector and tried it out. It works great so far. Thanks @maxkrieger!

maxkrieger:
Thanks for catching those missed cases with width/height computation. This might be a useful utility function to have more generally: computing the absolute bbox of any GPI.

hypotext:
@maxkrieger Sure thing. Actually Stella caught it when trying to run the continuous map example. I agree that in general, a bbox function for GPIs would be a useful utility function. I may well add it because it's useful for writing optimization functions as well.

hypotext:
@strout18 I added a temp SVG path here for you to use on completing the `mod` tab for shapes (black line below) in #362. Repro:

image

`runpenrose linear-algebra-domain/twoVectorsPerp.sub linear-algebra-domain/linear-algebra-paper-simple.sty linear-algebra-domain/linear-algebra.dsl`

(The path is hardcoded in `orientedSquare` for now; I plan to come back and finish this example later)

Things to finish:
- [ ] orientedSquare implementation with tensors
- [ ] unit mark implementation (commented out in the simple style, original full style to repro is `linear-algebra-paper.sty`)
- [ ] calculate curve bbox size for inspector
- [ ] implement label positioning objective function, `atDist`

---

BTW @wodeni where would be the right place to convert `pathData` information (in tensors) to normal numbers if autodiff is off? Is this the right place (in `resolvePath`, when a property value is looked up and returned)? https://github.com/penrose/penrose/pull/336/commits/369d5a5e5e2626230a4b4c8922d9447208b81411#diff-8b57d4ca1f6e63958534ea22760608a7R436

PR #338, opened 2020-04-15
Port optimizer and optimization problem generation to the web

hypotext:
**Not yet ready for review**

`optimizer.js` contains a version of the optimizer that is written to take gradients using tensorflow.js's `grads` function. I am slowly bringing it up to parity with the following parts of the Penrose system:

* Optimizer module
* `evalEnergyOn` function
* some parts of `GenOptProblem` 

I am working on:

* defining a clear interface between Style program types and TF types
* defining a clear interface between TF types and Penrose internal runtime types

---

# Description

Related issues: # (issue)

# Checklist

# Open questions


maxkrieger:
@hypotext try https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply

hypotext:
Hey @maxkrieger I think `apply` does the reverse of what we need. It seems to do the same thing as the spread syntax, which lets you call a tuple-function with a list as an argument. What I need to be able to convert a list-taking function into a tuple-function with the same number of args:

e.g. from `([x,y]) => x + y` to `(x, y) => x + y`

The problem was that `grads` in tf.js only accepts a tuple function: [see docs](https://js.tensorflow.org/api/latest/#grads). However, we don't know the number of arguments to an objective function statically in Penrose, since there's a user-defined number of varying variables. So in fact, I had removed a use of `apply` from the tf.js code for `grads` to make it work, since their autodiff implementation doesn't seem to rely on the function having a static number of arguments.

Am I missing something obvious? Would love for there to be a simpler solution.

hypotext:
However, I learned that the `Function` API seems like a pretty good solution to defining functions on the fly for Style: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function 

Only problem is it only has access to global scope. See example [here](https://github.com/penrose/penrose/pull/338/files#diff-a3157fc7530487c671dff1b6ebbf6444R90).

maxkrieger:
Ahhh sorry! I only read the commit message.

> `grads` in tf.js only accepts a tuple function

I don't think I follow. The example is a tuple function but:

`The provided f must take one or more tensors and return a single tensor y`

> removed a use of `apply` from the tf.js code for `grads` (on [this line](https://github.com/tensorflow/tfjs/blob/tfjs-core-v1.7.2/tfjs-core/src/gradients.ts#L123)) to make it work, since their autodiff implementation doesn't seem to rely on the function having a static number of arguments.

So whose end relies on tuple functions?

> Am I missing something obvious? Would love for there to be a simpler solution.

My mental API-occam-razor suggests yes since this sounds like a bug on either end. Does our implementation actually break their internal code? What errors were we getting?

wodeni:
@hypotext have you looked into [Rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters)? You can just write the overall objective as like `f(...args) => args.reduce(...)` and work with `args` as if it's a list. I tried it in your code (`demo.js`), see below:

```js
let centerFn3 = (x, y, z) => x.square().add(y.square()).add(z.square());
let centerFn3Rest = (...args) =>
  args.reduce((res, n) => res.add(n.square()), tf.scalar(0));
let gpi2_varying_vals = [num(10.0), num(25.0), num(3.0)];
const gradFn2 = tf.grads(centerFn3); // Use grads for multi-args
const gradFn2Rest = tf.grads(centerFn3Rest); // Use grads for multi-args
console.log("centerFn3'(gpi2) = ");
centerFn3(...gpi2_varying_vals).print();
console.log("centerFn3Rest'(gpi2) = ");
centerFn3Rest(...gpi2_varying_vals).print();
console.log("grads for gradFn2: ");
let [dx, dy] = gradFn2(gpi2_varying_vals);
// the x passed in grad(f)(x) must be a tensor
dx.print();
dy.print();
console.log("grads for gradFn2Rest: ");
let [dxr, dyr] = gradFn2Rest(gpi2_varying_vals);
// the x passed in grad(f)(x) must be a tensor
dxr.print();
dyr.print();
```
Console output:
```
centerFn3'(gpi2) = 
array_ops.ts:1067 Tensor
    734
demo.js:73 centerFn3Rest'(gpi2) = 
array_ops.ts:1067 Tensor
    734
demo.js:75 grads for gradFn2: 
array_ops.ts:1067 Tensor
    20
array_ops.ts:1067 Tensor
    50
demo.js:80 grads for gradFn2Rest: 
array_ops.ts:1067 Tensor
    20
array_ops.ts:1067 Tensor
    50
```

hypotext:
The rest params are exactly it, thanks @wodeni! I will revert the last two commits and use that instead.

Also, some possible next steps for the optimizer:

* Speed benchmarks
* Port LBFGS (see below) + line search
* Port constrained opt via exterior pt method
* Add different solver (from discussion with Keenan)

LMK if you have preferences @wodeni. Do you prefer I make changes here, or in `web-runtime`?

---

Putting this here for future reference (via @wodeni): 
* [LBFGS optimizer in tf.js files](https://github.com/tensorflow/tfjs-core/pull/1385/files)
* [PR](https://github.com/tensorflow/tfjs-core/pull/1385)
* [for style transfer](https://github.com/tensorflow/tfjs/issues/127)

Quoting:

```
Currently we're only focusing on SGD variants of optimizers, but if you are interested in making an L-BFGS, 2nd order gradient based optimizer you can either

extend the Optimizer base class and override computeGradients, and applyGradients
or
simply write your own optimizer without using the Optimizer class. You can compute gradients with the eager style method of differentiation (use methods from this section and then use dl.variable() and variable.assign() to do an update.
```

Some things to consider:

* What has changed between 2018 and now for the tf.js optimization API?
* Should we use the TF Optimizer class? What are the pros/cons (e.g. speed vs. having to satisfy the interface)?
* How do we write a _fast_ optimizer in tf.js? (Looks like the PR makes heavy use of `dispose`, etc.)
* The LBFGS in the PR looks decent on a quick skim, though of course haven't tested. Should we reuse that or write our own?
* The LBFGS in the PR also has nice tests. Should port/reuse those and test more thoroughly. (What are the standard ways to test it?)

wodeni:
Now that we have a functional optimizer in the frontend, I'd suggest that we keep all experimental/tfjs related things in this branch and Penrose-related things on `web-runtime`. So for the LBFGS experiments, it's probably better to do it on this branch until it's stable and I'll pull your changes. Does this make sense?

hypotext:
@wodeni Cool. So then I'll pull your changes from `web-runtime` to here and work in the Penrose optimizer framework you started (rather than in the separate "test" optimizer I started). Sound good?

wodeni:
Closing this PR since we no longer depend on TF.js.

PR #342, opened 2020-06-03
enable continuous map in typescript

strout18:
# Description

Ported additional constraints/objs from the Haskell into Constraints.ts to be able to generate the continuous map diagram. Ported "centerLabel", "below", and extended "contains", "minSize", "maxSize" to accomodate squares.

# Implementation strategy and design decisions

Did not modify any equations in the Haskell, just translation into TS.

# Examples
The following code now works:
`runpenrose set-theory-domain/continuousmap.sub set-theory-domain/continuousmap.sty set-theory-domain/setTheory.dsl`

image
image

PR #356, opened 2020-07-08
Fixing layering issues in web-runtime

strout18:
# Description

Issue #355 

Previously in web-runtime, shapes were not being sorted by ordering after each optimizer step, which caused layering keywords like `above` to behave unexpectedly when rendering diagrams. The fix ensures shape sorting after each step. I also updated the set theory example (`runpenrose set-theory-domain/tree.sub set-theory-domain/venn-opt-test.sty set-theory-domain/setTheory.dsl`) to explicitly define layering relationship between a Set shape and a Set text, which became necessary after the fix. I also included the original diagram mentioned in the relevant issue to show that the problem is fixed (`runpenrose vizdoc/near.sub vizdoc/near.sty vizdoc/shapes.dsl 
`).

Copied the function used in Canvas to originally sort the shapes to the Evaluator inside the `evalTranslation` function.

Screen Shot 2020-07-10 at 8 32 39 PM

`runpenrose set-theory-domain/tree.sub set-theory-domain/venn-opt-test.sty set-theory-domain/setTheory.dsl`

Screen Shot 2020-07-10 at 8 34 51 PM

`runpenrose set-theory-domain/tree.sub set-theory-domain/tree.sty set-theory-domain/setTheory.dsl`

Screen Shot 2020-07-10 at 8 47 39 PM

`runpenrose vizdoc/near.sub vizdoc/near.sty vizdoc/shapes.dsl 
`

# Open questions

In most diagrams, text is layered above the shape in a given object in Substance. It could be useful to implement this as the default setting, to prevent the user from having to write out the same layering in every Style file. Currently the default layering is unpredictable and depends on `Translation` order.

### DIFF
--- a/examples/set-theory-domain/venn-opt-test.sty
+++ b/examples/set-theory-domain/venn-opt-test.sty
@@ -10,6 +10,7 @@ forall Set x {
     ensure minSize(x.shape)
     ensure maxSize(x.shape)
     encourage sameCenter(x.text, x.shape)
+    x.textLayering = x.text above x.shape
 }
 
 forall Set x; Set y
--- a/examples/vizdoc/near.sty
+++ b/examples/vizdoc/near.sty
@@ -1,0 +1,33 @@
+Colors {
+    Colors.black = rgba(0.0, 0.0, 0.0, 1.0)
+    Colors.none = rgba(0.0, 0.0, 0.0, 0.0)
+    Colors.gray = rgba(0.6, 0.6, 0.6, 1.0)
+    Colors.red = rgba(1.0, 0.0, 0.0, 1.0)
+    Colors.lightblue = rgba(0.580, 0.910, 1.0, 1.0)
+    Colors.darkblue = rgba(0.078, 0.255, 0.788, 1.0)
+}
+
+Circ x {
+  x.shape = Circle {
+    strokeWidth : 2.0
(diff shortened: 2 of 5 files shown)


hypotext:
Hey! I was able to repro and it looks good overall. Two requests before I merge:

- [ ] FYI the path in the PR, `runpenrose set-theory-domain/continuousmap.sub set-theory-domain/continuousmap.sty set-theory-domain/setTheory.dsl` doesn't seem to work, fails with `Square doesn't have maxSize`. Can you remove this or fix it?
- [ ] Can you test with existing examples on web-runtime (the big set example with venn and tree styles, and the little set examples with points) to make sure there are no regressions, and post the results here?

In general I feel like the shapes only need to be sorted once (this will re-sort all the shapes on every evaluation step, and will be O(N^2) in the number of shapes, so may not scale well)—but this is a plenty good fix for now.

strout18:
Oops, the `runpenrose set-theory-domain/continuousmap.sub set-theory-domain/continuousmap.sty set-theory-domain/setTheory.dsl` line was a mis-paste (though that example should work once #342 is merged). I've edited the original PR to fix that example. The two set-theory examples work, and I have included those results in the original PR as well.

hypotext:
Looks great to me. Was able to repro those examples. Thanks for fixing this!

Issue #361, opened 2020-07-23
Canvas size shouldn't be hardcoded


strout18:
Canvas dimensions are currently set to 800 x 700 in `Canvas.tsx`. This is noticeable when downloading the SVG from the relevant button, since the downloaded SVG crops out everything outside of those dimensions.

wodeni:
Bumping this issue because it's long overdue.

Issue #21, opened 2017-11-04
Prodirect manipulation


hypotext:
GUI changes to the diagram (location, duplication, etc.) should be inferred/applied/stored in the Substance program, and vice versa. Depends on #17.

wodeni:
Nice research direction, but a too general for an issue. Closing for now.

PR #362, opened 2020-07-23
Extending Penrose viz-à-viz documentation

strout18:
# Description

This branch has multiple projects, which are each independent but share one common goal: increasing Penrose's visual documentation and enabling future documentation. Current code focuses mainly on shapes. 

## Mod tab on Inspector.
Added a tab to the inspector that lets you modify shapes declared in a Substance file. It works with any `sub/sty/dsl` files, but recommended as base examples are any of the sub files in `examples/vizdoc/baseshapes`. Select a shape on the left. You should then be able to adjust any of the attributes on the selected shape pre- or post-optimization. To add custom shapes or adjust input types just check out the “Adding shapes to mod” section inside of the Inspector README. The goal of this was to provide an in-Penrose shape explorer allowing the user to see in real-time how different attributes affect the shape. Could also see this being used to 'tweak' a final diagram post-optimization.

Screen Shot 2020-07-22 at 9 55 57 PM

`runpenrose vizdoc/baseshapes/circle.sub vizdoc/baseshapes/baseshapes.sty vizdoc/shapes.dsl`

## Docshapes
The `examples/vizdoc/docshapes` directory contains multiple `.sub .sty .dsl` triplets that can be used to create simple shape diagrams. The goal with these was to create aesthetically pleasing, informative diagrams accessible to the casual user that demonstrate relevant properties of the shapes. A couple examples and the paths to reproduce them are below.

circle

`runpenrose vizdoc/docshapes/circle.sub vizdoc/docshapes/circle.sty vizdoc/shapes.dsl`

line2

`runpenrose vizdoc/docshapes/line2.sub vizdoc/docshapes/line2.sty vizdoc/shapes.dsl
`
All example PNGs so far can be found in the Google Drive. Current shapes are `Square, Rectangle, Line, Ellipse, Arrow, Text, Circle,` and `Image.`

## Shape modifier - HTML style

Added Penrose-independent shape-explorers, similar to the Mod tab. They can be found in `examples/vizdoc/html`. There are shape-specific explorers for current shapes in Penrose, as well as `template.html` and `template.js` files that support custom shapes. The `template` files currently work only on shape attributes that have a one-to-one correspondence with an SVG attribute or a one-to-many (like how `Square.side` in Penrose maps to both the `width` and `height` properties of a corresponding SVG rectangle), but not a many-to-one or many-to-many. The template file could be extended further, but the Mod tab in the inspector covers most of its functionality so it's probably deprecated. The immediate advantage of the pure HTML/JS file is that it could be embedded in a documentation page, whereas the Mod tab is only accessible through Penrose. To run just open the HTML file in Chrome. If this were to be implemented in documentation, we might want to implement CSS styling. Currently very bare bones. The example below is `examples/vizdoc/html/rectangle.html`.

Screen Shot 2020-07-22 at 10 14 13 PM

## Heatmap

Implemented an energy heatmap generator for constraints/objectives at `examples/vizdoc/html/heatmap.html`. Currently, the expression for the constraint/objective is entered manually in the file, and a heatmap is rendered using d3. It also supports adjusting x and y coordinates of shapes that objectives are being applied to. Still in most basic stage. Would not advise using in documentation until further work is done. Example below uses the expression for the sameCenter objective.

heatmap

## Square hotfix

The Square shape did not support the strokeStyle property. I added that functionality. Can repro by `runpenrose vizdoc/baseshapes/square.sub vizdoc/baseshapes/baseshapes.sty vizdoc/shapes.dsl` and modifying strokeStyle in the mod view of the inspector. 

# Remaining issues

- Added an option in the Mod tab to support setting minima and maxima of ranges to Canvas boundaries. However, Canvas boundaries are currently hardcoded (see Issue #361). Once/if that changes, the `onCanvas` function in `LabeledInput.tsx` will need to be updated. For more information on this, see the "Adding shapes to mod” section in the inspector README.
- Mod tab does not support changing the `string` property of a `Text/Label` element. I think this is because the `string` property is set during the `collectLabels` method that occurs during optimization and so will not be affected by a change in Canvas state (which Mod changes in order to modify the shapes on the canvas). Perhaps this issue could be fixed in Mod, but the most straightforward fix seems to be to not gather labels during optimization...
- Mod tab has been tested on simple `Curve` instances with points only: `runpenrose linear-algebra-domain/twoVectorsPerp.sub linear-algebra-domain/linear-algebra-paper-simple.sty linear-algebra-domain/linear-algebra.dsl`
~~- Mod tab has not been tested on `Curve` instances. This should happen eventually, but most of the functions (in `Functions.hs` on `master`) that are used to generate points on a `Curve` have not yet been ported to `web-runtime` and I wasn't sure about manually defining points.~~


maxkrieger:
This is dope omg

wodeni:
The heatmap demo is super cool! I'm assuming this is done with respect to one shape and it's positional properties. Would be interesting to see if this can be generalized to other types of properties (for optimized colors, a heatmap on the color space??).

hypotext:
Yeah this is all great work! Stella and I have talked about extending the shape mod tab to incorporate information from the computational graph of the diagram. As I wrote in Slack:

> Also, at some point, may be good to think through which values may need to be recomputed or reoptimized to preserve the relationships in the style program. e.g. if `x.shape.z` was set to `a.shape.y + 2` in Style, then the user wants to change `x.shape.z`, what should happen?

hypotext:
Thanks @strout18, I was able to use the `mod` tab on some examples pre- and post-optimization and it looks really cool. I can do a general review Mon or Tues so we can get this into `web-runtime` and merge `web-runtime` ASAP. 

@maxkrieger or @wodeni, can you review how @strout18's `mod` tab integrates with the inspector, and in general, how well it works with the React canvas update cycle, and anything else relevant to `web-runtime`? (since it updates the shapes on drag) Before Weds would be great for the merge.

examples/vizdoc/html/heatmap.html
```diff
@@ -0,0 +1,193 @@
...
+    d3.select("#circ2y-value").text(val);
+    d3.select("#circ2y").property("value", val);
+    d3.select("#circ2").attr(
+        "cy", val
+    );
+    circ2[1] = val;
+  }
+  calcval();
+
+    // Note: If your code isn't working, try googling more specific search terms (I got here by searching "d3 update selection svg", since it seemed like the update selection for existing elements in the svg wasn't working)
```
hypotext, line 163: You can delete my old comments in this file @strout18

react-renderer/src/inspector/views/mod/shapedefs/arrow.json
```diff
@@ -0,0 +1,55 @@
+{
```
wodeni, line 1: Great to finally have shapedefs in the frontend! We will eventually move the backend shapedefs into the frontend as well. A more principled way to generate these mod widgets: each shapedef will also include the datatype for each property. We can then derive the `inputType` based on it (e.g. `FloatT` -> `range`, `StringT` -> `select'). The new shapedefs should also define the domain of their values (e.g. `arrowheadStyle` -> `["arrowhead-1", "arrowhead-2"]`). WIth this information, we can generate sample functions for resampling shapes as well.

wodeni (approved):
Looks great overall. Thanks for making this possible @strout18 ! Excited to integrate things like `mod` into the diagramming workflow eventually (#21 !). Most of my comments are for our future reference, not requests for changes.

examples/vizdoc/html/heatmap.html
```diff
@@ -0,0 +1,193 @@
...
+// Initialize each cell with energy of obj function
+function calcval(){
+    for (var j = 0, k = 0; j < m; ++j) {
+    for (var i = 0; i < n; ++i, ++k) {
+        values[k] = constraint(i, j);
+    }
+    }
+}
+
+// Sum the distances between a cell and the two circles - is this the right function
```
hypotext, line 54: So this is almost the right function, but not quite :) `sameCenter` is a function with four DOF: x1, y1, x2, y2, defined as ||c2-c1||^2 (where c1 = [x1, y2], etc). 

When you draw the energy function, it's only possible to plot two DOF in 2D, just an x and y.

So, really you should fix one of the points `c1`, and calculate sameCenter over all cells, using the distance from the moveable point `c2`. So the energy at cell `p := (x,y)` is `distsq(p, c1)`. The result is basically a "bowl" at `c1`, with no moveable point `c2`, because the energy function is taken over all points `c2`. But `c1` can move, and when it does, the bowl should move with it. Does that make sense? I would change the math here (and the sliders and README explanation) to reflect this. (I also recommend doing math in terms of vector operations, like norm, rather than coordinate-wise operations. It's just way more readable.)

Now, whether that is a good way to visualize the function is another question. Probably a better way is to show the gradient of sameCenter as two arrows `[df/dx1, df/fy1]` and `[df/dx2, df/dy2]`, which shows "everything" rather than the limited slice from the heatmap. But we can leave that for another time.

react-renderer/src/inspector/README.md
```diff
@@ -25,8 +25,37 @@ A useful library component is `import { ObjectInspector } from "react-inspector"
 
 See `ShapeView`, `Timeline`, and `Frames` for useful examples.
 
+## Adding shape support to Mod view
```
hypotext, line 28: Thanks for this thorough documentation!

react-renderer/src/inspector/views/Mod.tsx
```diff
@@ -0,0 +1,161 @@
...
+              listStyleType: "none",
+              padding: "0 0 1em 0",
+              margin: 0,
+              top: 0,
+              width: 156,
+              left: 0,
+              right: 0,
+            }}
+          >
+            {frame.shapes.map(
```
hypotext, line 74: This looks like a copy-paste of code from `ShapeView` to calculate the bbox of a shape. Can you factor it out into `Util.tsx` and use it in both files?

react-renderer/src/inspector/views/mod/LabeledInput.tsx
```diff
@@ -0,0 +1,258 @@
+import * as React from "react";
+import {toHex} from "../../../Util";
+import * as _ from "lodash";
+
+interface IProps {
+    inputProps: IInputProps,
+    eAttr: string;
+    eValue: Value<any>; // is this the best typing?
```
hypotext, line 8: We probably want `Value<VarAD>` when merged

react-renderer/src/inspector/views/mod/AttrPicker.tsx
```diff
@@ -0,0 +1,30 @@
...
+    sAttrs: IShapeDef;
+    modAttr(attrname: string, attrval: Value<any>): void;
+}
+
+interface IShapeDef {
+    shapeType: string;
+    properties: object;
+}
+
+// Q - should this update shape properties in state? not really necessary functionally but maybe ideologically
```
hypotext, line 15: Does this not update shape properties in the state? I expected it to do so, but if not can you say more about how what it is updating?

react-renderer/src/inspector/views/mod/LabeledInput.tsx
```diff
@@ -0,0 +1,258 @@
...
+        const {inputProps, eAttr} = this.props;
+        return (<input id={eAttr} type= "range" onChange={this.handleChange} min= {toCanvas(inputProps.min!)} max={toCanvas(inputProps.max!)
+        } value={_.round(this.props.eValue.contents)} />)
+    }
+    // for use in makePointRange or potentially other multi-input attributes (would need to be modified)
+    public makeSublabel = (id: string) => {
+        const itemprops = id.split("_");
+        const subindex = itemprops[4];
+        const ptindex = itemprops[3] === "x" ? 0 : 1;
+        console.log(this.state.eValue.contents[+itemprops[1]].contents[+subindex].contents[ptindex]);
```
hypotext, line 132: Remove extraneous `console.log`s?

hypotext (changes requested):
LGTM in general, thanks for the awesome work @strout18. If you can address the minor changes I pointed out in comments (forgot to put them in the GitHub review feature...) then we can move along with the merge (pending @maxkrieger's thoughts on the React integration). Thanks!

react-renderer/src/App.tsx
```diff
@@ -51,12 +51,27 @@ class App extends React.Component<any, ICanvasState> {
...
     this.setState({ penroseVersion: version });
   };
+  // same as onCanvasState but doesn't alter timeline or involve optimization
+  // used only in modshapes
+  public modCanvas = async (canvasState: State) => {
+    await new Promise((r) => setTimeout(r, 1));
+
+    await this.setState({
+      data: canvasState,
+      history: [...this.state.history],
```
maxkrieger, line 71: you can actually omit `history: [...this.state.history],` altogether -- I recently learned that React merges state updates https://reactjs.org/docs/state-and-lifecycle.html#state-updates-are-merged

react-renderer/src/inspector/views/IViewProps.tsx
```diff
@@ -7,5 +7,6 @@ export interface IViewProps {
   frame: State | null;
   // Either a valid index in History
   frameIndex: number;
+  modShapes?(state: IState): void; // todo - null check
```
maxkrieger, line 10: why is it nullable?

react-renderer/src/inspector/views/Mod.tsx
```diff
@@ -0,0 +1,161 @@
...
+  public setSelectedShape = (key: number) => {
+    this.setState({ selectedShape: key});
+  };
+  public modAttr = (attrname: string, attrval: Value<any>) => { 
+    const {frame, modShapes, frameIndex, history} = this.props;
+    const {selectedShape} = this.state;
+    if (frameIndex === -1 || frameIndex === history.length - 1) { // make it only work on last frame
+      const shap = frame!.shapes[(selectedShape === -1 ? 0 : selectedShape)]; // the -1 index behaves weirdly so workaround
+      if (shap.properties.hasOwnProperty(attrname)) {
+        shap.properties[attrname] = attrval;
```
maxkrieger, line 38: This seems prone to bugs because you're not supposed to mutate properties in-place if they're from props. Props should always be immutable https://reactjs.org/docs/optimizing-performance.html#the-power-of-not-mutating-data.

It shouuuuld be as simple a replacement as:
```ts
const shape = frame!.shapes[(selectedShape === -1 ? 0 : selectedShape)];
if (attrname in shape) {
  const newFrame = {...frame,
      shapes: {...frame.shapes,
         [(selectedShape === -1 ? 0 : selectedShape)]:
            {...shape, properties: {...shape.properties, [attrname]: attrval } }
  }; 
}
```

react-renderer/src/inspector/views/Mod.tsx
```diff
@@ -0,0 +1,161 @@
...
+class Mod extends React.Component<IViewProps, IState> {
+  public readonly state = { selectedShape: -1};
+  public setSelectedShape = (key: number) => {
+    this.setState({ selectedShape: key});
+  };
+  public modAttr = (attrname: string, attrval: Value<any>) => { 
+    const {frame, modShapes, frameIndex, history} = this.props;
+    const {selectedShape} = this.state;
+    if (frameIndex === -1 || frameIndex === history.length - 1) { // make it only work on last frame
+      const shap = frame!.shapes[(selectedShape === -1 ? 0 : selectedShape)]; // the -1 index behaves weirdly so workaround
```
maxkrieger, line 36: would be interested in why:

* `-1` index behaves weirdly

* `frame!` is possibly null: just check if it's null at the top level of the function?

react-renderer/src/inspector/views/Mod.tsx
```diff
@@ -0,0 +1,161 @@
...
+            // BUG: scroll doesnt really work
+            padding: "1em 1em 1em 1em",
+            overflow: "auto",
+            height: "100%",
+            flexBasis: 0,
+            flexGrow: 1,
+            boxSizing: "border-box"
+          }}
+        >
+          {/* todo enable only on last frame */}
```
maxkrieger, line 153: hm?

react-renderer/src/inspector/views/mod/AttrPicker.tsx
maxkrieger: ^Possibly related: my comments in `Mod.tsx` (github isn't letting me link them directly)

react-renderer/src/inspector/views/mod/LabeledInput.tsx
```diff
@@ -0,0 +1,258 @@
...
+    public updateAttr = (id: string, evalue: string | number | Color<number> | boolean | SubPath<number>[]) => {
+        // this.setState((prevState) => ({
+        //     eValue: {
+        //         ...prevState.eValue,
+        //         contents: evalue
+        //     }
+        // }))
+        
+        // this way is not safe apparently. https://stackoverflow.com/questions/43638938/updating-an-object-with-setstate-in-react
+        // however, the alternative is above and gives an error!!!
```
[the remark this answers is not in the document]
maxkrieger: The un-commented method is totally safe and the SO answers are terribly misleading. The commented method isn't typical and is generally phased-out as a style.

react-renderer/src/inspector/views/mod/LabeledInput.tsx
```diff
@@ -0,0 +1,258 @@
...
+        }
+        this.setState(newstate);  // will update span values - could be phased out if spans are set manually
+        this.props.modAttr(id, newstate.eValue as Value<any>);
+    }
+    public handleChange = (event: React.ChangeEvent<HTMLInputElement|HTMLSelectElement>) => {
+        const teval = (isNaN(Number(event.target.value))) ? event.target.value : +event.target.value; // convert to number if necessary
+        this.updateAttr(event.target.id, teval); // a little hacky - id is name of property
+    }
+    // when an input corresponding to a part of an attribute is modified instead of an entire attribute
+    public handleSubChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
maxkrieger, line 83: I have no idea what this does but thats fine I guess

react-renderer/src/inspector/views/mod/LabeledInput.tsx
```diff
@@ -0,0 +1,258 @@
...
+        const ptinfo = event.target.id.split("_");
```
maxkrieger, line 84: Don't encode data into a string and parse it afterwards! There's always, ALWAYS a better way!

react-renderer/src/inspector/views/mod/LabeledInput.tsx
```diff
@@ -0,0 +1,258 @@
...
+                return (
+                <React.Fragment key={"S" + index}>
+                {ptarray.map((pt: Elem<number>, subindex: number)=> {
+                    if (pt.tag !== "Pt") throw new Error('No current support for Bezier curves!');
+                    const xid = ["S", index.toString(), eAttr, "x", subindex.toString()].join("_");
+                    const yid = ["S", index.toString(), eAttr, "y", subindex.toString()].join("_");
+                    return (
+                        <React.Fragment key={"S" + index + "pt" + subindex}>
+                        <div style={{display: "inline-block"}} className="sublabinput">
+                            <input id={xid} type="range" onChange={this.handleSubChange} min={toCanvas(inputProps.minX!)}
```
maxkrieger, line 158: Using `id`s anywhere in React is unsupported and, in this case, indicative of a bigger misconception: if you want to reuse an event handler, there are lots of ways to pass that data into it!

One way: wrap the `onChange` in a lambda, and you can pass all the necessary data upstream like so: `onChange={(e) => this.handleSubChange(index, eAttr, subindex, e, ...)`. Or curry the `handleSubChange` for slightly more readability (effectively the same thing).

maxkrieger (changes requested):
Overall looks good, but there are some issues for `mod` that we'll have to refactor eventually because of readability/immutability/maintainability issues (see inline comments incl those not attached below).

react-renderer/src/inspector/views/Mod.tsx
```diff
@@ -0,0 +1,161 @@
...
+              width: 156,
+              left: 0,
+              right: 0,
+            }}
+          >
+            {frame.shapes.map(
+              ({ properties, shapeType }: Shape, key: number) => {
+		  // If the inspector is crashing around here, then probably the shape doesn't have the width/height properties, so add a special case as below
+		  // console.log("properties, shapeType", properties, shapeType, properties.w, properties.h);
+
```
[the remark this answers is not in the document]
strout18: That code is taken directly from `Shapeview`, so maybe to change this we should involve @maxkrieger

react-renderer/src/inspector/views/mod/shapedefs/arrow.json
strout18: @wodeni I did consider that, but I opted for extra customization, in case someone thinks a `number` input is more appropriate than `range` for their custom property of type `FloatT`. Also, I think the same types can map to multiple input options; e.g, the option to edit the content of a `text` should be presented as a text field, but `arrowheadStyle` should be a select, though both have values of type `String`. But I think once the backend shapedefs are moved to the front end it could be possible to do what you're suggesting.

react-renderer/src/inspector/views/IViewProps.tsx
strout18: Not sure... I'll change it.

react-renderer/src/inspector/views/Mod.tsx
strout18: @maxkrieger `frame.shapes` is an array, so that code won't exactly work. I tried replacing it with the following code:

```
      const shapeIndex = selectedShape === -1 ? 0 : selectedShape;
      const newShapes = _.cloneDeep(frame!.shapes);
      let shape = newShapes[shapeIndex];
      shape.properties[attrname] = attrval;
      const newFrame = {
        ...frame,
        shapes: newShapes
      } as IState;
      modShapes(newFrame);
```
but it didn't want to let me cast to IState.

react-renderer/src/inspector/views/Mod.tsx
strout18: Expired comment... I'll get it rid of it.

react-renderer/src/inspector/views/mod/LabeledInput.tsx
strout18: Currently it modifies inputs that map to only part of a shape attribute, e.g. how both the `opacity` and `color` input map to the `color` attribute. I plan on factoring it out.

react-renderer/src/inspector/views/IViewProps.tsx
strout18: Addressed in commit 2b9315a4a27f35fb751ac317f15cdbf9a18ec5e4

react-renderer/src/inspector/views/Mod.tsx
strout18: Addressed in 2b9315a4a27f35fb751ac317f15cdbf9a18ec5e4

react-renderer/src/inspector/views/mod/LabeledInput.tsx
strout18: Addressed in 2b9315a4a27f35fb751ac317f15cdbf9a18ec5e4

react-renderer/src/inspector/views/mod/LabeledInput.tsx
strout18: Addressed in 2b9315a4a27f35fb751ac317f15cdbf9a18ec5e4

examples/vizdoc/html/heatmap.html
strout18: Addressed in 2b9315a4a27f35fb751ac317f15cdbf9a18ec5e4

react-renderer/src/inspector/views/mod/shapedefs/arrow.json
wodeni: Good point. My current thought is to annotate shape properties with types and a range of values, and then make the decision on how to present them. Maybe an extension point within vizdoc to override defaults will give users enough flexibility.

react-renderer/src/inspector/views/Mod.tsx
wodeni: > but the compiler didn't want to let me cast to IState.

Maybe check the type of `attrval` and see if that conform to `IState`'s types?

react-renderer/src/inspector/views/mod/AttrPicker.tsx
strout18: @hypotext It doesn't change the state of that component specifically, just passes up the changes to `App.tsx` where that state gets modified (which then triggers a re-render, so it does change the state indirectly...)

strout18:
@hypotext This is ready for re-review! Also dealt with the merge conflicts.

react-renderer/src/inspector/views/Mod.tsx
strout18: Fixed in 0e5d3a803d6d199268633fdf5d6c1d38a92acaea

react-renderer/src/inspector/views/Mod.tsx
strout18: Fixed in 0e5d3a803d6d199268633fdf5d6c1d38a92acaea

examples/vizdoc/html/heatmap.html
strout18: Tried fixing it in 0e5d3a803d6d199268633fdf5d6c1d38a92acaea @hypotext , still might not be right! Also, how would you recommend doing math with vectors? I was hesitant to use TFJS since it seemed like it was being factored out...

hypotext: Yes, this looks right to me. Don't worry about the vector math (especially since you're making the visualization in plain JS), though for future reference I'd just define some plain JS helper functions on lists, or use lodash.

hypotext (approved):
LGTM, the heatmap looks fine now, though on trying to re-run your branch (pulling and building) I got this error:

```
Failed to compile.

./src/inspector/views/mod/LabeledInput.tsx
Module not found: Can't resolve '../../../Util' in '~/Desktop/Code/penrose2/react-renderer/src/inspector/views/mod'
```

Can you fix this first?

In GitHub web it says there are several merge conflicts, but on CLI I can merge with no problem (which I just did).

hypotext:
@maxkrieger Can you give Stella's changes a quick re-review? If all looks ok (after Stella fixes the Utils import problem above), I'll merge into `master`.

strout18:
The import problem should be fixed in 6153a1b1

penrose-web/src/inspector/views/mod/LabeledInput.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from "react";
-import {toHex} from "../../../Util";
+import {toHex} from "../../../utils/Util"
```
wodeni, line 2: BTW I changed the module resolution rules and you can just use absolute paths! `utils/Util` would work just fine.

maxkrieger (approved):
LGTM, thanks for fixing this stuff!

react-renderer/src/inspector/views/mod/LabeledInput.tsx
maxkrieger: I think both, but it's fine for now

hypotext:
Thanks @strout18 for building this! And @wodeni @maxkrieger for reviewing.

PR #370, opened 2020-08-26
Change "engine" to run on autodiff types only

hypotext:
# Description

Previously, the "engine" had the option of using `number` or `Tensor` types in the evaluator via the `autodiff` flag, because we needed the former type for displaying shapes (evaluating the translation) and the latter type for optimization (evaluating the energy).

While concise, the design started to cause some challenges:

* Computations appear in both kinds of evaluation, thus had to be written with both numeric types
* Lots of code duplication in the evaluator, since it needs to interpret expressions on both numeric types and deal with mismatches
* Hard to debug if a single number of the wrong type gets into the wrong evaluation phase (e.g. `number`s in the optimization)

Therefore, this PR changes all evaluation + optimization-related functions and types to be on autodiff types only (in this case, Tensors). 

The tradeoff is that the display step may take slightly longer and use more space, since it requires doing all work on autodiff types (which are objects that store a good amount of information), then converting them to numbers. However, that tradeoff is acceptable because the engine does much more optimization-type evaluation than display-type evaluation.

The new pipeline is as follows:

- decode state from backend `JSON`, which gives us numbers 
  - note that the deserialization is the _least type-safe_ part of our code [0]
- **in the initial state, convert all `Value<number>` into `Value<Tensor>` to store in the `Translation`, which now only holds `Tensors`**
  - the `varyingValues` are also converted to be all `Tensor`s, as are the pending values (label dimensions) after they are calculated
  - expressions (`OptEval`) that can be evaluated at optimization-time are **not** converted; if a `number` is encountered in evaluation, it's converted on the spot in `evalExpr`
- *all* evaluation happens on `Tensors`
  - in the optimization, it just uses the `Tensors` stored in the translation
  - for displaying shapes via `evalTranslation`, it also does all computations on `Tensors`, and **does a final conversion pass from `Value<Tensor>` to `Value<number>`**

In general, numeric values are converted to tf `Scalars` via the `scalar` function (so they just tensors, *not* variables in the optimization), unless they are varying values, in which case they are converted to tf `Variable`s.

Most changes are in `Evaluator` and `types.d.ts`, with some changes in `PropagateUpdate` and `Canvas`, and new files `EngineUtils` and `Computations`.

Also, the diff is a little misleading because I changed my editor's indentation, but I kept it because I will probably use this indentation in the future.

[0] If something is deserialized with the wrong type (e.g. the deserializer doesn't know that something is supposed to be a `Tensor`, not a `number`), our types are gone at runtime, so there are no guarantees and the wrong type just gets everywhere.

* Test inline computation with `Tensor`s: `runpenrose hyperbolic-domain/hyperbolic-example.sub hyperbolic-domain/PoincareDisk.sty hyperbolic-domain/hyperbolic.dsl`

The following three are successfully-repro'ed examples from the [wiki page](https://github.com/penrose/penrose/wiki/Contributing-to-the-web-runtime) for `web-runtime` working examples.

* Test labels, pending values, varying values, and optimization: `runpenrose set-theory-domain/tree.sub set-theory-domain/venn-opt-test.sty set-theory-domain/setTheory.dsl`

* `runpenrose set-theory-domain/tree.sub set-theory-domain/tree.sty set-theory-domain/setTheory.dsl`

* `runpenrose set-theory-domain/continuousmap.sub set-theory-domain/continuousmap.sty set-theory-domain/setTheory.dsl`

# Checklist

# Open questions

- [ ] In the future, the system should use a custom autodiff type that is easily swappable (i.e. not `Tensor`s). Using `Tensor`s is a stopgap until `web-perf` and @strout18's macro code get merged.
- [ ] I wasn't sure if tf `Variable`s needed to have their gradients cleared after each step, so for safety, I did that in `Optimizer` at the end of each `stepEP`. In general, when we use custom AD vars, we may need to clear them.
- [ ] I wasn't super careful about the distinction between tf `Scalar`s and `Variable`. Hopefully this generates the right computational graph when we use custom AD vars.
- [ ] Maybe just convert Shapes from autofloats to numbers, instead of converting the whole translation? (@wodeni's suggestion)

### DIFF
--- a/react-renderer/src/Circle.tsx
+++ b/react-renderer/src/Circle.tsx
@@ -14,6 +14,14 @@ class Circle extends React.Component<IGPIProps> {
     const strokeAlpha = shape.strokeColor.contents.contents[3];
     const thickness = shape.strokeWidth.contents;
 
+    let dashArray = "7,5";
+    if( "strokeDashArray" in shape )
+    {
+       if( shape.strokeDashArray ) {
+          dashArray = shape.strokeDashArray.contents;
+       }
+    }
+
     return (
       <circle
         cx={x}
@@ -23,7 +31,7 @@ class Circle extends React.Component<IGPIProps> {
         fillOpacity={fillAlpha}
         stroke={strokeColor}
         strokeOpacity={strokeAlpha}
-        strokeDasharray={shape.strokeStyle.contents === "dashed" ? "7, 5" : ""}
+        strokeDasharray={shape.strokeStyle.contents === "dashed" ? dashArray : ""}
         strokeWidth={thickness}
       >
         <title>{shape.name.contents}</title>
--- a/react-renderer/src/Circle.tsx
+++ b/react-renderer/src/Circle.tsx
@@ -15,11 +15,10 @@ class Circle extends React.Component<IGPIProps> {
     const thickness = shape.strokeWidth.contents;
 
     let dashArray = "7,5";
-    if( "strokeDashArray" in shape )
-    {
-       if( shape.strokeDashArray ) {
-          dashArray = shape.strokeDashArray.contents;
-       }
+    if ("strokeDashArray" in shape) {
+      if (shape.strokeDashArray) {
+        dashArray = shape.strokeDashArray.contents;
+      }
     }
 
     return (
--- a/react-renderer/src/Constraints.ts
+++ b/react-renderer/src/Constraints.ts
@@ -13,14 +13,13 @@ export const objDict = {
     return distsq(center(s1), center(s2));
   },
 
-
-  below: ([t1, bottom]: [string, any], [t2, top]: [string, any], offset = 100) => 
+  below: ([t1, bottom]: [string, any], [t2, top]: [string, any], offset = 100) =>
     square(top.y.contents.sub(bottom.y.contents).sub(scalar(offset))),
-    // can this be made more efficient (code-wise) by calling "above" and swapping arguments? - stella
+  // can this be made more efficient (code-wise) by calling "above" and swapping arguments? - stella
 
 
   centerLabel: ([t1, arr]: [string, any], [t2, text1]: [string, any], w: number): Tensor => {
-    if (typesAre([t1,t2], ["Arrow", "Text"])) {
+    if (typesAre([t1, t2], ["Arrow", "Text"])) {
       const mx = arr.startX.contents.add(arr.endX.contents).div(scalar(2.0));
       const my = arr.startY.contents.add(arr.endY.contents).div(scalar(2.0));
       // entire equation is (mx - lx) ^ 2 + (my + 1.1 * text.h - ly) ^ 2 from Functions.hs - split it into two halves below for readability
@@ -96,7 +95,7 @@ export const constrDict = {
       const d = dist(center(s1), center(s2));
       const textR = maximum(s2.w.contents, s2.h.contents);
       return d.sub(s1.r.contents).add(textR);
-    } else if (t1 === "Square" && t2 === "Circle"){
+    } else if (t1 === "Square" && t2 === "Circle") {
       // dist (outerx, outery) (innerx, innery) - (0.5 * outer.side - inner.radius)
       const sq = stack([s1.x.contents, s1.y.contents]);
       const d = dist(sq, center(s2));
@@ -187,8 +186,9 @@ export const looseIntersect = (center1: Tensor, r1: Tensor, center2: Tensor, r2:
   dist(center1, center2).sub(r1.add(r2).sub(scalar(padding)));
 // dist (x1, y1) (x2, y2) - (s1 + s2 - 10)
 
+// HACK: need to annotate the types of x and y to be Tensor
 export const center = (props: any): Tensor =>
-  stack([props.x.contents, props.y.contents]); // HACK: need to annotate the types of x and y to be Tensor
+  stack([props.x.contents, props.y.contents]);
 
 export const dist = (p1: Tensor, p2: Tensor): Tensor => p1.sub(p2).norm();
 
@@ -198,7 +198,6 @@ export const distsq = (p1: Tensor, p2: Tensor): Tensor => {
   const dp = p1.sub(p2);
   return dp.dot(dp);
 }
-
 
 // with epsilon to avoid NaNs
 export const normalize = (v: Tensor): Tensor => v.div(v.norm().add(epsd));
--- a/react-renderer/src/EngineUtils.ts
+++ b/react-renderer/src/EngineUtils.ts
@@ -1,0 +1,284 @@
+// Utils that are unrelated to the engine, but autodiff/opt/etc only
+
+import { sc } from "./Optimizer";
+import { Tensor, scalar } from "@tensorflow/tfjs";
+import { mapValues } from 'lodash';
+
+// TODO: Is there a way to write these mapping/conversion functions with less boilerplate?
+
+// Generic utils for mapping over values
+
+function mapTup2<T, S>(
+  f: (arg: T) => S,
+  t: [T, T]
+): [S, S] {
+  // TODO: Polygon seems to be getting through with null to the frontend with previously working set examples -- not sure why?
+  // TODO: Should do null checks more systematically across the translation
+  if (t[0] === null || t[1] === null) {
+    console.error("null elements in polygon");
+    return [scalar(0.0), scalar(0.0)] as unknown as [S, S];
+  }
+
+  return [f(t[0]), f(t[1])];
+};
+
+function mapTup4<T, S>(
+  f: (arg: T) => S,
+  t: [T, T, T, T]
+): [S, S, S, S] {
+  return [f(t[0]), f(t[1]), f(t[2]), f(t[3])];
+};
+
+function mapTup2LList<T, S>(
+  f: (arg: T) => S,
+  xss: [T, T][][]
+): [S, S][][] {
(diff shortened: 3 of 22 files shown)


hypotext:
Hey @wodeni, requesting a review as you wrote the original code. It works on the examples so I'm pretty confident, but can you give it a once-over by Monday so I can merge into `web-runtime`? Let me know! (See also my note in the README about editor indentation re: this large diff.)

wodeni:
> See also my note in the README about editor indentation re: this large diff.

I'll go over the changes soon. Quick q about indentation: did you just change from 2 to 4 spaces? If so, I'd love to hear the rationale. Where's this README you were referring to?

hypotext:
Yeah, 2 to 4 spaces, because it's what makes my editor auto-indent work correctly on both my machines... (not very principled!). I'm just referring to the PR README [here](https://github.com/penrose/penrose/pull/370#issue-474304001).

react-renderer/src/Circle.tsx
```diff
@@ -14,6 +14,14 @@ class Circle extends React.Component<IGPIProps> {
     const strokeAlpha = shape.strokeColor.contents.contents[3];
     const thickness = shape.strokeWidth.contents;
 
+    let dashArray = "7,5";
+    if( "strokeDashArray" in shape )
```
wodeni, line 18: Is this the new formatting? Not super standard in JS/TS projects. @maxkrieger any opinions?

react-renderer/src/Constraints.ts
```diff
@@ -3,173 +3,173 @@ import { canvasSize } from "./Canvas";
...
+        } else throw new Error(`${[t1, t2]} not supported for centerLabel`)
+    },
+
+    // Generic repel function for two GPIs with centers
+    repel: ([t1, s1]: [string, any], [t2, s2]: [string, any]) => {
+        // HACK: `repel` typically needs to have a weight multiplied since its magnitude is small
+        // TODO: find this out programmatically
+        const repelWeight = 10e6;
+        // 1 / (d^2(cx, cy) + eps)
+        console.log("repel", t1, s1, "shape2", t2, s2);
```
wodeni, line 38: Remember to comment this out. Loosely related: maybe we should start using loggers.

react-renderer/src/Circle.tsx
maxkrieger: I just use whatevers default for prettier + ts. I agree this formatting is atypical

hypotext:
@wodeni Sure, I'll try to get the 2-space indent to work. In the meantime, can you let me know if you were planning to review this PR further? Or did you want me to fix the indent first?

react-renderer/src/Evaluator.ts
```diff
@@ -208,23 +106,32 @@ const checkComp = (fn: string, args: ArgVal<number>[]) => {
...
-      ? (evalExpr(prop.contents, trans, varyingVars, false) as IVal<number>).contents
-      : prop.contents
-  );
+    const [shapeType, propExprs] = shapeExpr.contents;
+
+    // Make sure all props are evaluated to values instead of shapes
+    const props = mapValues(propExprs, (prop: TagExpr<Tensor>): Value<number> => {
+        if (prop.tag === "OptEval") {
+            // For display, evaluate expressions with autodiff types (incl. varying vars as AD types), then convert to numbers
+            // (The tradeoff for using autodiff types is that evaluating the display step will be a little slower, but then we won't have to write two versions of all computations)
```
wodeni, line 121: Will need to see profiling results to see the overhead of the `Tensor` => `number` step, otherwise okay with me.

react-renderer/src/Optimizer.ts
```diff
@@ -213,72 +215,75 @@ export const stepEP = (state: State, steps: number, evaluate = true) => {
...
-  return evalTranslation(newState);
+    const f = evalEnergyOn(state);
+    const fgrad = gradF(f);
+    const xs = state.varyingValues;
+    // const xs = state.varyingState; // NOTE: use cached varying values
+    // NOTE: minimize will mutates xs
+    const { energy } = minimize(f, fgrad, xs, steps);
+    // insert the resulting variables back into the translation for rendering
+    // NOTE: this is a synchronous operation on all varying values; may block
+    // Note: after we finish one evaluation, we "destroy" the AD variables (by making them into scalars) and remake them
```
wodeni, line 261: What's the consequence of not "destroying" the AD variables?

hypotext: If you reuse an AD variable later with our dynamic evaluator approach, then my guess is that Tensorflow records those additional operations in its computational graph, which might affect the gradient. But it's a moot point after we convert to using our custom AD + compiled gradients, because the AD vars become irrelevant after the gradient is compiled.

react-renderer/src/PropagateUpdate.ts
```diff
@@ -1,24 +1,27 @@
+import { Tensor } from "@tensorflow/tfjs";
 import { insertExpr } from "./Evaluator";
+import { valueNumberToAutodiff } from "./EngineUtils";
 
 /**
  * Find the value of a property in a list of fully evaluated shapes.
  * @param shapes a list of shapes
  * @param path a path to a property value in one of the shapes
  */
+// the `any` is to accomodate `collectLabels` storing updated property values in a new property that's not in the type system
```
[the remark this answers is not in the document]
hypotext: Yes...

react-renderer/src/EngineUtils.ts
```diff
@@ -0,0 +1,284 @@
+// Utils that are unrelated to the engine, but autodiff/opt/etc only
+
+import { sc } from "./Optimizer";
+import { Tensor, scalar } from "@tensorflow/tfjs";
+import { mapValues } from 'lodash';
+
+// TODO: Is there a way to write these mapping/conversion functions with less boilerplate?
```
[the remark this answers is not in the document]
hypotext: Yeah, true... I liked that the Shape type enforced that it only contains `Value<number>`, so there is no chance an autofloat makes it to the frontend. But it's probably faster to render if we only convert shapes. Maybe I'll make this change in the future?

react-renderer/src/Util.tsx
```diff
@@ -275,3 +275,9 @@ export const roundTo = (n: number, digits: number) => {
 
   return n3;
 };
+
+// https://stackoverflow.com/questions/31084619/map-a-javascript-es6-map
+// Basically Haskell's mapByValue (?)
+export function mapMap(map: Map<any, any>, fn: any) {
+  return new Map(Array.from(map, ([key, value]) => [key, fn(value, key, map)]));
```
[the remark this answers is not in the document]
hypotext: Got it. `mapMap` isn't currently used anywhere, so I'll just leave it in as a util.

react-renderer/src/Evaluator.ts
hypotext: Here's a quick shot of three steps for the tree example (using tf.js on this branch). You can see that optimization takes most of the time and the conversion time seems negligible. Maybe this will change once we have compiled gradients. 

image

react-renderer/src/Circle.tsx
hypotext: Oh, this was Keenan's code. I just fixed the formatting.

react-renderer/src/Constraints.ts
hypotext: Fixed

react-renderer/src/EngineUtils.ts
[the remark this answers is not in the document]
wodeni: Yeah the timing is up to you. If it's too involved, let's make sure we come back to it. Seems like this is even more code duplication than before :(

react-renderer/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "outDir": "build/dist",
     "module": "esnext",
     "target": "es5",
-    "lib": ["es5", "es6", "es7", "dom"],
+      "lib": ["es5", "es6", "es7", "esnext", "dom"],
```
hypotext, line 7: FYI I added esnext here @wodeni @maxkrieger. Wanted to use `Object.fromEntries` in `EngineUtils`. Lmk if unnecessary

react-renderer/src/EngineUtils.ts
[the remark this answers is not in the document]
hypotext: This code is pretty generic, so can be used for mapping over any translation stuff, not specific to autofloat/floats. However I am down to use the shape conversion approach instead, later.

(The code duplication problem earlier involved computations + operations in evaluations needing to be defined on both autofloats and floats, e.g. `add` had to use both `tf.add` and the js `+` depending on the inputs, and it was difficult for me to debug why autofloats/floats ended up where they weren't supposed to be.)

hypotext:
FYI @wodeni I just made changes WRT the above suggestions and reverted the indentation to 2 spaces, and squashed/force-pushed to this branch. Diff should make more sense now. (Also added `react-renderer/tsfmt.json` for my editor.) Thanks for the review!

wodeni:
> FYI @wodeni I just made changes WRT the above suggestions and reverted the indentation to 2 spaces, and squashed/force-pushed to this branch. Diff should make more sense now. (Also added `react-renderer/tsfmt.json` for my editor.) Thanks for the review!

Thanks! BTW GitHub does give you an option to ignore whitespace changes, so the diff was no issue for me :D

react-renderer/tsconfig.json
maxkrieger: Lgtm

maxkrieger: Not sure how the polyfill coverage is but shouldn't be a problem

Issue #360, opened 2020-07-23
Flexible command-line argument order


strout18:
When running Penrose through the command line, the general format is `runpenrose X.sub X.sty X.dsl`. Switching the order of the three passed files results in a parsing error because a Substance file is being parsed as a Style file, for instance. It seems natural (and more lenient to the user's memory) to parse/eval files based on their extension, rather than the order in which they are passed.

hypotext:
This would be a good place to start: https://github.com/penrose/penrose/blob/web-runtime/src/Penrose/ShadowMain.hs#L66

It would be good to test that the existing programs work, and that programs in any new order work.

hypotext:
Also for this issue, it's fine to branch off of `master` since it's unrelated to `web-runtime` (cc @apatriarca)

PR #379, opened 2020-09-22
Use file extensions instead of order in command line arguments

apatriarca:
# Description

Related issues: #360 

This pull request change the way the input files are handled in the command line. The type of the source file was previously associated according to their order in the command line. With this change the type will be derived from the file extension. Each type is required to have only one associated input file.


hypotext:
Hey @apatriarca, thanks for working on this issue! The code looks good to me, and I was able to build, reproduce, and test this change as well. In the future, can you just document the new test cases that work for your code? Thank you!

PR #388, opened 2020-10-16
Update React dependencies and retrofit clients (IDE and automator)

wodeni:
# Description

Related issues: #336 

Our dependencies `create-react-app-ts` and `tslint` are both deprecated. This PR moves our web version to `create-react-app` and `eslint`, which support TypeScript now. 

The new TypeScript runtime also includes some breaking changes that affect both the automator and IDE. These two projects depend on `penrose-web` and use its API functions. This PR also aims to retrofit both modules to the new API.

* Due to the change from relative to absolute paths in `penrose-web`, the IDE fails to resolve modules of it. I used [ttsc](https://github.com/cevek/ttypescript/) (`tsc` with transformers) to transform paths in the compiled JavaScript files.
* Migration to CRA and eslint, as well as associated version bumps
* Additional configuration file for `build-lib` to bypass CRA limitations (e.g. `noEmit`).
* Add `tslog` for logging

* Reinstall both `penrose-web` and the IDE by following the wiki
* Repro: `npx ts-node ./index.tsx "batch" "substanceLibrary.json" "styleLibrary.json" "elementLibrary.json" "out"`

# Open questions

- [x] installation: make sure the installation process is reproducible 
- [x] Fix CI memory issue: https://github.com/facebook/create-react-app/issues/8096
- [x] Porting automator: there seem to be some type issues with automator
- [x] Rename `react-renderer`: I'm thinking __`penrose-web`__


react-renderer/src/engine/Optimizer.ts
```diff
@@ -243,11 +275,21 @@ export const step = (state: State, steps: number, evaluate = true) => {
         optParams.optStatus.tag = "UnconstrainedConverged";
         memoryManager.deleteExcept([]); // Clear allocated matrix, vector objects in L-BFGS params
         optParams.lbfgsInfo = defaultLbfgsParams;
-        console.log("Unconstrained converged with energy", energyVal, "gradient norm", normGrad);
+        console.log(
```
hypotext, line 278: Is it possible to turn off the "automatically put all arguments on a new line" setting? I'm fine with it being used in some files, but prefer my code to be more concise 🙂

react-renderer/src/engine/PropagateUpdate.ts
```diff
@@ -41,7 +41,7 @@ export const insertPending = (state: State) => {
       .map((p: Path) => [p, findShapeProperty(state.shapes, p).updated])
       // .updated is from `collectLabels` updating pending properties
       .reduce(
-        (trans: Translation, [path, v]: [Path, Value<number>]) =>
```
hypotext, line 44: Can you say why the types were made more permissive here? I'd prefer them to have more specific types if possible

hypotext:
Thanks for doing this! I took a quick look but am not super familiar with ts internals so can't say much about that. Is this ready to merge? 

Also, +1 to `Rename react-renderer: I'm thinking penrose-web`

react-renderer/src/engine/PropagateUpdate.ts
wodeni: The version bump actually made typechecking more rigorous. I think it has to do with the usage of `zip`. Same thing is true in the Optimizer module. Thanks for noting this. We will need to fix a couple of them before merging this PR

react-renderer/src/engine/Optimizer.ts
hypotext: Yeah, I commented on the `console.log` but I would also like the function calls, JS objects, etc. to be OK to write on one line, rather than spacing out all the arguments. Would that be possible?

react-renderer/src/engine/Optimizer.ts
```diff
@@ -531,9 +601,12 @@ const lbfgsInner = (grad_fx_k: any, ss: any[], ys: any[]): any => {
   // The length of any list should be the number of stored vectors
   const rhos = _.zip(ss, ys).map(([s, y]) => calculate_rho(s, y));
   const q_k = grad_fx_k;
-  // Note the order of alphas will be from k-1 through k-m for the push_r_forward loop  
+  // Note the order of alphas will be from k-1 through k-m for the push_r_forward loop
   // (Note that `reduce` is a left fold)
-  const [q_k_minus_m, alphas] = _.zip(rhos, ss, ys).reduce(pull_q_back, [q_k, []]);
+  const [q_k_minus_m, alphas] = (_.zip(rhos, ss, ys) as any).reduce(
```
wodeni, line 606: This is another instance of the new type errors. This is the hacky solution. @hypotext this might be a lodash thing or lack of typing in this module. I see lots of `any`s, which can be confusing to the compiler.

react-renderer/src/engine/Optimizer.ts
wodeni: I'd look at https://github.com/penrose/penrose/blob/57a0e7fe954146761270ff37f1190f34e234cea0/react-renderer/.eslintrc.js and refer to the link above. Not sure what the defaults are, but I'm open to change lint rules if you think it'd be helpful.

react-renderer/src/engine/Optimizer.ts
hypotext: I have had lodash problems with typing, but all the `any`s here are coming from `linear-algebra-js`, which doesn't have typings for its vector/matrix types. See here for when I am converting `number[]` to la.js types, which turns them into `any`. https://github.com/penrose/penrose/pull/388/files/57a0e7fe954146761270ff37f1190f34e234cea0#diff-e4cc280e366c644b840a2c549d706b1a08768a305ef0f9225323e0d622bc5c97R539

hypotext: Not sure why the diff link didn't work, but I just mean the function called `colVec` in `Optimizer`.

wodeni: `zip` will give you `undefined` sometimes (as noted in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23305). Not clear on why exactly it's a problem here. BTW here's a site that helps you get rid of lodash https://youmightnotneed.com/lodash/, not sure if it's any good.

`any`: One way is to add types to la.js if you think it helps. If it's too much work, then some casting is perhaps inevitable.

wodeni:
Retrofitted `automator`. Only two examples in `substanceLibrary.json` work with the new optimizer. Fresh out of the pipeline:
image
image

Repro: `npx ts-node ./index.tsx "batch" "substanceLibrary-new.json" "styleLibrary.json" "elementLibrary.json" "out"`

@hypotext can you update `venn.sty` (and any other Style files that can be easily fixed)?

react-renderer/src/engine/Optimizer.ts
wodeni: The reason for the new error is that your first zip statement has a type of `[number | undefined, any, any][]` perhaps due to the use the zip. I know `rhos` is definitely `number[]`, but the `any`s probably caused the extra `undefined`. I won't fix this in this PR. I think you'll need another pass on the Optimizer module to fix the types later.

Issue #375, opened 2020-09-02
Vectors should be first-class type in Style


keenancrane:
Anybody writing graphics code will expect to be able to work with (2D and 3D) vectors using nice syntax.  Currently, for instance, if we want to implement the computation

```
u = (c1 - c0)/norm(c1-c0)
```

where `u`, `c0`, and `c1` are 2D vectors, we have to write it out in terms of components

```
   LOCAL.ux = LOCAL.c1x - LOCAL.c0x
   LOCAL.uy = LOCAL.c1y - LOCAL.c0y
   LOCAL.m = sqrt( LOCAL.ux*LOCAL.ux + LOCAL.uy*LOCAL.uy )
   LOCAL.vx = LOCAL.ux / LOCAL.m
   LOCAL.vy = LOCAL.uy / LOCAL.m
```

Even if we get rid of the `LOCAL` keyword (a la #372), the code would still look something like

```
   var ux = c1x - c0x
   var uy = c1y - c0y
   var m = sqrt( ux*ux + uy*uy )
   var vx = ux / m
   var vy = uy / m
```

which is **extremely** verbose by graphics standards.  (Compare to languages like [GLSL](https://en.wikipedia.org/wiki/OpenGL_Shading_Language).)

Some things we really want to support:

- At least 2D, 3D, and 4D vector types (4D are super common for working with homogeneous coordinates...); nD may be overkill, but we should discuss.
- Definitely want syntactic sugar for vector operations, e.g., `u + v` not `add( u, v )`
- Also have to think about component accessors, e.g., saying something like `myScalar = myVector.y` to get the _y_ component, versus, `myScalar = myVector[1]`; some languages (GLSL?) support both.
- A "nice to have" feature, that actually can make a lot of code quite nice/clean, is "[swizzling](https://en.wikipedia.org/wiki/Swizzling_(computer_graphics))", e.g., `myVector.yxz = yourVector.xyz` will swap the _x_ and _y_ components.

In general, would be good to look at languages (such as GLSL, [HLSL](https://en.wikipedia.org/wiki/High-Level_Shading_Language)) and other graphics code (such as PBRT) to see what kinds of design decisions people have made.

**Note:** TensorFlow currently makes it tough to implement this stuff, since the TF paradigm is to basically never pull elements out of tensors.  Some of these issues should be resolved when we switch to @hypotext's autodiff system.

wodeni:
Addressed in #403

Issue #372, opened 2020-09-02
Allow local vars


hypotext:
Instead of requiring the user to declare variables like `LOCAL.x = ...` and refer to them as `LOCAL.x`...

allow the user to declare variables like `var x = ...` and refer to it as simply `x` (implicitly in the local scope).

keenancrane:
To give a concrete example, for the `hyperbolic` domain we wanted to implement the computation

```
u = (h.center - H.center)/norm( h.center - H.center)
```

i.e., compute some vector by normalizing the difference of two vectors.  Currently the best we can do is this (!):

```
   LOCAL.ux = LOCAL.c1x - LOCAL.c0x
   LOCAL.uy = LOCAL.c1y - LOCAL.c0y
   LOCAL.m = sqrt( LOCAL.ux*LOCAL.ux + LOCAL.uy*LOCAL.uy )
   LOCAL.vx = LOCAL.ux / LOCAL.m
   LOCAL.vy = LOCAL.uy / LOCAL.m
```

Part of making this code simpler is better handling of vector types, but it would already be very helpful to get rid of the `LOCAL.`s everywhere, turning it into this:

```
   local ux = c1x - c0x
   local uy = c1y - c0y
   local m = sqrt( ux*ux + uy*uy )
   local vx = ux / m
   local vy = uy / m
```

keenancrane:
Some design questions:

- What keyword(s) should we use? `local`? `val`? `var`? `let`? `const`? …. Or should there even be a keyword for local variables?  Having a keyword may make it easier to see where something was declared vs. assigned to.
- Should variables all be immutable?
- Should we have local scoping rules even within blocks?  E.g., variables that get declared inside a shape?  (Probably not, since we don't have nesting.)
- Should we have types?  E.g., `scalar` vs. `vector2` vs. `vector3` (just for readability)?  Note that we have a pretty domain-specific scope here, so it's not like we're gonna go wild with all sorts of crazy/custom types.

Probably not a bad idea to look at GLSL:

- Wikipedia page: https://en.wikipedia.org/wiki/OpenGL_Shading_Language
- Specification: https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.60.pdf

wodeni:
Addressed by #403.

PR #398, opened 2020-10-29
Style 1.01

wodeni:
# Description

This PR extends Style and the shape definitions to handle local vars and vector/matrix types, and ports existing Style programs to use these features. The motivation is to address the dealbreakers that @keenancrane pointed out when writing graphics code in Style.

Related issues: #375 #372 

- Parsing:
  - [x] optional type annotation and import parsing (not used for now)
  - [x] vector & matrix as first class objects in Style
  - [x] new syntax for lists and tuples
  - [x] separate definition of Style identifiers

- Compilation: 
  - Local variables (e.g. `x`) are handled by replacing them, and all references to them, by a field path with a Substance object named uniquely according to the block number and substitution number.
  - Anonymous statements (e.g. `encourage near(x.shape, y.shape)`) are handled by turning them into local variable-named statements with a unique anonymous keyword, using a counter
  - Vector-valued properties are sampled like other properties; to refer to anonymous elements (e.g. `x[0]`) in the list of varying paths, `AccessPath` is used in `findNestedVarying`

These include all the currently working Style programs for `penrose-web`. (These are mostly taken from the first sample, though some may not look exactly like the image since they are from old version of the system.)

* Orthogonal vector example from paper: `runpenrose linear-algebra-domain/twoVectorsPerp.sub linear-algebra-domain/linear-algebra-paper-simple.sty linear-algebra-domain/linear-algebra.dsl`

image

* Two sets with labels: `runpenrose set-theory-domain/twosets-simple.sub set-theory-domain/venn-small.sty  set-theory-domain/setTheory.dsl`

image

* Big set example with circles: `runpenrose set-theory-domain/tree.sub set-theory-domain/venn-opt-test.sty set-theory-domain/setTheory.dsl` 

image

* Big set example with tree:`runpenrose set-theory-domain/tree.sub set-theory-domain/tree.sty set-theory-domain/setTheory.dsl` 

image

* Continuous map: `runpenrose set-theory-domain/continuousmap.sub set-theory-domain/continuousmap.sty set-theory-domain/setTheory.dsl`

image

* Hyperbolic: `runpenrose hyperbolic-domain/hyperbolic-example.sub hyperbolic-domain/PoincareDisk.sty hyperbolic-domain/hyperbolic.dsl`

image

* Geometry (partial): `runpenrose geometry-domain/pythagorean-theorem-sugared.sub geometry-domain/euclidean-simple.sty geometry-domain/geometry.dsl`

image

* Sets (optimizing thru computation): `runpenrose set-theory-domain/multisets.sub set-theory-domain/venn-comp-test.sty  set-theory-domain/setTheory.dsl`

image

* Cross-domain mesh-set example: `runpenrose mesh-set-domain/DomainInterop.sub mesh-set-domain/DomainInterop.sty mesh-set-domain/DomainInterop.dsl`

image

# Checklist

# Open questions

- Ints are not auto-cast to float. (Right now if say `side` refers to a path that's an int, it won't get converted, and won't be VarAD in the autodiff, causing a runtime crash)
- Fully implement `AccessPath` in `Evaluator.ts` (see `TODO`s)
- Look for varying vals in other composite types like matrix, list, tuple in `findNestedVarying` (right now if a matrix contains varying vals, the path won't be found; `AccessPath` needs to include it)
- Dimensionality of vectors/matrices is hardcoded to 2 in some places
- Parametric types in general are handled poorly, with special cases to handle e.g. lists containing floats or vectors


wodeni:
Created a PR for this branch. @hypotext I'm seeing a lot of comments. Maybe some of them can go into the PR description?

wodeni:
Making some progress on parsing the example program. Skipping all imports for now (still parsing, just not including them in the AST). 

```
import Colors as C
import Vectors -- TODO: do something with this import?
import Penrose.Math as M -- Can you import a specific name, like "dot"?

OrthogonalVectors {
    const {
      scalar perpLen = 25.0
      -- For unit mark
      scalar markerPadding = 15.0
      scalar barSize = 5.0
    }
}
```

```haskell
[ Module
    "OrthogonalVectors"
    [ ( Namespace (StyVar "const")
      , [ VarAssign (TypeOf "scalar") "perpLen" (AFloat (Fix 25.0))
        , VarAssign (TypeOf "scalar") "markerPadding" (AFloat (Fix 15.0))
        , VarAssign (TypeOf "scalar") "barSize" (AFloat (Fix 5.0))
        ])
    ]
]
```

wodeni:
Parsed https://github.com/penrose/penrose/blob/4132619c2df90a4621319f056ec78ea2a060f381/examples/new-style/style-vecmath-local.sty
```Haskell
ast =
  [ ( Namespace (StyVar "Const")
    , [ VarAssign Nothing (LocalVar "perpLen") (AFloat (Fix 25.0))
      , VarAssign Nothing (LocalVar "markerPadding") (AFloat (Fix 15.0))
      , VarAssign Nothing (LocalVar "barSize") (AFloat (Fix 5.0))
      , VarAssign Nothing (LocalVar "vectorSpaceSize") (IntLit 350)
      , VarAssign Nothing (LocalVar "repelWeight") (AFloat (Fix 0.7))
      , VarAssign Nothing (LocalVar "arrowheadSize") (AFloat (Fix 0.6))
      , VarAssign Nothing (LocalVar "lineThickness") (IntLit 1)
      ])
  , ( Namespace (StyVar "Colors")
    , [ VarAssign
          Nothing
          (LocalVar "white")
          (CompApp "rgba" [IntLit 1, IntLit 1, IntLit 1, IntLit 1])
      , VarAssign
          Nothing
          (LocalVar "lightBlue")
          (CompApp
             "rgba"
             [ AFloat (Fix 0.1)
             , AFloat (Fix 0.1)
             , AFloat (Fix 0.9)
             , AFloat (Fix 0.1)
             ])
      , VarAssign
          Nothing
          (LocalVar "darkBlue")
          (CompApp
             "rgba"
             [ BinOp
                 Divide
                 (EPath (FieldPath (BStyVar (StyVar "lightBlue")) "r"))
                 (IntLit 2)
             , BinOp
                 Divide
                 (EPath (FieldPath (BStyVar (StyVar "lightBlue")) "g"))
                 (IntLit 2)
             , BinOp
                 Divide
                 (EPath (FieldPath (BStyVar (StyVar "lightBlue")) "b"))
                 (IntLit 2)
             , AFloat (Fix 0.5)
             ])
      , VarAssign
          Nothing
          (LocalVar "darkGray")
          (CompApp
             "rgba"
             [AFloat (Fix 0.4), AFloat (Fix 0.4), AFloat (Fix 0.4), IntLit 1])
      , VarAssign
          Nothing
          (LocalVar "gray")
          (CompApp
             "rgba"
             [AFloat (Fix 0.6), AFloat (Fix 0.6), AFloat (Fix 0.6), IntLit 1])
      , VarAssign
          Nothing
          (LocalVar "green")
          (CompApp "rgba" [IntLit 0, AFloat (Fix 0.8), IntLit 0, IntLit 1])
      , VarAssign
          Nothing
          (LocalVar "none")
          (CompApp "rgba" [IntLit 0, IntLit 0, IntLit 0, IntLit 0])
      ])
  , ( Namespace (StyVar "testing")
    , [ VarAssign
          Nothing
          (LocalVar "x")
          (Tuple (IntLit 1) (Tuple (IntLit 2) (IntLit 3)))
      , VarAssign
          Nothing
          (LocalVar "y")
          (List
             [ UOp UMinus (AFloat (Fix 2.0))
             , EPath (FieldPath (BStyVar (StyVar "const")) "perpLen")
             , BinOp
                 BPlus
                 (EPath (FieldPath (BStyVar (StyVar "const")) "markerPadding"))
                 (IntLit 3)
             ])
      , VarAssign
          Nothing
          (LocalVar "a")
          (Vector [UOp UMinus (AFloat (Fix 1.0)), AFloat Vary])
      , VarAssign
          Nothing
          (LocalVar "a1")
          (BinOp
             BPlus
             (Vector [UOp UMinus (AFloat (Fix 1.0)), AFloat (Fix 2.0)])
             (Vector [AFloat (Fix 100000.0), AFloat (Fix 2.0)]))
      , VarAssign
          Nothing
          (LocalVar "m")
          (Vector
             [ EVar (LocalVar "a")
             , Vector [UOp UMinus (AFloat (Fix 1.0)), AFloat (Fix 2.0)]
             ])
      , VarAssign
          Nothing
          (LocalVar "v")
          (BinOp
             Divide
             (BinOp BPlus (EVar (LocalVar "a")) (Vector [IntLit 2, IntLit 900]))
             (BinOp BPlus (AFloat (Fix 4.0)) (IntLit 3)))
      , VarAssign
          Nothing
          (LocalVar "asum")
          (BinOp
             BPlus
             (VectorAccess (EVar (LocalVar "a")) (IntLit 1))
             (VectorAccess (EVar (LocalVar "a")) (IntLit 0)))
      , VarAssign Nothing (LocalVar "c") (IntLit 0)
      , VarAssign Nothing (LocalVar "b") (IntLit 1)
      , VarAssign
          Nothing
          (LocalVar "msum")
          (BinOp
             BPlus
             (MatrixAccess (EVar (LocalVar "m")) (IntLit 1) (IntLit 0))
             (MatrixAccess
                (EVar (LocalVar "m"))
                (EVar (LocalVar "c"))
                (EVar (LocalVar "b"))))
      , VarAssign Nothing (LocalVar "nv") (UOp UMinus (EVar (LocalVar "v")))
      , VarAssign
          Nothing
          (LocalVar "t1")
          (MatrixAccess (EVar (LocalVar "x")) (IntLit 1) (EVar (LocalVar "b")))
      ])
  , ( Select
        (Selector
           { selHead =
               [PatternDecl' (STCtor "VectorSpace" []) (BStyVar (StyVar "U"))]
           , selWith = []
           , selWhere = []
           , selNamespace = Nothing
           })
    , [ VarAssign
          Nothing
          (LocalVar "axisSize")
          (BinOp
             Divide
             (EPath (FieldPath (BStyVar (StyVar "const")) "vectorSpaceSize"))
             (IntLit 2))
      , PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "U")) "origin")
          (Vector [AFloat Vary, AFloat Vary])
      , VarAssign
          Nothing
          (LocalVar "o")
          (EPath (FieldPath (BStyVar (StyVar "U")) "origin"))
      , PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "U")) "shape")
          (Ctor
             "Square"
             [ PropertyDecl
                 "center"
                 (EPath (FieldPath (BStyVar (StyVar "U")) "origin"))
             , PropertyDecl
                 "side"
                 (EPath (FieldPath (BStyVar (StyVar "const")) "vectorSpaceSize"))
             , PropertyDecl
                 "color"
                 (EPath (FieldPath (BStyVar (StyVar "C")) "none"))
             , PropertyDecl
                 "stroke"
                 (EPath (FieldPath (BStyVar (StyVar "C")) "black"))
             ])
      , PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "U")) "xAxis")
          (Ctor
             "Arrow"
             [ PropertyDecl
                 "start"
                 (Vector
                    [ BinOp
                        BMinus
                        (EPath (FieldPath (BStyVar (StyVar "o")) "x"))
                        (EVar (LocalVar "axisSize"))
                    , EPath (FieldPath (BStyVar (StyVar "o")) "y")
                    ])
             , PropertyDecl
                 "end"
                 (Vector
                    [ BinOp
                        BPlus
                        (EPath (FieldPath (BStyVar (StyVar "o")) "x"))
                        (EVar (LocalVar "axisSize"))
                    , EPath (FieldPath (BStyVar (StyVar "o")) "y")
                    ])
             , PropertyDecl
                 "thickness"
                 (EPath (FieldPath (BStyVar (StyVar "const")) "lineThickness"))
             , PropertyDecl "style" (StringLit "solid")
             , PropertyDecl
                 "color"
                 (EPath (FieldPath (BStyVar (StyVar "U")) "axisColor"))
             , PropertyDecl
                 "arrowheadSize"
                 (EPath (FieldPath (BStyVar (StyVar "const")) "arrowheadSize"))
             ])
      , PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "U")) "yAxis")
          (Ctor
             "Arrow"
             [ PropertyDecl
                 "start"
                 (Vector
                    [ EPath (FieldPath (BStyVar (StyVar "o")) "x")
                    , BinOp
                        BPlus
                        (EPath (FieldPath (BStyVar (StyVar "o")) "y"))
                        (EVar (LocalVar "axisSize"))
                    ])
             , PropertyDecl
                 "end"
                 (Vector
                    [ EPath (FieldPath (BStyVar (StyVar "o")) "x")
                    , BinOp
                        BPlus
                        (EPath (FieldPath (BStyVar (StyVar "o")) "y"))
                        (EVar (LocalVar "axisSize"))
                    ])
             , PropertyDecl
                 "thickness"
                 (EPath (FieldPath (BStyVar (StyVar "const")) "lineThickness"))
             , PropertyDecl "style" (StringLit "solid")
             , PropertyDecl
                 "color"
                 (EPath (FieldPath (BStyVar (StyVar "U")) "axisColor"))
             , PropertyDecl
                 "arrowheadSize"
                 (EPath (FieldPath (BStyVar (StyVar "const")) "arrowheadSize"))
             ])
      , PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "U")) "text")
          (Ctor
             "Text"
             [ PropertyDecl
                 "string"
                 (EPath (FieldPath (BStyVar (StyVar "U")) "label"))
             , PropertyDecl
                 "center"
                 (Vector
                    [ BinOp
                        BMinus
                        (EPath
                           (PropertyPath (BStyVar (StyVar "U")) "origin" "x"))
                        (EVar (LocalVar "axisSize"))
                    , BinOp
                        BPlus
                        (EPath
                           (PropertyPath (BStyVar (StyVar "U")) "origin" "y"))
                        (EVar (LocalVar "axisSize"))
                    ])
             , PropertyDecl
                 "color"
                 (EPath (FieldPath (BStyVar (StyVar "U")) "axisColor"))
             ])
      ])
  , ( Select
        (Selector
           { selHead =
               [ PatternDecl' (STCtor "Vector" []) (BStyVar (StyVar "u"))
               , PatternDecl' (STCtor "VectorSpace" []) (BStyVar (StyVar "U"))
               ]
           , selWith = []
           , selWhere =
               [ RelPred
                   "In"
                   [ PE (SEBind (BStyVar (StyVar "u")))
                   , PE (SEBind (BStyVar (StyVar "U")))
                   ]
               ]
           , selNamespace = Nothing
           })
    , [ PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "u")) "text")
          (Ctor
             "Text"
             [ PropertyDecl
                 "string"
                 (EPath (FieldPath (BStyVar (StyVar "u")) "label"))
             , PropertyDecl
                 "color"
                 (EPath (PropertyPath (BStyVar (StyVar "u")) "shape" "color"))
             ])
      , PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "u")) "arrow")
          (Ctor
             "Arrow"
             [ PropertyDecl
                 "start"
                 (EPath (FieldPath (BStyVar (StyVar "U")) "origin"))
             , PropertyDecl "end" (Vector [AFloat Vary, AFloat Vary])
             , PropertyDecl "thickness" (AFloat (Fix 3.0))
             , PropertyDecl
                 "color"
                 (EPath (FieldPath (BStyVar (StyVar "C")) "blue"))
             , PropertyDecl
                 "arrowheadSize"
                 (EPath (FieldPath (BStyVar (StyVar "const")) "arrowheadSize"))
             ])
      , PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "u")) "vector")
          (BinOp
             BMinus
             (EPath (PropertyPath (BStyVar (StyVar "u")) "arrow" "end"))
             (EPath (PropertyPath (BStyVar (StyVar "u")) "arrow" "start")))
      , VarAssign
          Nothing
          (LocalVar "contains")
          (EPath (FieldPath (BStyVar (StyVar "functions")) "contains"))
      , AnonAssign
          (ConstrFn
             "contains"
             [ EPath (FieldPath (BStyVar (StyVar "U")) "shape")
             , EPath (FieldPath (BStyVar (StyVar "u")) "shape")
             ])
      , AnonAssign
          (ConstrFn
             "contains"
             [ EPath (FieldPath (BStyVar (StyVar "U")) "shape")
             , EPath (FieldPath (BStyVar (StyVar "u")) "text")
             ])
      , AnonAssign
          (ConstrFn
             "atDist"
             [ EPath (FieldPath (BStyVar (StyVar "u")) "shape")
             , EPath (FieldPath (BStyVar (StyVar "u")) "text")
             , AFloat (Fix 5.0)
             ])
      , AnonAssign
          (Layering
             (FieldPath (BStyVar (StyVar "U")) "xAxis")
             (FieldPath (BStyVar (StyVar "u")) "text"))
      , AnonAssign
          (Layering
             (FieldPath (BStyVar (StyVar "U")) "yAxis")
             (FieldPath (BStyVar (StyVar "u")) "text"))
      ])
  , ( Select
        (Selector
           { selHead =
               [ PatternDecl' (STCtor "Vector" []) (BStyVar (StyVar "u"))
               , PatternDecl' (STCtor "Vector" []) (BStyVar (StyVar "v"))
               ]
           , selWith =
               [PatternDecl' (STCtor "VectorSpace" []) (BStyVar (StyVar "U"))]
           , selWhere =
               [ RelPred
                   "Orthogonal"
                   [ PE (SEBind (BStyVar (StyVar "u")))
                   , PE (SEBind (BStyVar (StyVar "v")))
                   ]
               , RelPred
                   "In"
                   [ PE (SEBind (BStyVar (StyVar "u")))
                   , PE (SEBind (BStyVar (StyVar "U")))
                   ]
               , RelPred
                   "In"
                   [ PE (SEBind (BStyVar (StyVar "v")))
                   , PE (SEBind (BStyVar (StyVar "U")))
                   ]
               ]
           , selNamespace = Nothing
           })
    , [ VarAssign
          Nothing
          (LocalVar "startR")
          (EPath (PropertyPath (BStyVar (StyVar "u")) "shape" "start"))
      , VarAssign
          Nothing
          (LocalVar "endR")
          (EPath (PropertyPath (BStyVar (StyVar "u")) "shape" "end"))
      , VarAssign
          Nothing
          (LocalVar "startL")
          (EPath (PropertyPath (BStyVar (StyVar "v")) "shape" "start"))
      , VarAssign
          Nothing
          (LocalVar "endL")
          (EPath (PropertyPath (BStyVar (StyVar "v")) "shape" "end"))
      , VarAssign
          Nothing
          (LocalVar "dirR")
          (CompApp
             "normalize"
             [BinOp BMinus (EVar (LocalVar "endR")) (EVar (LocalVar "startR"))])
      , VarAssign
          Nothing
          (LocalVar "dirL")
          (CompApp
             "normalize"
             [BinOp BMinus (EVar (LocalVar "endL")) (EVar (LocalVar "startL"))])
      , VarAssign
          Nothing
          (LocalVar "ptL")
          (BinOp
             BPlus
             (EVar (LocalVar "startR"))
             (BinOp
                Multiply
                (EPath (FieldPath (BStyVar (StyVar "const")) "perpLen"))
                (EVar (LocalVar "dirL"))))
      , VarAssign
          Nothing
          (LocalVar "ptR")
          (BinOp
             BPlus
             (EVar (LocalVar "startR"))
             (BinOp
                Multiply
                (EPath (FieldPath (BStyVar (StyVar "const")) "perpLen"))
                (EVar (LocalVar "dirR"))))
      , VarAssign
          Nothing
          (LocalVar "ptLR")
          (BinOp
             BPlus
             (EVar (LocalVar "ptL"))
             (BinOp Multiply (EVar (LocalVar "len")) (EVar (LocalVar "dirR"))))
      , VarAssign
          Nothing
          (LocalVar "pts")
          (List
             [ EVar (LocalVar "ptL")
             , EVar (LocalVar "ptLR")
             , EVar (LocalVar "ptR")
             ])
      , VarAssign
          Nothing
          (LocalVar "toPath")
          (EPath (FieldPath (BStyVar (StyVar "functions")) "pathFromPoints"))
      , VarAssign
          Nothing
          (LocalVar "perpMark")
          (Ctor
             "Curve"
             [ PropertyDecl
                 "pathData"
                 (CompApp "toPath" [StringLit "closed", EVar (LocalVar "pts")])
             , PropertyDecl "strokeWidth" (AFloat (Fix 2.0))
             , PropertyDecl
                 "color"
                 (EPath (FieldPath (BStyVar (StyVar "C")) "black"))
             , PropertyDecl
                 "fill"
                 (EPath (FieldPath (BStyVar (StyVar "C")) "white"))
             ])
      , VarAssign
          Nothing
          (LocalVar "eq")
          (EPath (FieldPath (BStyVar (StyVar "functions")) "equal"))
      , AnonAssign
          (ObjFn
             "eq"
             [ CompApp
                 "dot"
                 [ EPath (FieldPath (BStyVar (StyVar "u")) "vector")
                 , EPath (FieldPath (BStyVar (StyVar "v")) "vector")
                 ]
             , AFloat (Fix 0.0)
             ])
      , AnonAssign
          (Layering
             (FieldPath (BStyVar (StyVar "LOCAL")) "perpMark")
             (FieldPath (BStyVar (StyVar "v")) "shape"))
      , AnonAssign
          (Layering
             (FieldPath (BStyVar (StyVar "LOCAL")) "perpMark")
             (FieldPath (BStyVar (StyVar "u")) "shape"))
      ])
  , ( Select
        (Selector
           { selHead =
               [PatternDecl' (STCtor "Vector" []) (BStyVar (StyVar "v"))]
           , selWith =
               [ PatternDecl' (STCtor "VectorSpace" []) (BStyVar (StyVar "U"))
               , PatternDecl' (STCtor "Vector" []) (BStyVar (StyVar "w"))
               ]
           , selWhere =
               [ RelPred
                   "In"
                   [ PE (SEBind (BStyVar (StyVar "v")))
                   , PE (SEBind (BStyVar (StyVar "U")))
                   ]
               , RelPred "Unit" [PE (SEBind (BStyVar (StyVar "v")))]
               , RelPred
                   "Orthogonal"
                   [ PE (SEBind (BStyVar (StyVar "v")))
                   , PE (SEBind (BStyVar (StyVar "w")))
                   ]
               ]
           , selNamespace = Nothing
           })
    , [ VarAssign Nothing (LocalVar "strokeWidth") (IntLit 2)
      , VarAssign
          Nothing
          (LocalVar "toPath")
          (EPath (FieldPath (BStyVar (StyVar "functions")) "pathFromPoints"))
      , VarAssign
          Nothing
          (LocalVar "dir")
          (CompApp
             "normalize"
             [ BinOp
                 BMinus
                 (EPath (PropertyPath (BStyVar (StyVar "w")) "shape" "end"))
                 (EPath (PropertyPath (BStyVar (StyVar "w")) "shape" "start"))
             ])
      , VarAssign
          Nothing
          (LocalVar "normal")
          (UOp UMinus (EVar (LocalVar "dir")))
      , VarAssign
          Nothing
          (LocalVar "markStart")
          (BinOp
             BPlus
             (EPath (PropertyPath (BStyVar (StyVar "v")) "shape" "start"))
             (BinOp
                Multiply
                (EVar (LocalVar "padding"))
                (EVar (LocalVar "normal"))))
      , VarAssign
          Nothing
          (LocalVar "markEnd")
          (BinOp
             BPlus
             (EPath (PropertyPath (BStyVar (StyVar "v")) "shape" "end"))
             (BinOp
                Multiply
                (EVar (LocalVar "padding"))
                (EVar (LocalVar "normal"))))
      , PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "v")) "markerLine")
          (List [EVar (LocalVar "markStart"), EVar (LocalVar "markEnd")])
      , PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "v")) "unitMarkerLine")
          (Ctor
             "Curve"
             [ PropertyDecl
                 "pathData"
                 (CompApp
                    "toPath"
                    [ StringLit "open"
                    , EPath (FieldPath (BStyVar (StyVar "v")) "markerLine")
                    ])
             , PropertyDecl "strokeWidth" (EVar (LocalVar "strokeWidth"))
             , PropertyDecl
                 "color"
                 (EPath (FieldPath (BStyVar (StyVar "C")) "black"))
             , PropertyDecl
                 "fill"
                 (EPath (FieldPath (BStyVar (StyVar "C")) "none"))
             ])
      , VarAssign
          Nothing
          (LocalVar "rot90CW")
          (Vector
             [ Vector [IntLit 0, IntLit 1]
             , Vector [UOp UMinus (IntLit 1), IntLit 0]
             ])
      , VarAssign
          Nothing
          (LocalVar "markNormal")
          (CompApp "mul" [EVar (LocalVar "rot90CW"), EVar (LocalVar "normal")])
      , VarAssign
          Nothing
          (LocalVar "c")
          (EPath (FieldPath (BStyVar (StyVar "const")) "barSize"))
      , VarAssign
          Nothing
          (LocalVar "halfvec")
          (BinOp Multiply (EVar (LocalVar "c")) (EVar (LocalVar "markNormal")))
      , PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "v")) "unitMarkerEnd1")
          (Ctor
             "Curve"
             [ PropertyDecl
                 "pathData"
                 (CompApp
                    "toPath"
                    [ StringLit "open"
                    , List
                        [ BinOp
                            BMinus
                            (EVar (LocalVar "markStart"))
                            (EVar (LocalVar "halfvec"))
                        , BinOp
                            BPlus
                            (EVar (LocalVar "markStart"))
                            (EVar (LocalVar "halfvec"))
                        ]
                    ])
             , PropertyDecl "strokeWidth" (EVar (LocalVar "strokeWidth"))
             , PropertyDecl
                 "color"
                 (EPath (FieldPath (BStyVar (StyVar "C")) "black"))
             , PropertyDecl
                 "fill"
                 (EPath (FieldPath (BStyVar (StyVar "C")) "none"))
             ])
      , PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "v")) "unitMarkerEnd2")
          (Ctor
             "Curve"
             [ PropertyDecl
                 "pathData"
                 (CompApp
                    "toPath"
                    [ StringLit "open"
                    , List
                        [ BinOp
                            BMinus
                            (EVar (LocalVar "markEnd"))
                            (EVar (LocalVar "halfvec"))
                        , BinOp
                            BPlus
                            (EVar (LocalVar "markEnd"))
                            (EVar (LocalVar "halfvec"))
                        ]
                    ])
             , PropertyDecl "strokeWidth" (EVar (LocalVar "strokeWidth"))
             , PropertyDecl
                 "color"
                 (EPath (FieldPath (BStyVar (StyVar "C")) "black"))
             , PropertyDecl
                 "fill"
                 (EPath (FieldPath (BStyVar (StyVar "C")) "none"))
             ])
      , VarAssign
          Nothing
          (LocalVar "midpointLoc")
          (BinOp
             Divide
             (BinOp
                BPlus
                (VectorAccess
                   (EPath (FieldPath (BStyVar (StyVar "v")) "markerLine"))
                   (IntLit 0))
                (VectorAccess
                   (EPath (FieldPath (BStyVar (StyVar "v")) "markerLine"))
                   (IntLit 1)))
             (IntLit 2))
      , VarAssign
          Nothing
          (LocalVar "labelPos")
          (BinOp
             BPlus
             (EVar (LocalVar "midpointLoc"))
             (BinOp
                Multiply
                (EPath (FieldPath (BStyVar (StyVar "const")) "markerPadding"))
                (EVar (LocalVar "normal"))))
      , PathAssign
          Nothing
          (FieldPath (BStyVar (StyVar "v")) "unitMarkerText")
          (Ctor
             "Text"
             [ PropertyDecl "string" (StringLit "1")
             , PropertyDecl "center" (EVar (LocalVar "labelPos"))
             , PropertyDecl
                 "color"
                 (EPath (FieldPath (BStyVar (StyVar "C")) "black"))
             ])
      , AnonAssign
          (Layering
             (FieldPath (BStyVar (StyVar "U")) "xAxis")
             (FieldPath (BStyVar (StyVar "v")) "unitMarkerLine"))
      , AnonAssign
          (Layering
             (FieldPath (BStyVar (StyVar "U")) "yAxis")
             (FieldPath (BStyVar (StyVar "v")) "unitMarkerLine"))
      ])
  , ( Select
        (Selector
           { selHead =
               [PatternDecl' (STCtor "Vector" []) (BSubVar (VarConst "x2"))]
           , selWith = []
           , selWhere = []
           , selNamespace = Nothing
           })
    , [ Override
          (PropertyPath (BSubVar (VarConst "x2")) "shape" "color")
          (EPath (FieldPath (BStyVar (StyVar "C")) "green"))
      ])
  ]

```

tsdef/types.d.ts
```diff
@@ -176,6 +201,10 @@ interface IPropertyPath {
   contents: [BindingForm, string, string];
 }
 
+interface ITypePropertyPath {
+  tag: "TypePropertyPath";
+}
```
hypotext, line 206: @wodeni Should this type be just a tag, with no contents? I wasn't sure what this corresponded to in the Haskell grammar.

wodeni: Oh it’s probably left in the HS file. We are not actively using this, right? I thought we opted for the [] access instead

hypotext: I see, this is the type property access, like `v.x`. Yeah, we aren't actively using it. Thanks for clarification.

src/Penrose/Style.hs
```diff
@@ -239,6 +245,10 @@ data Expr
     | UOp UnaryOp Expr
     | List [Expr]
     | Tuple Expr Expr
+    | Vector [Expr]
+    | Matrix [Expr]
+    | VectorAccess Expr Expr -- ^ Reference to vector, index
```
hypotext, line 250: @wodeni I just realized that the VectorAccess and MatrixAccess hardcode the dimensions of the vectors and matrices. Could you generalize them to be a list of indices (so we can deal with 3-vectors, etc.)? Thanks!

src/Penrose/Style.hs
```diff
@@ -230,6 +235,7 @@ data Expr
     | AFloat AnnoFloat
     | StringLit String
     | BoolLit Bool
+    | EVar LocalVar
```
hypotext, line 238: @wodeni Could we make local vars part of `Path` instead of `Expr`? I know that we missed this in the Style 1.1 grammar doc, but the rationale is that some of the expressions that currently only involve `Path` should be writeable using local vars. 

Example: `layer v.shape above perpMark` (where `perpMark` is a local var that is assigned to a shape)

hypotext:
Ok, some wacky merge commits later here is the diff for merged version: https://github.com/penrose/penrose/pull/398/commits/7164fa0c95f74a29ba8e753694317e81a2824e70

And AST: https://gist.github.com/hypotext/f9f18dd00323c1a234e58dcdd6c89a5f

The only bug I see (besides the comment I made above) is that vector accesses, like `o[0]`, are parsed as matrix accesses, when they should be vector accesses in the AST. Could you take a look at that @wodeni?

hypotext:
FYI, I plan to make a pretty big breaking change tomorrow: All float properties (like `startX, startY, endX, endY, x, y`) are going to be replaced by their vector equivalents (`start`, `end`, `center`) in the backend shapes, frontend shape rendering code, and objective/constraint/function code. (I can't do this in a backward-compatible manner because then it's not clear which set of properties should be used to render the function.) LMK if you have any reservations @wodeni.

wodeni:
> In the most general case, the VectorAccess and MatrixAccess should still have the first arg as `Expr`, right? e.g. someone might write `f(args)[0]`. Can we parse that form instead?

We probably missed that in the first pass. Do you mind adding a few examples in the style test file so we can test it?

wodeni:
> FYI, I plan to make a pretty big breaking change tomorrow: All float properties (like `startX, startY, endX, endY, x, y`) are going to be replaced by their vector equivalents (`start`, `end`, `center`) in the backend shapes, frontend shape rendering code, and objective/constraint/function code. (I can't do this in a backward-compatible manner because then it's not clear which set of properties should be used to render the function.) LMK if you have any reservations @wodeni.

That does sound like a huge change. If we are talking about all props then yes I do want to know your rationale. Also I don’t think changes to shapedefs belong to this PR?

hypotext:
> We probably missed that in the first pass. Do you mind adding a few examples in the style test file so we can test it?

Yep, added in most recent commit.

hypotext:
@wodeni Hey, another minor parser request. It looks like boolean literals are getting parsed as local vars (which makes sense, as they look similar). Can you change the parser to accept `true` and `false` as bool lits?

Test case (commented out in most recent commit): https://github.com/penrose/penrose/pull/398/commits/968dff7161f39ce551a61c6fb04beeb894496695#diff-4dcdd303cd238ea2ac30bcc15c7ebffbc603ebd9230ebdeac9a6969887ed7edaR82

hypotext:
I also noticed that the parser now doesn't parse named fns, e.g.: `ensure x.fn = equal(0.0, 0.0)`. We probably want those to continue working. :)

Example in `venn-small.sty` (most recent commit):

```
penrose: Error when parsing Style: 
36:26:
   |
36 |     x.containFn = ensure contains(x.shape, x.text) -- NEW
   |                          ^
unexpected 'c'
expecting '*', '+', '-', '/', '^', '}', or newline
```

hypotext:
Also ran into a weird parse error here: https://github.com/penrose/penrose/pull/398/commits/c0f0c682656d4828f0b028769a9e0e9bdccd3947#diff-3c5304a2337ed9ba947c774af21f5ec47163ae5a38b96174cacd8d903d5bffffR429

Not sure why it doesn't parse an anon shape when there is a similar example in `style-vecmath-local.sty`.

```
penrose: Error when parsing Style: 
429:4:
    |
429 |    shape = Line {
    |    ^
unexpected 's'
expecting '}'
```

repro: `runpenrose mesh-set-domain/DomainInterop.sub mesh-set-domain/DomainInterop.sty mesh-set-domain/DomainInterop.dsl`

hypotext:
Ok, I've implemented all the compiler stuff for local vars, vectors, and vector shape properties, and ported the relevant Style/contrib programs. All the existing Style programs work with the new features (pending stuff that depends on parser). I also pulled + merged with master. 

Once the parser is good to go, I can clean up and document this PR and ask for your review @wodeni, then make changes and merge to master. Questions:

* Do we think we can turn this around (i.e. get stuff merged) by Friday? (I'm not blocked for Thurs, since I'll be cleaning up our documentation)
* For reviewing my changes, do you prefer if I squash everything into one commit, or not? Wasn't sure if you preferred parser stuff to be kept separate. (I'll do that the one-commit squash for the merge)

wodeni:
> @wodeni Hey, another minor parser request. It looks like boolean literals are getting parsed as local vars (which makes sense, as they look similar). Can you change the parser to accept `true` and `false` as bool lits?
> 
> Test case (commented out in most recent commit): [968dff7#diff-4dcdd303cd238ea2ac30bcc15c7ebffbc603ebd9230ebdeac9a6969887ed7edaR82](https://github.com/penrose/penrose/commit/968dff7161f39ce551a61c6fb04beeb894496695#diff-4dcdd303cd238ea2ac30bcc15c7ebffbc603ebd9230ebdeac9a6969887ed7edaR82)

Our convention has been "True" and "False" for boolean literals (see other style files). That aside, I do realize that our like of reserved words is very out of date. Any chance we can include that in the language spec?

hypotext:
> Our convention has been "True" and "False" for boolean literals (see other style files). That aside, I do realize that our like of reserved words is very out of date. Any chance we can include that in the language spec?

Do you mean in the Style 1.01 doc with the grammar revision? If so, yes, I can/should put the list of rwords there.

wodeni:
> venn-small.sty

Resolved in 457a41f867ca414d8516a94273c6c2ea12eb9f2e. The issue was the new members of arithmetic exprs get parsed first. Promoting constraints/objectives in the list solved the problem.

wodeni:
> > Our convention has been "True" and "False" for boolean literals (see other style files). That aside, I do realize that our like of reserved words is very out of date. Any chance we can include that in the language spec?
> 
> Do you mean in the Style 1.01 doc with the grammar revision? If so, yes, I can/should put the list of rwords there.

Yep, that'd be helpful for us going forward.

wodeni:
> runpenrose mesh-set-domain/DomainInterop.sub mesh-set-domain/DomainInterop.sty mesh-set-domain/DomainInterop.dsl

This is expected, as `shape` is one of our reserved words for type annotation. I temporarily removed the optional type annotation. If we decide to add it back in, none of the local vars or fields can be named as `shape` (including `x.shape`).

wodeni:
> * Do we think we can turn this around (i.e. get stuff merged) by Friday? (I'm not blocked for Thurs, since I'll be cleaning up our documentation)

Just fixed all the parser issues. To merge into master, we should probably address the CI failure (likely caused by shape prop changes in the backend and unknown issues in the frontend).

> * For reviewing my changes, do you prefer if I squash everything into one commit, or not? Wasn't sure if you preferred parser stuff to be kept separate. (I'll do that the one-commit squash for the merge)
Doesn't really matter for review purposes. I'm not really a fan of huge squashed commits.

hypotext:
Cleaned things up over at #403.

wodeni:
Thinking about our workflow a bit more, I still prefer keeping the original commit history and discussion when we merge PRs. These individual commits let us find out the original author for each line later if any of us has trouble understanding it. It also keeps the history of contribution, which is important as we have more contributors. Squashing makes more sense for individual feature/fix branches. For large PR like this, one could squash their own commits if the history is a little messy, but squashing the whole PR makes it more difficult for collaborators. If you do want to note the milestones in the commit history, I’d suggest using tags instead.

@hypotext: Can you update the description of this PR instead of merging #403?

hypotext:
Hey @wodeni I see the case for leaving some commits here.  I agree that keeping some of the commits with authorship is better for `git blame`.

In general, though, if we're going to add commits to the main branch, I'd prefer them to be as clean as possible, and for each one to not break the build (this is important for `git bisect`, etc). I just think this branch history is too messy to merge, as there are commits that break the build for only part of the system, and then fix it for only part of the system, or add/remove one line of logging, etc, whereas if it's all one commit, then each change can be understood as part of one cohesive change to the system. And all the discussion for this branch is available since #403 links here.

BTW it's hard to squash individual commits within a branch, as git then "replays" subsequent commits on the branch and causes merge conflicts on each replay (especially if there are merge commits with other branches), but if you know if a better way to do it, I'm happy to try.

For this PR specifically, I've also made a lot of cleanups which are already squashed into the one-commit branch #403. I'm not sure how I would "remove" those cleanups and apply them onto this branch. So, can we move forward with #403 since it's time-sensitive, and start a conversation about a PR cleanup policy in the future? I am open to keeping multiple commits for big changes.

hypotext:
Also @maxkrieger feel free to chime in with your preferences on future commit style!

(Part of my desire to clean things up, btw, is because we are an open source project. Don't want guests poking around in our history and seeing old design discussions, debugging dead-ends, etc. If we were closed source I'd be more amenable to keeping a more messy history. 🙂 )

wodeni:
> In general, though, if we're going to add commits to the main branch, I'd prefer them to be as clean as possible, and for each one to not break the build (this is important for `git bisect`, etc). I just think this branch history is too messy to merge, as there are commits that break the build for only part of the system, and then fix it for only part of the system, or add/remove one line of logging, etc, whereas if it's all one commit, then each change can be understood as part of one cohesive change to the system. And all the discussion for this branch is available since #403 links here.

For `git bisect`, my understanding is that you generally want to isolate a commit for some unknown behavior in the system. In that case, I would argue that smaller commits w/ meaningful commit message are even more helpful. In general, I think everyone is doing a great job at documenting their commits! I would suggest, again, using `git tag` for marking stable versions that can later be used with `bisect` if desired. You can always `bisect skip` version ranges, as demonstrated in the [official doc](https://git-scm.com/docs/git-bisect#_bisect_skip).

> BTW it's hard to squash individual commits within a branch, as git then "replays" subsequent commits on the branch and causes merge conflicts on each replay (especially if there are merge commits with other branches), but if you know if a better way to do it, I'm happy to try.

My original suggestion was that if one person is working on a PR, they should feel free to squash breaking commits that were originally pushed for testing. Like you said, in branches like this one, squashing is just more effort with very little return to me.

> For this PR specifically, I've also made a lot of cleanups which are already squashed into the one-commit branch #403. I'm not sure how I would "remove" those cleanups and apply them onto this branch. So, can we move forward with #403 since it's time-sensitive, and start a conversation about a PR cleanup policy in the future? I am open to keeping multiple commits for big changes.

This is exactly why I argue against squashing here. In any case, I merged the cleanup code back in and tested it with the examples in the description. Since there's no difference between #403 and this PR, I can start reviewing in here instead.

penrose-web/src/engine/EngineUtils.ts
```diff
@@ -38,7 +38,20 @@ function mapFloat<T, S>(f: (arg: T) => S, v: IFloatV<T>): IFloatV<S> {
...
 }
 
-function mapPt<T, S>(f: (arg: T) => S, v: IPtV<T>): IPtV<S> {
+function mapInt<T, S>(
+  f: (arg: T) => S,
+  v: IIntV<T>
+): IFloatV<S> {
+  return {
+    tag: "FloatV",
+    contents: f(v.contents as unknown as T) as S // Hack because we know that Ints only contain `number`s
```
wodeni, line 47: Did TS throw an error for this? Surprised that it couldn't figure out `IntV` and `FloatV` have the same types

penrose-web/src/engine/EngineUtils.ts
```diff
@@ -175,25 +214,28 @@ function mapPalette<T, S>(f: (arg: T) => S, v: IPaletteV<T>): IPaletteV<S> {
...
-    "ColorV",
-    "PaletteV",
-    "FileV",
-    "StyleV",
-  ];
+export function mapValueNumeric<T, S>(
+  f: (arg: T) => S,
+  v: Value<T>
+): Value<S> {
+  const nonnumericValueTypes = ["BoolV", "StrV", "ColorV", "PaletteV", "FileV", "StyleV"];
```
wodeni, line 221: Maybe we can make this a type? I bet this will be useful for the evaluator/compiler when it's trying to find (non)numeric types.

hypotext:
Ok, feel free to start reviewing here, but I need to check how easy it is to squash my old commits on this branch before I commit to using this branch or the other one. (Gotta give a talk until 4:30p, will be back after)

penrose-web/src/engine/Evaluator.ts
```diff
@@ -206,11 +196,13 @@ export const evalExprs = (
...
 ): ArgVal<VarAD>[] => es.map((e) => evalExpr(e, trans, varyingVars, optDebugInfo));
 
-function toFloatVal<T>(a: ArgVal<T>): T {
+function toFloatVal<VarAD>(a: ArgVal<VarAD>): VarAD {
   if (a.tag === "Val") {
     const res = a.contents;
     if (res.tag === "FloatV") {
       return res.contents;
+    } else if (res.tag === "IntV") {
+      return (constOf(res.contents) as unknown) as VarAD; // Not sure why TS compiler has problem here
```
wodeni, line 205: Umm it's not recognizing `res.contents` has to be `number` here?

penrose-web/src/engine/Evaluator.ts
```diff
@@ -221,6 +213,21 @@ function toFloatVal<T>(a: ArgVal<T>): T {
   }
 }
 
+function toVecVal<T>(a: ArgVal<T>): T[] {
+  if (a.tag === "Val") {
+    const res = a.contents;
+    if (res.tag === "VectorV") {
+      return res.contents;
+    } else {
+      console.log("res", res);
```
wodeni, line 222: Will need to clean up these logs soon

penrose-web/src/engine/Evaluator.ts
```diff
@@ -252,44 +260,23 @@ export const evalExpr = (
 
     case "AFloat": {
       if (e.contents.tag === "Vary") {
+        console.error("expr", e, "trans", trans, "varyingVars", varyingVars);
         throw new Error("encountered an unsubstituted varying value");
```
wodeni, line 264: Seems like we are throwing lots of errors everywhere. Would be good to have a module just for error msgs and just call something like `throw UnsubstitutedVaryingError()` here.

penrose-web/src/engine/Evaluator.ts
```diff
@@ -252,44 +260,23 @@ export const evalExpr = (
...
         const val = e.contents.contents;
 
+        // Don't convert to VarAD if it's already been converted
         return {
           tag: "Val",
           // Fixed number is stored in translation as number, made differentiable when encountered
           contents: {
             tag: "FloatV",
-            contents: constOf(val),
+            contents: val.tag ? val : constOf(val),
```
wodeni, line 274: Can we encode "already converted floats" as a type?

penrose-web/src/engine/Evaluator.ts
```diff
@@ -332,19 +323,141 @@ export const evalExpr = (
     }
     case "List": {
       const argVals = evalExprs(e.contents, trans, varyingVars, optDebugInfo);
+
+      // The below code makes a type assumption about the whole list, based on the first elements
+      // Is there a better way to implement parametric lists in typescript?
```
wodeni, line 328: For now, one way is to figure out the element type in `Lists`s in Haskell land (i.e. style compiler when checking blocks) and encode them with this info. In general, the new TS parser should probably have a more specific type for lists.

penrose-web/src/engine/Evaluator.ts
```diff
@@ -355,15 +468,22 @@ export const evalExpr = (
           throw Error(`expected 1 argument to ${fnName}; got ${argExprs.length}`);
         }
 
-        if (argExprs[0].tag !== "EPath") {
-          throw Error(`expected 1 path as argument to ${fnName}; got ${argExprs[0].tag}`);
```
wodeni, line 359: This is a very special case for these two functions. Maybe separate the body out to a function?

penrose-web/src/engine/Evaluator.ts
```diff
@@ -648,18 +873,80 @@ export const insertExpr = (
...
-    case "FieldPath":
+    case "FieldPath": {
       [name, field] = path.contents;
       // NOTE: this will overwrite existing expressions
       trans.trMap[name.contents][field] = { tag: "FExpr", contents: expr };
       return trans;
-    case "PropertyPath":
+    }
+    case "PropertyPath": {
       // TODO: why do I need to typecast this path? Maybe arrays are not checked properly in TS?
```
wodeni, line 883: And the same thing is not happening in the previous case?

penrose-web/src/engine/Evaluator.ts
```diff
@@ -648,18 +873,80 @@ export const insertExpr = (
...
+            res3[indices[0]] = floatValToExpr(expr.contents);
+            return trans;
+          } else if (res.tag === "Done") { // Deal with vector values
+            const res2 = res.contents;
+            if (res2.tag !== "VectorV") {
+              throw Error("expected Vector");
+            }
+            const res3 = res2.contents;
+            res3[indices[0]] = expr.contents.contents; // TODO: Does this actually leave a gradient trail, though?
+            // COMBAK: Why does the metadata contain extraneous elements?
```
wodeni, line 940: Is this resolved?

penrose-web/src/engine/EngineUtils.ts
hypotext: Thanks for pointing this out, I decided to get rid of it as the code currently doesn't aim to do int handling. (In general TS throws an error because `Int` always contains `number`, but the conversion function is general from `T` to `S`, and it can' t unify `Int` with `T`, which makes sense. I don't think here is the right place to handle Int conversion anyway.)

penrose-web/src/engine/Evaluator.ts
hypotext: Yeah, I only left in `console.log` right before an error is thrown, for debugging.

penrose-web/src/engine/EngineUtils.ts
hypotext: Sure, I can do that. But how would I use that type as in L239 in `EngineUtils` (when I need to check if a tag is a member of that set of types)? Not sure how to do that in TypeScript.

penrose-web/src/engine/Evaluator.ts
hypotext: Oh, I fixed it. L199 had a fake `<VarAD>` type parameter here which was causing a super weird error message. Removing that fixes it.

penrose-web/src/engine/Evaluator.ts
hypotext: Sure, I'm happy to use those kinds of error messages in Evaluator if you want to set up an error message module (or send an example of how to do it).

penrose-web/src/engine/Evaluator.ts
hypotext: I thought that's what this line is doing -- if val is tagged (i.e. is `VarAD`), then don't do anything (because it's an already converted float), otherwise convert it (because it has type `number`).

penrose-web/src/engine/Evaluator.ts
hypotext: Oh yeah, encoding the expected type of a (singly-typed) list in the backend would be helpful in the frontend. But once we move to Typescript then we should figure out how to model lists properly.

penrose-web/src/engine/Evaluator.ts
hypotext: I don't think that was my TODO, but anyway, I removed the type annotations and it works fine.

penrose-web/src/engine/Evaluator.ts
hypotext: Yes

penrose-web/src/engine/Evaluator.ts
hypotext: I think it's okay to leave this inline since there would be no other callers to the function if I separated it out.

hypotext:
Thanks for the review @wodeni, I addressed your comments in a0338f9. Let me know what you think.

Here are my preferences, though, regarding main branch history:

1) I would like to squash and clean up my commits from this branch first, as I don't want my personal unsquashed commits from this branch to appear in the main branch's history.

    However, it doesn't seem to be possible to clean up just my commits on this branch, as when I run (say) `git rebase ff17027`—my first commit on this branch—it pops open a rebase window that includes all commits from master interleaved with this branch's. That makes it very hard to clean things up. If you can think of some other git-fu for me to clean up my commits without touching yours or the ones we pulled from main, I'm happy to try it.

2) I would like for only build-preserving commits to appear in the main branch's history. This is part of the squash and rebase philosophy outlined [here](https://blog.carbonfive.com/always-squash-and-rebase-your-git-commits/):

    > If you follow this process it guarantees that ALL commits in master build and pass tests. This simple fact makes debugging an issue much easier. You can use git bisect when trying to find the source of a bug. Git bisect becomes almost completely ineffective  if there are broken commits on the master branch; if you jump to a commit that isn’t clean, it’s difficult or impossible to tell if it introduced the bug.
    >
    > The process also cleans up your git log. [see inline pictures]
    >
    > Because each commit contains all the code for a feature or bug fix, you can easily see the whole unit of work in a commit, and you don’t have to  worry about checking out the correct commit from a branch.
    >
    > My favorite outcome of this workflow is that it makes handling conflicts from rebasing simple. Since you’ve squashed down to one commit, you only have to deal with those conflicts once, rather than having to work against half-baked code. It reduces the risk of losing code when dealing with the conflicts.

    Regarding your suggestion to use `git tag`, I think it is generally used for version/release numbers, not marking every buildable version of the repo, no? Anyway, it seems better for CI if we don't push any build-breaking / test-failing commits, and that seems like a reasonable policy that many other big projects follow.

3) Otherwise, I'm fine with other folks having unsquashed commits in the main branch as long as they are reasonably clean and not build-breaking. 

    So, how can we clean up this branch to make 1) and 2) happen? My solution would still be to pull my last commit onto the `style-1.01` branch. I'm totally open to other solutions as long as they satisfy 1) and 2). What do you propose @wodeni? We can also talk about it on Monday.

wodeni:
> 1. I would like to squash and clean up my commits from this branch first, as I don't want my personal unsquashed commits from this branch to appear in the main branch's history.
>    However, it doesn't seem to be possible to clean up just my commits on this branch, as when I run (say) `git rebase ff17027`—my first commit on this branch—it pops open a rebase window that includes all commits from the main branch interleaved with this branch's, since I'd run `git merge` earlier. That makes it very hard to clean things up. If you can think of some other git-fu for me to clean up my commits without touching yours or the ones we pulled from main, I'm happy to try it. (Or maybe the solution is to rebase onto the main branch instead of merging, in the future.)

Right, the earlier merge with `master` probably made the rebase very hard, and I don't recommend spending more time to squash a few commits earlier on this branch. In general, I think the lesson learned here is we can't have just one base branch (i.e. `master`) when multiple people are working on a big feature. In this case, the early experimentation should happen on a separate branch off from `new-style` and squashed individually within that branch. Maybe it's better to mirror some common branching strategies like [this one](https://nvie.com/posts/a-successful-git-branching-model/). 

> 2. I would like for only build-preserving commits to appear in the main branch's history. This is part of the squash and rebase philosophy outlined [here](https://blog.carbonfive.com/always-squash-and-rebase-your-git-commits/):

People have different philosophies depending on where you look ([here](https://www.felixmoessbauer.com/blog-reader/why-git-squash-merges-are-bad.html) and [here](https://sdtimes.com/bugs/guest-view-hidden-hazards-squashing-git-commits/)). I tend to agree with the following: 
> Part of what makes squashing a poor default practice is that it’s somewhat at odds with other things Git tends to encourage. Git as a VCS excels at helping engineers move quickly: branching is cheap, committing frequently is encouraged, and there are lots of great Git power-user features that make it very easy to recover.
> So while there’s nothing inherently bad about squashing commits, and tactical squashing is a valuable thing, it’s important to remember that squashing is an inherently destructive act—one that removes development breakpoints. If overused, squashing can significantly increase the cost of finding and fixing flaws when things go awry.

>    Regarding your suggestion to use `git tag`, I think it is generally used for version/release numbers, not marking every buildable version of the repo, no? Anyway, it seems better for CI if we don't push any build-breaking / test-failing commits, and that seems like a reasonable policy that many other big projects follow.

I was addressing the need to `bisect` and find out bug causes. `git tag` is a good complement to a more granular commit history. I believe having both in the future will be helpful, instead of very big commits. 

> 3. Otherwise, I'm fine with other folks having unsquashed commits in the main branch as long as they are reasonably clean and not build-breaking.

Agreed! Also wanted to point out that we don't have a good test suite in place, so the notion of "build-breaking" is not exactly the same as "CI-breaking". In the future, I'd suggest [pre-commit hooks for testing](https://typicode.github.io/husky/#/) to encourage more meaningful pushed commits. Again, I think a single squashed commit from a separate branch that's not the feature branch is absolutely fine, and we should be more "tactical" about squashing next time :D. 

>    So, how can we clean up this branch to make 1) and 2) happen? My solution would still be to pull my last commit onto the `style-1.01` branch. I'm totally open to other solutions as long as they satisfy 1) and 2). What do you propose @wodeni? We can also talk about it on Monday.

Since `rebase` is not an effcient solution, I agree with your solution for this branch. Hopefully this discussion will be a good source when we start writing our community guideline doc.

penrose-web/src/engine/EngineUtils.ts
wodeni: Well in this case we should probably change `IIntV` to a non-generic type instead

penrose-web/src/engine/EngineUtils.ts
wodeni: How about [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)?

penrose-web/src/engine/EngineUtils.ts
hypotext: Done in 02977b327e2e89d8ede9c0f5ecef51d71e6ff438

penrose-web/src/engine/EngineUtils.ts
hypotext: Let's hold off on doing this until there's a second use case for `nonnumericValueTypes`.

hypotext:
Going to move the last two commits to `style-1.01` and merge there, as we discussed. Thanks!

Issue #301, opened 2019-11-27
Computing label dimensions reliably


wodeni:
Related: #299 

To generate figures in headless mode, we still need to know the dimensions of the labels at runtime, since we delegate label rendering to MathJax. The current approach gets SVG DOM elements of labels, call `getBbox()` on a ghost `<g>` element that wraps each of them, and return the width and height of the `SVGRect` returned. @maxkrieger and I did some experiments on the label generation & dimensions of the labels. Here's what we know so far:

* The MathJax SVG labels has path strings that is in some random model space, not the screen space, where the y-axis is facing "up". See an example below:
```svg
<svg xmlns:xlink="http://www.w3.org/1999/xlink" width="1.764ex" height="2.041ex" viewBox="0 -773.9 759.5 878.8" role="img" focusable="false" xmlns="http://www.w3.org/2000/svg" style="font-size: 19pt">
  <g stroke="none" fill="#000000" stroke-width="0" transform="matrix(1 0 0 -1 0 0)">
    <path stroke-width="1" d="M231 637Q204 637 199 638T194 649Q194 676 205 682Q206 683 335 683Q594 683 608 681Q671 671 713 636T756 544Q756 480 698 429T565 360L555 357Q619 348 660 311T702 219Q702 146 630 78T453 1Q446 0 242 0Q42 0 39 2Q35 5 35 10Q35 17 37 24Q42 43 47 45Q51 46 62 46H68Q95 46 128 49Q142 52 147 61Q150 65 219 339T288 628Q288 635 231 637ZM649 544Q649 574 634 600T585 634Q578 636 493 637Q473 637 451 637T416 636H403Q388 635 384 626Q382 622 352 506Q352 503 351 500L320 374H401Q482 374 494 376Q554 386 601 434T649 544ZM595 229Q595 273 572 302T512 336Q506 337 429 337Q311 337 310 336Q310 334 293 263T258 122L240 52Q240 48 252 48T333 46Q422 46 429 47Q491 54 543 105T595 229Z">
    </path>
  </g>
</svg>
```
* To get the labels into the screen space, MathJax does some mysterious computation to compute width and height of the labels in `ex` unit. Which matchs of ratio of the `viewBox` exactly. In the previous example:
`2.041ex / 1.764ex = 1.1570294785`
`878.8 / 759.5 = 1.1570770244`
[...]

PR #421, opened 2020-12-08
Caching labels in the frontend

wodeni:
# Description

Related issues: #301 

In the frontend, we used to pre-render labels, compute their dimensions, and store them directly in `Shape` as additional properties like `shape.w.updated`. It was a hacky solution that prevented us from typing the shapes properly and was very error-prone. 

Instead of injecting values into `Shape` directly, these front-end computed values are stored in (temporarily named) `LabelCache`. Each entry in the cache is a `LabelData` defined as:
```ts
interface LabelData {
  w: Value<number>;
  h: Value<number>;
  rendered: HTMLElement;
}
```

Tested with `venn-3d` and `twoVectorPerp`.

# Checklist

# Open questions

* The current implementation only works with `pendingPaths` of `Text` GPIs, _not `Image`s_. It's easy to generalize. Are there any other use case we can foresee beyond `Text` and `Image`?
* I haven't completely resolved the lifecycle of `LabelCache` and `pendingPaths`. How often do they need to be updated?


wodeni:
@maxkrieger lmk if this seems to be a good approach. Happy to generalize it to work with `Image`, iterate on the impl etc if you're on board with caching labels separately.

maxkrieger:
Code looks good. Only bit I don't understand is how `w` and `h` work if theyre potentially optimized values - shouldnt they be built into the shape properties? Are they?

maxkrieger (approved):

wodeni:
`loadImages` only seems to be used for `ImageTransform` specifically. It's unclear to me if it's still needed. I'll merge this PR now and we can revisit this if there's need.

PR #419, opened 2020-12-03
Porting the compilation pipeline

wodeni:
# Description

⚠️ __DNR: Do Not Rebase__ ⚠️ 

Related issues: #336 #428  

We will port the Domain, Substance Style parsers/checkers and compiler into TypeScript.

# Implementation strategy and design decisions

* AST types are converted to TS. See `penrose-web/src/compiler/Style.ts` for a possible place to put the Style compiler code. The top-level function should be `compileStyle`, and its signature looks like something like this: 
```ts
export const compileStyle = (
  env: VarEnv,
  subAST: SubProg,
  styAST: StyProg
) /* :State */ => {
  // TODO: fill in the style compiler here
};

``` 
# Examples with steps to reproduce them

TODO

# Checklist

TODO: update checklist for TS

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass locally using `npm test`
- [ ] I ran `npm run docs` and there were no errors when generating the HTML site
- [ ] My code follows the style guidelines of this project (e.g.: no ESLint warnings)

# Open questions

Questions that require more discussion or to be addressed in future development:

PR #446, opened 2021-01-14
Port Style compiler to typescript

hypotext:
# Description

Related issue/PR: #419 #441

This PR ports the Style compiler to typescript. All past examples work, with the exception of very minor issues listed below.

This is basically a 1:1 port of the Style compiler from Haskell, with some minor changes made to accommodate the fact that we did the Evaluator in ts first (e.g. use of `insertExpr` in Style compiler) and minor changes in the grammar, as well as minor changes in the functionality of the Domain/Substance toolchain. Errors are dealt with idiosyncratically, usually marked with `TODO(errors)`.

## Done

- [x] Port previous/new compiler features
    - [x] Selector typechecking (relies on Substance typechecking, @wodeni)
    - [x] Subtypes
    - [x] Path aliasing
    - [x] Disambiguating `SEFuncOrValsCons`
    - [x] Adding labels and names to translation
- [x] Add some tests
- [x] Integrate with system
    - [x] Repro all existing examples with new toolchain (with test suite)
    - [x] Requires Evaluator to work with new grammar
    - [x] Address important `TODO`s/`COMBAK`s
- [x] Integrate with new Domain/Substance parsers (w/ @wodeni)
- [x] Document compiler (moving comments from Haskell functions)
- [x] Integrate with App + CLI

## Changes to be made in the near future

- [x] Layering (@wodeni)
- [ ] Fix minor label/feature problems, documented in #460 #461 #462 (@hypotext)
- [ ] Ad-hoc unit testing of negative examples (e.g. testing that invalid programs have the correct errors reported) and specific features in isolation (e.g. path aliasing)
- [ ] Deprecating StyleTestData? (@hypotext)
- [ ] Break up `Style` into modules?
- [ ] Move comments into docstring format?
- [ ] Overhaul error and warning handling (w/ @wodeni)

All examples from the previous compiler, on this [page](https://github.com/penrose/penrose/wiki/Example-diagrams), work.

For more documentation, see `compiler/README.md`.

# Checklist

- [x] New and existing tests pass locally using `npm test`
- [x] I ran `npm run docs` and there were no errors when generating the HTML site

# Open questions

- [ ] Pseudorandomness
- [ ] Int/Float disambiguation
- [ ] Syntactic sugar
- [ ] Plugins??


hypotext:
Style compiler now generates the right images (with optimization and resample):

image
image

hypotext:
@wodeni My next step is to pull from `style-compiler` and integrate the compiler with the new parsers. Questions:

* What is the recommended workflow for using the new parsers/checkers to generate the inputs to the compiler? (i.e. what command replaces the `runpenrose your.dsl your.sub your.sty` command? Or, how do I specify the program inputs?) 
* Where do the new parsers/checkers output their inputs in the web runtime? Currently the compiler runs in `Canvas.tsx`'s `processData` function -- what is the right place to move it in the new workflow?

(I wrote the layering function you requested btw, see `computeShapeOrdering` and `topSortLayering`)

wodeni:
@hypotext ate you done with all of the items in this PR? If so, I think it’s a good time to do some cleanup, maybe add some tests, and then merge it first. In the meantime, I can take care of the UX for loading files and integration with the rest of the system on another branch.

penrose-web/src/compiler/Style.ts
```diff
@@ -1763,31 +1772,37 @@ const genOptProblemAndState = (trans: Translation): State => {
...
-  const subOut: SubOut = info[3];
+  const domainProgRes: Result<Env, PenroseError> = compileDomain(domainStr);
+  const env0: Env = unsafelyUnwrap(domainProgRes);
 
-  const subProg: SubProg = subOut[0];
-  const varEnv: VarEnv = subOut[1][0];
-  const subEnv: SubEnv = subOut[1][1];
-  // TODO: Bring back `eqEnv`?
-  const labelMap: LabelMap = subOut[2];
+  // TODO: Could be more efficient if compileSubstance also outputs parsed Sub program
```
wodeni, line 1790: Didn't know `compileStyle` will be using the Substance AST. Will it be okay if I add the AST as a key in `SubstanceEnv`, instead of adding yet another member to the list of return values?

wodeni: Side note: you'll probably get to this when you are closer to the finish line. Remember to stop using `unsafelyUnwrap`. As the name suggests, it's a function for debugging. You can do an `if` check by calling `isOk()` on the result.

hypotext: > Didn't know compileStyle will be using the Substance AST. 

Yes, it's needed to match selectors against the Substance program.

> Will it be okay if I add the AST as a key in SubstanceEnv, instead of adding yet another member to the list of return values?

Yes, that's fine.

> Remember to stop using unsafelyUnwrap. As the name suggests, it's a function for debugging. You can do an if check by calling isOk() on the result or use functions like andThen. I can help you refactor this after you are done with other things.

It looks like this `Result` monad is going to require using it and its helper functions everywhere in the Style compiler, right? To test things, I just need the value out of the `Result`; the rest of it seems to be related to error handling so I'm going to leave this for later. It would be helpful if you can address the `Result` refactor when we get to error handling in the Style compiler.

wodeni: > It looks like this `Result` monad is going to require using it and its helper functions everywhere in the Style compiler, right? To test things, I just need the value out of the `Result`; the rest of it seems to be related to error handling so I'm going to leave this for later. It would be helpful if you can address the `Result` refactor when we get to error handling in the Style compiler.

It shouldn't be that viral in the Style compiler. My plan is to only use `Result` in `compileStyle` and all other style errors will be handled as usual, just a giant catch in `compileStyle`. We can do that once this PR is merged.

hypotext: Ok, that sounds good.

hypotext:
@wodeni Let me know as soon as there is a more ergonomic way to load programs in the frontend than pasting each one into a JS string (which is what I'm currently doing in `TestSetTree.ts`, etc). That would make testing easier. 🙂

wodeni:
> @wodeni Let me know as soon as there is a more ergonomic way to load programs in the frontend than pasting each one into a JS string (which is what I'm currently doing in `TestSetTree.ts`, etc). That would make testing easier. 🙂

There already is one. See https://github.com/penrose/penrose/pull/451

hypotext:
Cool thanks, will try tomorrow!

hypotext:
New cli looks good. @wodeni I had another question about cli integration: how should I integrate the Style compiler with the new inputs?

I see the `files` appear in `App.tsx`'s `componentDidMount()`. Currently, the Style compiler is bootstrapped off of `Canvas`'s `processData` function. I'm not clear how to deprecate `processData` (since IDK how this all fits with the React app lifecycle, etc). Do I move all of `processData`'s contents (including `compileStyle`, label generation, etc.) into `App`'s `componentDidMount`?

wodeni:
Yeah that sounds about right.

You can more or less follow the pattern in functions like `resample` in App.tsx. Basically after compiling Style and generating the initial state, you’ll need to set `data` in the component’s state and maybe call `onCanvasState`.

hypotext:
Thanks! I'll try that now.

Do I need to set this `processedInitial` flag that is set to `false` in `resample`? In the definition, I see that there is a comment that reads `processedInitial: false, // TODO: clarify the semantics of this flag` 😛

wodeni:
Lol. I’m inclined to say you can leave as false, as onCanvasState will set it to true once it’s called. Yeah the semantics aren’t clear, this was for the back and forth after label generation, which is probably still relevant.

@maxkrieger refactoring the renderer should be in operation bento box.

maxkrieger:
The renderer is a nightmare. Processed initial, my guess, is only there for the colored indicator light, so you know when the labels have been generated (and possibly the other computational graph state transitions @wodeni and I did in summer 2019).

hypotext:
@maxkrieger I see there are two places where `processedInitial` is set to `true`, which seems like the flag has a meaning.

wodeni:
> @maxkrieger I see there are two places where `processedInitial` is set to `true`, which seems like the flag has a meaning.

@maxkrieger is probably right. Basically the flag is not gonna affect the behavior of the component, so feel free to experiment with either. Should get the job done regardless

penrose-web/src/compiler/Style.ts
```diff
@@ -1,7 +1,2068 @@
...
+  if (t.tag === "OptEval") {
+    if (t.contents.tag === "Vary") {
+      return true;
+    }
+  }
+
+  return false;
+};
+
+const mkPath = (strs: string[]): Path => {
```
wodeni, line 1442: Instead of constructing dummy nodes, consider creating a hash or just use a concatenated string (if they are guaranteed to be unique) of the paths. The AST node types are really meant to be _just for the parser/checker phase_. These big `ASTNode`s are probably not great for lookups and are prone to errors.

penrose-web/src/engine/EngineUtils.ts
```diff
@@ -306,3 +314,391 @@ export const makeTranslationDifferentiable = (trans: any): Translation => {
...
+          const ip = innerPath as IPropertyPath;
+          // a.x.y[0] = e
+          [name, field, prop] = [ip.name, ip.field, ip.property];
+          const gpi = trans.trMap[name.contents.value][field.value] as IFGPI<VarAD>;
+          const [, properties] = gpi.contents;
+          const res = properties[prop.value];
+
+          if (res.tag === "OptEval") {
+            // Deal with vector expresions
+            const res2 = res.contents;
```
wodeni, line 522: Check tag first and destructure next. Probably more descriptive names + type annotations will make the code more readable.

penrose-web/src/engine/EngineUtils.ts
```diff
@@ -306,3 +314,391 @@ export const makeTranslationDifferentiable = (trans: any): Translation => {
...
+        trans.trMap[name.contents.value] = {};
+      }
+
+      const fieldRes: FieldExpr<IVarAD> = trans.trMap[name.contents.value][field.value];
+
+      if (fieldRes.tag === "FExpr") {
+        // Deal with GPI aliasing (i.e. only happens if a GPI is aliased to another, and some operation is performed on the aliased GPI's property, it happens to the original)
+        // TODO: Test this
+        if (fieldRes.contents.tag === "OptEval") {
+          if (fieldRes.contents.contents.tag === "FieldPath") {
```
wodeni, line 429: Do you think the types are nested too much? If so, we can refactor them into simpler ones wherever applicable.

penrose-web/src/types/types.d.ts
```diff
@@ -1427,6 +1543,72 @@ interface IProp {
...
+// Style static semantics for selectors
+
+// Whether a variable in a selector is a Style variable or a Substance variable -- NOTE: This is new from the semantics because we can only store string in sTypeVarMap. So you have to look up the ProgType here too
+// const enum ProgType {
+//   Sub,
+//   Sty
+// }
+// TODO why doesn't this work
+
+type ProgType =
```
wodeni, line 1557: Just `type ProgType = "SubProg" | "StyProg"` would be just fine.

penrose-web/src/types/types.d.ts
```diff
@@ -1447,8 +1629,50 @@ interface ASTNode {
...
+}
+
+//#endregion
+
+//#region
+
+type StyErrors = string[];
+// TODO: Convert this to StyleError[]
+
+interface Left<A> {
```
wodeni, line 1643: Use true-myth's either (i.e. `Result`) instead? Doesn't really matter if `Either` is only internal to the Style compiler. If so, it'll make more sense for them to be defined in the module itself?

penrose-web/src/ui/Canvas.tsx
```diff
@@ -45,34 +59,78 @@ class Canvas extends React.Component<ICanvasProps> {
    * @memberof Canvas
    */
   public static processData = async (data: any) => {
-    const state: State = decodeState(data);
+    // COMBAK: Clean up this use
+    const res: Either<StyErrors, State> = compileStyle();
```
wodeni: We'll definitely be using `Result` for this I think. I'll get to it after the merge.

penrose-web/src/utils/OtherUtils.ts
```diff
@@ -1,11 +1,14 @@
-// @ts-nocheck
-// TODO: COMBAK: HACK: remove this directive to re-enable tsc. Temporarily turned off to accommondate new AST types
-
 import * as _ from "lodash";
 
 const RAND_RANGE = 100;
 const TOL = 1e-3;
 
+// HACK: Copied from EngineUtils
```
wodeni, line 6: Duplicates in `EngineUtils`?

penrose-web/src/utils/Util.tsx
```diff
@@ -299,14 +299,14 @@ export const hsvToRGB = (
...
     : h < 120
-    ? hsv2rgb(x, c, 0, m)
-    : h < 180
-    ? hsv2rgb(0, c, x, m)
-    : h < 240
-    ? hsv2rgb(0, x, c, m)
-    : h < 300
-    ? hsv2rgb(x, 0, c, m)
-    : hsv2rgb(c, 0, x, m);
+      ? hsv2rgb(x, c, 0, m)
```
wodeni, line 302: BTW what formatter are you running? We should probably all use prettier. I'm adding a config in the next PR.

penrose-web/src/compiler/Style.ts
```diff
@@ -1,7 +1,2068 @@
...
+const LABEL_FIELD = "label";
+
+const UnknownTagError = new Error("unknown tag");
+
+//#endregion
+
+//#region utils
+
+// numbers from 0 to r-1 w/ increment of 1
+const numbers = (r: number): number[] => {
```
wodeni, line 55: Lodash has a function for this: https://lodash.com/docs/4.17.15#range

penrose-web/src/compiler/Style.ts
```diff
@@ -1,7 +1,2068 @@
...
+  const arr = [];
+  for (let i = l; i < r; i++) {
+    arr.push(i);
+  }
+  return arr;
+}
+
+export function numbered<A>(xs: A[]): [A, number][] {
+  if (!xs) throw Error("fail");
+  return _.zip(xs, numbers(xs.length)) as [A, number][]; // COMBAK: Don't know why typescript has problem with this
```
wodeni, line 67: Because `zip` defaults to `undefined` if a list is longer, which is different from Haskell. Try numbering things with just `map` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map). It gives you the index too if all you want is the sequential numbering of items.

penrose-web/src/compiler/Style.ts
```diff
@@ -1,7 +1,2068 @@
...
+    // G |- B : T1
+    const res1 = checkVar(rel.id.contents, varEnv);
+
+    // TODO (error)
+    if (isErr(res1)) {
+      // TODO(error): extract and return error -- how to convert from SubstanceError to StyError?
+      return ["substance typecheck error in B"];
+    }
+
+    const [vtype, env1] = unsafelyUnwrap(res1);
```
wodeni, line 303: The right way is to do this in the `else`. TS will infer the types if you call `res1.value`

penrose-web/src/compiler/Style.ts
```diff
@@ -1,7 +1,2068 @@
...
+  // All keys already need to be unique in js, so only checking values
+  const vals = Object.values(subst);
+  const valsSet = {};
+
+  for (let i = 0; i < vals.length; i++) {
+    valsSet[vals[i]] = 0; // This 0 means nothing, we just want to create a set of values
+  }
+
+  // All entries were unique if length didn't change (ie the nub didn't change)
+  return Object.keys(valsSet).length === vals.length;
```
wodeni, line 450: Not sure what this function is doing, just looking for unique keys? If so, use `uniq` (https://lodash.com/docs/4.17.15#uniq)?

penrose-web/src/compiler/Style.ts
```diff
@@ -1,7 +1,2068 @@
...
+    // Variable in backticks in block or selector (e.g. `X`), so nothing to substitute
+    return bform;
+  } else if (bform.tag === "StyVar") {
+    // Look up the substitution for the Style variable and return a Substance variable
+    // Returns result of mapping if it exists (y -> x)
+    const res = subst[bform.contents.value];
+
+    if (res) {
+      return {
+        ...bform, // Copy the start/end loc of the original Style variable, since we don't have Substance parse info (COMBAK)
```
wodeni, line 477: Again, the `ASTNode` type is not really designed for this. You could do a postprocessing step to throw away all node-specific data.

penrose-web/src/compiler/Style.ts
```diff
@@ -1,7 +1,2068 @@
...
+
+const argsEq = (a1: SubPredArg, a2: SubPredArg): boolean => {
+  if (a1.tag === "ApplyPredicate" && a2.tag === "ApplyPredicate") {
+    return subFnsEq(a1, a2);
+  } else if (a1.tag === a2.tag) { // both are SubExpr, which are not explicitly tagged
+    return subExprsEq(a1 as SubExpr, a2 as SubExpr);
+  } else return false; // they are different types
+};
+
+const subFnsEq = (p1: any, p2: any): boolean => {
```
wodeni, line 792: Why `any`? At least you'd know it's definitely an `ASTNode` (or something more specific)?

wodeni (changes requested):
See my comments below for details. This is a pretty huge module. In terms of code organization, maybe divide it into modules like `Selector`, `Block`, and `State` for clarity?

hypotext:
@wodeni Hey! This is ready for a first review. I updated the description on this PR to reflect the minor cleanup I plan to do. 

When you review, can you mark the changes you are requesting that are absolutely necessary for this branch, vs. those that are nice-to-have? Given our timeframe, I would prioritize doing the urgent ones first and getting this PR merged so the whole system can be tested. I can queue minor TODOs to be addressed in a more leisurely fashion afterward.

hypotext:
Also thanks for your prev comments! I'll plan to address them some of them now, unless you want me to wait on any of them?

wodeni:
> Also thanks for your prev comments! I'll plan to address them soon, unless you want me to wait on any of them?

Awesome! I’ll do a detailed pass tomorrow morning. My main issue is the overuse of ASTNode types, for which I suggested a few possible solutions above. My last pass was mostly code style stuff. I’ll try to understand the module better in the next pass. Meanwhile, please feel free to address the comments above, mostly importantly the ASTNode thing.

Also lmk if there are specific concerns you have, so I can pay close attention to them when I read the code again.

hypotext:
Cool. Off the top of my head, some concerns are:

- how to break up the Style module
- messiness (or possible inconsistencies) in path handling between AccessPath, VectorAccess, MatrixAccess

Will let you know if others come up, and feel free to ask any qs on slack, I know it's not the world's best documented code right now!

penrose-web/src/compiler/Style.ts
```diff
@@ -1,7 +1,2131 @@
...
+    constrFns,
+
+    // `params` are initialized properly by optimization; the only thing it needs is the weight (for the objective function synthesis)
+    params: {
+      optStatus: { tag: "NewIter" },
+      weight: initConstraintWeight,
+      lbfgsInfo: defaultLbfgsParams,
+      UOround: -1,
+      EPround: -1,
+    } as unknown as Params,
```
wodeni, line 2013: Minor: for now, you can use something like `const rng: prng = seedrandom(json.rng, { global: true });` for the generator. We can come back to this later.

penrose-web/src/compiler/Style.ts
```diff
@@ -1,7 +1,2131 @@
...
+  return trans;
+};
+
+// NOTE: This function mutates the translation
+// rule Line-delete
+const deletePath = (trans: Translation, path: Path): Either<StyErrors, Translation> => {
+  // TODO(errors): Maybe fix how warnings are reported? Right now they are put into the "translation"
+  if (path.tag === "FieldPath") {
+    const transWithWarnings = deleteField(trans, path.name, path.field);
+    return Right(transWithWarnings);
```
wodeni, line 1229: Minor: why `let`? Looks like `const` is better suited for this. BTW these are all caught by ESLint. You can start using it to spot lower-level issues like this.

wodeni (changes requested):
Done with my second pass. It does seem to be a good one-to-one port. Here are the main comments:

* Error reporting: there are 154 results for `error` in this module. They are (1) reported in inconsistent ways (thrown vs. returned) and (2) only returning a string without any location info. It would be helpful for our error reporting refactor later to know: (a) what types of error there are in Style and (2) for each type of error, what information it will need for reporting the error in any readable and traceable way. I think I'll need your help on both (a) and (b).
* Error testing: a reasonable next step of identifying the error types is to write some test cases that trigger these errors. 
* Module separation: like my earlier suggestion. Separating the modules along the `#region` line seems to be reasonable. Everything about generating state should probably be in a `State` module.
* Regarding `AccessPath`: I thought that was the only type we use for both vector and matrix access? Anyway, the way of encoding varying paths is a little messy. I still recommend hashing or finding another representation of paths for cleaner code and better performance.
* All requested changes are marked below as `Major` or `Minor`

hypotext:
Thanks for the speedy review @wodeni! 

I'll address comments in this order:

- [ ] Consistently throw all errors instead of returning them
- [ ] Start a separate list of kinds of errors and the information needed
- [ ] Add some test cases to trigger errors
- [ ] Split Style into modules
- [ ] Fix minor label bugs (might do this after merge depending on how long the prev items take)

Aiming to turn it around in a day or so.

Questions:

* For the error list, what information do you want me to include about each error in the Style compiler? Or is "error message" and "info needed to report error" enough?
* Re this comment: https://github.com/penrose/penrose/pull/446#discussion_r568679610 -- Where do you want me to catch the errors? In `compileStyle`? Also, is there a decent way to handle nested error handling, that will be closest to the `Result` approach you plan to ultimately take? (e.g. `compileStyle` calls a function that might throw error, which itself calls a function that might throw error, and so on -- what is the preferred way to address that?)
* I'm confused about what you mean about "the overuse of ASTNode types" and how you would like me to address that. The only relevant comments I see are these:

https://github.com/penrose/penrose/pull/446#discussion_r568190233

https://github.com/penrose/penrose/pull/446#discussion_r568191872

hypotext:
There are also some potential errors to handle in functions in `EngineUtils`, like `insertExpr` and `findExpr`. Some of these functions are shared by the Style compiler and the Evaluator/runtime, making their error-handling more tricky as they might fail at either stage. 

Is there a way you would like me to deal with these errors now? Otherwise we can discuss the approach more and then address it in the next round.

wodeni:
> Thanks for the speedy review @wodeni!

For reviewing large PRs, the [VSCode plugin](https://code.visualstudio.com/blogs/2018/09/10/introducing-github-pullrequests) is pretty useful for navigating among functions. Would recommend 😄 .

> I'll address comments in this order:
> * [ ]  Consistently throw all errors instead of returning them

This decision depends on your desired error reporting behavior: if you do want to catch 1+ errors in the compiler, you will need to accumulate them. It's reasonable to have a `StyleEnv` as a single data structure to pass around, just like in the Domain and Substance checker, and you can accumulate error there instead of dealing with `Either`s constantly.

> Aiming to turn it around in a day or so.

Awesome!

> * For the error list, what information do you want me to include about each error in the Style compiler? Or is "error message" and "info needed to report error" enough?

Ideally, you can append need error types to `error.d.ts`. My workflow is: (1) think about what the error message should be, (2) figure out what members go into this error type based on the error message, and (3) make sure all call sites of this error actually have the necessary members in scope. For instance, `DuplicateName` error:

```ts
// in errors.d.ts
interface DuplicateName {
  tag: "DuplicateName";
  name: Identifier;
  location: ASTNode;
  firstDefined: ASTNode;
}
// in `showError` of Error.ts
case "DuplicateName": {
  const { firstDefined, name, location } = error;
  return `Name ${name.value} (at ${loc(
    location
  )}) already exists, first declared at ${loc(firstDefined)}.`;
}
```

In this case, `DuplicateName` will need the duplicated id, the location of the statement where the error occurs, and the location where the node this first defined. As a result, whenever you construct this error, you will need all 3 pieces of information, which sometimes requires extra code written to accommodate it.

In this PR, I think a type definition, an example error messages, and extra members needed for the error message should be enough for a reasonable first step. I tried to go the extra mile for Substance and Domain (e.g. providing suggestions for, say, `TypeNotFound`), but given the size of this module, we can try to minimize actual changes to the surrounding code first and improve the error quality as we go.  

> * Re this comment: [#446 (comment)](https://github.com/penrose/penrose/pull/446#discussion_r568679610) -- Where do you want me to catch the errors? In `compileStyle`? Also, is there a decent way to handle nested error handling, that will be closest to the `Result` approach you plan to ultimately take? (e.g. `compileStyle` calls a function that might throw error, which itself calls a function that might throw error, and so on -- what is the preferred way to address that?)

Regardless of how errors are handled internally, `compileStyle` should catch or retrieve errors and return them in a `Result` type. Like I said in the comment, "nested error handling" (if you mean handling 1+ errors) is not possible with exceptions, since it throws the first thing it sees and stops execution immediately. I think the safest option is to actually collect them in a list instead.

> * I'm confused about what you mean about "the overuse of ASTNode types" and how you would like me to address that. The only relevant comments I see are these:
> [#446 (comment)](https://github.com/penrose/penrose/pull/446#discussion_r568190233)
> [#446 (comment)](https://github.com/penrose/penrose/pull/446#discussion_r568191872)

I was referring to the construction of these dummy nodes. I think it's a symptom of overusing AST node types for later stages of compilation (in this case, mostly in generating the translation). The source-related info is there because it's actually designed as an output type for parsers, not internal contexts like the translation. Ideally, I think we should eliminate all the dummy node construction code somehow. Possible solutions:
*  Simplify ASTs by postprocessing them: after parsing, you create a "cleaner" AST by traversing the tree and constructing nodes that are subsets of `ASTNode`. They can be [`Partial<ASTNode>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype) or other TS-supported higher-order types. 
* Create new types specifically for internal representation: separately create `type Node = ASTNode | SyntheticNode` and handle them differently (maybe useful for errors).

hypotext:
Thanks for the comments! 

Hmm, the approach you described will take longer than I thought, since it requires formally modeling and categorizing all error types, and requires big code changes to figure out + pass in the relevant info for error reporting. I don't think I can do that in a day. 

For this branch, I would prefer to return error strings from every function, which are collected and reported in `compileStyle` via the current `foldM` approach, as that is already an improvement on what `compileStyle` does in the old compiler. Later we can work on the more formal error reporting. How does that sound?

> Ideally, I think we should eliminate all the dummy node construction code somehow. Possible solutions:

I guess there are other design requirements we should discuss before picking a solution:

- compiler still needs to be able to report errors (maybe all the way into the evaluator/runtime?), so it can't discard the AST info entirely (eg via cleaning)
- ideally compiler can still construct + operate generically on things that are Paths, and so on, without additional code to check if something is a synthetic node or not

What do you think is the best TS solution for that?

> Also, any plan to address the following? I think all the Path construction code also suffer from the dummy node problem. They are also not suitable for indexing, like in varyingPaths.

For indexing, I can convert every path to a string using a custom function like `pathStr: Path -> string` (? forgot the exact name -- is there a Hoogle equivalent for typescript?) that discards the AST node info. The output is a string like `a.shape.r`.

wodeni:
> For this branch, I would prefer to return error strings from every function, which are collected and reported in `compileStyle` via the current `foldM` approach, as that is already an improvement on what `compileStyle` does in the old compiler. Later we can work on the more formal error reporting. How does that sound?

Sure, that'd be a good first step.

> * compiler still needs to be able to report errors (maybe all the way into the evaluator/runtime?), so it can't discard the AST info entirely (eg via cleaning)

Instead of discarding everything upfront, you could either keep a frozen copy of the AST or have more flexible types that will allow synthetic nodes. 

> * ideally compiler can still construct + operate generically on things that are Paths, and so on, without additional code to check if something is a synthetic node or not

Sure. Perhaps the simplest approach would be to actually create types for these synthetic cases (e.g. `Path` and `LocalVar`). These shouldn't even go into the AST itself, but belong to another branch in a sum type. Let me try to illustrate this:
```ts
type StyleNode = ASTNode | SyntheticNode;
type SyntheticNode = InternalLocalVar | InternalPath | ....
interface InternalPath {  // NOTE: _not_ extending `ASTNode`
  tag: "InternalPath", // NOTE: all internal paths should have `tag` for compatibility with `ASTNode`!
  // ... more data
}
```

> For indexing, I can convert every path to a string using a custom function like `pathStr: Path -> string` (? forgot the exact name -- is there a Hoogle equivalent for typescript?) that discards the AST node info. The output is a string like `a.shape.r`.

Yeah, sounds good to me. 

(Hoogle for TS: nothing super mature. I saw this: https://tsearch.io/)

wodeni (approved):
@hypotext and I agreed to address the error-related issues and `ASTNode` refactoring in a later PR. This one is good to go!

Issue #423, opened 2020-12-09
Reduce bundle size of `@penrose/core`


wodeni:
Our production build of `penrose-web` is quite big (see output for `npm run build` or `npm run build:bundle`). At some point, we will need to cut down our dependencies or look into other optimization options. The bundle size also affects our CI, as `terser` sometimes reports `Call retries were exceeded` even with increased memory.

wodeni:
`@penrose/linear-algebra` is quite large: https://bundlephobia.com/result?p=@penrose/linear-algebra@1.0.1
Overall size distribution here:
![image](https://user-images.githubusercontent.com/11740102/102419187-e3a03780-3fc4-11eb-8079-0997765ff963.png)

wodeni:
New stats coming from #465 
![image](https://user-images.githubusercontent.com/11740102/107861568-4c146500-6e0c-11eb-8ddb-196a89454a9d.png)

samestep:
Using [esbuild-visualizer](https://www.npmjs.com/package/esbuild-visualizer) on #1165:

<img width="1362" alt="Screen Shot 2022-12-09 at 11 25 13 AM" src="https://user-images.githubusercontent.com/8246041/206747138-1691f362-b97e-4d72-95b4-7a17f9e2acd3.png">

PR #455, opened 2021-02-01
Restructure Penrose packages + better tooling 🍱

wodeni:
# Description

Related issue/PR: #419 #423 

As we finish up the port to TS, we can leverage some existing tools for building, testing, and distributing JavaScript packages. The overall goal is the following:

* Separate the __core__ of Penrose (parsers, compilers, and runtime) into a standalone TS package called `penrose-core` (if we are using namespace, then `@penrose/core`). As a result, we will no longer use CRA as our dependency.
* The old renderer will also be another package `penrose-renderer`, which still uses CRA.
* Other tools we built over the years and future packages,  like the IDE, automator, and roger, will all be separate packages with `penrose-core` as a dependency. 

* [x] Deprecate all Haskell code and configs
* [x] Use [lerna](https://lerna.js.org/) to manage a monorepo of multiple packages (i.e. this repo will become a monorepo)
* [x] Add test data collection and uploading via CircleCI and [codecov](https://codecov.io/gh).
* [x] Single test command to test all packages: `npm test`
* [x] Auto semantic version bump (with conventional commit support), git tagging, and changelog generation  
* [x] Single build command to build all packages: currently `npm bootstrap`
  * [ ] `bootstrap` only installs dependencies and doesn't seem to build local dependencies, only links them. TODO: figure out a way to automatically build all local dependencies.
* [ ] Separate renderer code from core
* [ ] Cleanup renderer
* [ ] New build tool for `penrose-core`: possibly rollup
* [ ] Prettier support + ESLint overhaul
* [ ] Publish all packages on npm (TODO: as a single package? as multiple?)

* New CI pipeline: https://app.circleci.com/pipelines/github/penrose/penrose/1416/workflows/4d6f0436-8f06-49d5-9c59-7bf542b5d4e6/jobs/3295/steps
* Test coverage: 
  image (2)
  image (3)

# Checklist

- [ ] New and existing tests pass locally using `npm test`
- [ ] I ran `npm run docs` and there were no errors when generating the HTML site

# Open questions


maxkrieger:
> * if we are using namespace, then `@penrose/core`

I think we should indeed do this scheme. Seems more standard and looks nicer

PR #473, opened 2021-02-09
Improve Style compiler error-handling

hypotext:
# Description

Related issue/PR: #446 

This PR improves the Style compiler error-handling so that all existing errors are modeled and reported separately, and adds tests for most of them.

Note that this is a **first pass** at cleaning up the errors that **already exist** in the Style compiler; I anticipate that actually using the system will help us discover many more kinds of errors that should be reported, and in other parts of the system, especially at runtime.

The PR follows the strategy broadly described in https://github.com/penrose/penrose/pull/446#issuecomment-772064166. 

It extends `errors.d.ts` with error types and `showError` in `Error.ts` with error-printing code. Tests are in `Style.test.ts`. Runtime errors are in `insertExpr`.

Errors are roughly split up between misc errors, selector checking errors, errors in generating the translation (deleting or inserting paths), and runtime errors.

Run the test suite: `yarn run test -t Compiler --watchAll=true`

# Checklist

# Open questions

- [ ] There are a couple of modeled errors that aren't properly thrown because the system fails with a different error before then (see a couple of examples in `Style.test.ts`), so their tests are commented out. Ideally, these should be caught earlier, at the point of failure.
- [ ] Improve the amount of information reported by `showError`.
- [ ] Improve handling of warnings in `compileStyle`. First, they are accumulated as warnings in the `Translation` but maybe they should be fatal? (Unless there are some design problems with Style that make them unavoidable, as has happened in the past?) Second, all warnings are currently treated as errors anyway.
- [ ] Improve handling of multiple errors; currently Style compiler reports the tag of the first error, plus the messages for all of them.
- [ ] Suggest ways to improve errors.
- [ ] Handle runtime errors.

### DIFF
--- a/packages/penrose-core/src/compiler/Style.ts
+++ b/packages/penrose-core/src/compiler/Style.ts
@@ -38,6 +38,7 @@ import {
   genericStyleError,
   unsafelyUnwrap,
   unsafelyGetErr,
+  firstStyleError
 } from "utils/Error";
 import { randFloats } from "utils/Util";
 import { checkTypeConstructor, Env, isDeclaredSubtype } from "./Domain";
@@ -261,7 +262,7 @@ const addMapping = (
 };
 
 // add warning/error to end of existing errors in selector env
-const addErrSel = (selEnv: SelEnv, err: StyError): SelEnv => {
+const addErrSel = (selEnv: SelEnv, err: StyleError): SelEnv => {
   return {
     ...selEnv,
     errors: selEnv.errors.concat([err]),
@@ -342,7 +343,7 @@ const checkDeclPatternsAndMakeEnv = (
 
 // TODO: Test this function
 // Judgment 4. G |- |S_r ok
-const checkRelPattern = (varEnv: Env, rel: RelationPattern): StyErrors => {
+const checkRelPattern = (varEnv: Env, rel: RelationPattern): StyleErrors => {
   // rule Bind-Context
   if (rel.tag === "RelBind") {
     // TODO: use checkSubStmt here (and in paper)?
@@ -365,7 +366,7 @@ const checkRelPattern = (varEnv: Env, rel: RelationPattern): StyErrors => {
 
     // TODO(error)
     if (isErr(res2)) {
-      // TODO(error): extract and return error -- how to convert from SubstanceError to StyError?
+      // TODO(error): extract and return error -- how to convert from SubstanceError to StyleError?
       const subErr2: SubstanceError = unsafelyGetErr(res2);
       return [{ tag: "TaggedSubstanceError", error: subErr2 }];
       // return ["substance typecheck error in E"];
@@ -388,7 +389,7 @@ const checkRelPattern = (varEnv: Env, rel: RelationPattern): StyErrors => {
     // G |- Q : Prop
     const res = checkPredicate(toSubPred(rel), varEnv);
     if (isErr(res)) {
-      // TODO(error): extract and return error -- how to convert from SubstanceError to StyError?
+      // TODO(error): extract and return error -- how to convert from SubstanceError to StyleError?
       const subErr3: SubstanceError = unsafelyGetErr(res);
       return [{ tag: "TaggedSubstanceError", error: subErr3 }];
       // return ["substance typecheck error in Pred"];
@@ -400,10 +401,10 @@ const checkRelPattern = (varEnv: Env, rel: RelationPattern): StyErrors => {
 };
 
 // Judgment 5. G |- [|S_r] ok
-const checkRelPatterns = (varEnv: Env, rels: RelationPattern[]): StyErrors => {
+const checkRelPatterns = (varEnv: Env, rels: RelationPattern[]): StyleErrors => {
   return _.flatMap(
     rels,
-    (rel: RelationPattern): StyErrors => checkRelPattern(varEnv, rel)
+    (rel: RelationPattern): StyleErrors => checkRelPattern(varEnv, rel)
   );
 };
 
@@ -1559,7 +1560,7 @@ const deleteField = (
 const deletePath = (
   trans: Translation,
   path: Path
-): Either<StyErrors, Translation> => {
+): Either<StyleErrors, Translation> => {
   if (path.tag === "FieldPath") {
     const transWithWarnings = deleteField(trans, path.name, path.field);
     return Right(transWithWarnings);
@@ -1573,7 +1574,7 @@ const deletePath = (
     return Right(transWithWarnings);
   } else if (path.tag === "AccessPath") {
     // TODO(error)
-    const err: StyError = { tag: "DeletedVectorElemError", path };
+    const err: StyleError = { tag: "DeletedVectorElemError", path };
     return Left([err]);
   } else if (path.tag === "InternalLocalVar") {
     throw Error(
@@ -1588,7 +1589,7 @@ const addPath = (
   trans: Translation,
   path: Path,
   expr: TagExpr<VarAD>
-): Either<StyErrors, Translation> => {
+): Either<StyleErrors, Translation> => {
   // Extended `insertExpr` with an optional flag to deal with errors and warnings
   // `insertExpr` replaces the old .hs functions `addField` and `addProperty`
 
@@ -1604,7 +1605,7 @@ const addPath = (
 const translateLine = (
   trans: Translation,
   stmt: Stmt
-): Either<StyErrors, Translation> => {
+): Either<StyleErrors, Translation> => {
   if (stmt.tag === "PathAssign") {
     return addPath(false, trans, stmt.path, {
       tag: "OptEval",
@@ -1626,7 +1627,7 @@ const translateBlock = (
   blockWithNum: [Block, number],
   trans: Translation,
   substWithNum: [Subst, number]
-): Either<StyErrors, Translation> => {
+): Either<StyleErrors, Translation> => {
   const blockSubsted: Block = substituteBlock(substWithNum, blockWithNum, name);
   return foldM(blockSubsted.statements, translateLine, trans);
 };
@@ -1637,7 +1638,7 @@ const translateSubstsBlock = (
   trans: Translation,
   substsNum: [Subst, number][],
   blockWithNum: [Block, number]
-): Either<StyErrors, Translation> => {
+): Either<StyleErrors, Translation> => {
   return foldM(
     substsNum,
     (trans, substNum, i) =>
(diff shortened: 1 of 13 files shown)


packages/penrose-core/src/compiler/Style.ts
```diff
@@ -350,8 +354,10 @@ const checkRelPattern = (varEnv: Env, rel: RelationPattern): StyErrors => {
 
     // TODO(error)
     if (isErr(res1)) {
-      // TODO(error): extract and return error -- how to convert from SubstanceError to StyError?
-      return ["substance typecheck error in B"];
+      const subErr1: SubstanceError = unsafelyGetErr(res1);
```
wodeni, line 357: Use `res1.error` instead. The `if` check will make sure this is the right error type.

packages/penrose-core/src/compiler/Style.ts
```diff
@@ -350,8 +354,10 @@ const checkRelPattern = (varEnv: Env, rel: RelationPattern): StyErrors => {
...
+      // TODO(error): Do we need to wrap this error further, or is returning SubstanceError with no additional Style info ok?
+      // return ["substance typecheck error in B"];
+      return [{ tag: "TaggedSubstanceError", error: subErr1 }];
```
wodeni, line 360: Ideally, I think you'll need to add the source locations of the relevant selector clauses as well. I can make `checkVar` return a more specific type and you can add a wrapper around those types and customize the error messages further if needed.

packages/penrose-core/src/compiler/Style.ts
```diff
@@ -361,18 +367,21 @@ const checkRelPattern = (varEnv: Env, rel: RelationPattern): StyErrors => {
 
     // TODO(error)
     if (isErr(res2)) {
-      // TODO(error): extract and return error -- how to convert from SubstanceError to StyError?
-      return ["substance typecheck error in E"];
+      // TODO(error): extract and return error -- how to convert from SubstanceError to StyleError?
+      const subErr2: SubstanceError = unsafelyGetErr(res2);
```
wodeni, line 371: Again. you can use `res2.error` here. Looks like you are actually checking Style AST nodes transformed to Substance ones? If so, your source locs will be Style-specific? In general, the `ASTNode` type doesn't have file related information yet. Maybe we'll need it later.

packages/penrose-core/src/compiler/Style.ts
```diff
@@ -361,18 +367,21 @@ const checkRelPattern = (varEnv: Env, rel: RelationPattern): StyErrors => {
...
+      return [{ tag: "TaggedSubstanceError", error: subErr2 }];
+      // return ["substance typecheck error in E"];
     }
 
-    const [etype, env2] = unsafelyUnwrap(res1);
+    const [etype, env2] = unsafelyUnwrap(res2);
```
wodeni, line 376: Do an `else` and then `res2.value`

packages/penrose-core/src/compiler/Style.ts
```diff
@@ -381,9 +390,10 @@ const checkRelPattern = (varEnv: Env, rel: RelationPattern): StyErrors => {
     // G |- Q : Prop
     const res = checkPredicate(toSubPred(rel), varEnv);
     if (isErr(res)) {
-      // TODO(error): extract and return error -- how to convert from SubstanceError to StyError?
-      // return unsafelyUnwrapErr(res);
-      return ["substance typecheck error in Pred"];
+      // TODO(error): extract and return error -- how to convert from SubstanceError to StyleError?
+      const subErr3: SubstanceError = unsafelyGetErr(res);
```
wodeni, line 394: `res.error`

packages/penrose-core/src/compiler/Style.ts
```diff
@@ -1459,16 +1470,23 @@ const deleteProperty = (
 
   if (!fieldDict) {
     // TODO(errors / warnings): Should this be fatal?
-    const err = `Err: Sub obj '${nm}' has no fields; can't delete path '${pathStr}'`;
-    return addWarn(trans, err);
+    return addWarn(trans, {
+      tag: "DeletedPropWithNoSubObjError",
+      subObj: name,
+      path: { tag: "PropertyPath", name, field, property, } as Path
```
wodeni, line 1476: `path` not typechecking?

packages/penrose-core/src/compiler/Style.ts
```diff
@@ -1718,13 +1749,21 @@ const insertLabels = (trans: Translation, labels: LabelMap): Translation => {
...
+    if (!labelRes) {
+      // We skip here, to avoid putting spurious labels in for namespaces in the translation.
+      return [name, fieldDict];
+    }
+
+    let label;
+    if (labelRes.isJust()) {
+      label = labelRes.value;
+    } else {
+      // TODO: Distinguish between no label and empty label?
```
wodeni, line 1763: Is this related to #474?

packages/penrose-core/src/compiler/Style.ts
```diff
@@ -2665,20 +2705,26 @@ export const compileStyle = (
...
-  // TODO(error return) ^
 
   log.info("translation (before genOptProblem)", translateRes);
 
-  // Translation failed somewhere. // TODO: Check that errors are returned in a consistent order
+  // Translation failed somewhere
   if (translateRes.tag === "Left") {
-    // const err = `Could not compile. Error(s) in Style while generating translation`;
-    return err(genericStyleError(translateRes.contents));
+    return err(firstStyleError(translateRes.contents));
```
wodeni, line 2713: Not sure why you are reporting just the first error. Any reason? You could do `interface StyleErrorList { errors: StyleError[] }` as a branch of `StyleError`.

packages/penrose-core/src/engine/EngineUtils.ts
```diff
@@ -615,32 +642,32 @@ export const insertExpr = (
             // Deal with vector values
             const res2 = res.contents;
             if (res2.tag !== "VectorV") {
-              const err = "expected Vector";
               if (compiling) {
-                return addWarn(trans, err);
+                return addWarn(trans, runtimeValueTypeError(path, "Vector", res2.tag));
               }
+              const err = "expected Vector";
               throw Error(err);
```
wodeni, line 649: Did you miss this one?

packages/penrose-core/src/types/errors.d.ts
```diff
@@ -7,7 +7,8 @@
 type PenroseError =
   | (DomainError & { errorType: "DomainError" })
   | (SubstanceError & { errorType: "SubstanceError" })
-  | (StyleError & { errorType: "StyleError" });
+  | (StyleError & { errorType: "StyleError" })
+  | StyError; // COMBAK: Remove this
```
wodeni, line 11: Remember to remove.

packages/penrose-core/src/types/types.d.ts
```diff
@@ -56,7 +56,7 @@ type TrMap<T> = { [k: string]: { [k: string]: FieldExpr<T> } };
 interface ITrans<T> {
   // TODO: compGraph
   trMap: TrMap<T>;
-  warnings: string[];
+  warnings: StyError[];
```
wodeni, line 59: Does `StyError` exist, or it should be `StyleError`? I saw some renaming earlier

packages/penrose-core/src/utils/Error.ts
```diff
@@ -249,17 +329,30 @@ export const parseError = (message: string): ParseError => ({
...
+    tag: messages[0].tag, // COMBAK: is this confusing for showError?
+    messages: messages.map(showError),
+  }
+};
+
+export const genericStyleError = (messages: StyleError[]): PenroseError => ({
   errorType: "StyleError",
   tag: "GenericStyleError",
-  messages,
+  messages: messages.map(showError),
```
wodeni, line 348: `GenericStyleError` shouldn't exist anymore. Refer to my suggestion about `StyleErrorList`.

wodeni (changes requested):
Looks good overall. I'm mainly not sure about why `firstStyleError` is in use and gave you my suggestion below. `unsafelyGetErr` is not needed either, and I noted the correct usage below.

packages/penrose-core/src/compiler/Style.ts
hypotext: Nah, just wondering if something doesn't get a label in Substance, should it have no label at all, or if its label should be the empty string. So far the latter seems safer for Style.

packages/penrose-core/src/engine/EngineUtils.ts
hypotext: No, it's a consistent pattern in `insertExpr` that the error is returned if `compiling=true`, otherwise (at runtime) it makes an error that is thrown.

packages/penrose-core/src/compiler/Style.ts
hypotext: Resolved by passing in original path

packages/penrose-core/src/compiler/Style.ts
hypotext: > Ideally, I think you'll need to add the source locations of the relevant selector clauses as well. 

Yeah, the right thing to do is to pass in the selector environments, which have the locations.

> I can make checkVar return a more specific type and you can add a wrapper around those types and customize the error messages further if needed.

Not sure what you mean by this (but thanks for the offer); I feel like the above should suffice.

packages/penrose-core/src/compiler/Style.ts
hypotext: >  Looks like you are actually checking Style AST nodes transformed to Substance ones? If so, your source locs will be Style-specific? In general, the ASTNode type doesn't have file related information yet. Maybe we'll need it later.

Yeah, this should return Style source location. Since I'm reporting from Style, I can customize the error (as I guess you were suggesting here? https://github.com/penrose/penrose/pull/473#discussion_r574171134) to specify that the file is Style.

hypotext:
Thanks for the comments @wodeni! I addressed them in f7ac841. Want to take a look and let me know when this is good to merge?

packages/penrose-core/src/compiler/Style.ts
wodeni: Okay. Remember to fix the `unsafelyGetErr`. We can revisit this if it ends up harder to read for the users.

packages/penrose-core/src/compiler/Style.ts
wodeni: Yep, that was my suggestion. If you already have the right locations, we can just report this error as-is for now. File info is probably another PR.

wodeni (approved):
LGTM!

Issue #475, opened 2021-02-10
paths with references to nonexistent shape proeprty cause a confusing frontend crash


hypotext:
For a shape like a Circle, write a path to a nonexistent property like `x.shape.nonexistentProperty`. The frontend crashes as follows:

`Unhandled Rejection (TypeError): Cannot read property 'tag' of undefined`

I would expect it to fail during a translation checking phase with an error like this: `Path (...) contained a reference to a nonexistent property (...) on shape type (...). Did you mean (...)?`

hypotext:
Fixed in #484

Issue #474, opened 2021-02-10
labeling statement for a nonexistent GPI crashes system with confusing error


hypotext:
Repro: `roger watch mesh-set-domain/DomainInterop-unsugared.sub mesh-set-domain/DomainInterop.sty mesh-set-domain/DomainInterop.dsl` on most recent `style-errors`

Comment out `K.text =...` (L202) and `encourage centerLabel(K.icon, K.text)`. The line `K.text above K.icon` is still there. The system crashes:

`Unhandled Rejection (TypeError): Cannot destructure property 'shapeType' of 'undefined' as it is undefined.`

I would expect it to report an error like this before runtime:

`Cannot find GPI 'k.text' in layering statement (...) at line (...). Did you declare one first?`

hypotext:
Fixed in #484

PR #484, opened 2021-02-17
Do more Style/translation checking statically

hypotext:
# Description

**Motivation**: Currently Style does not fail if code does not run, just like JS. So if there is some wrong Style code that does not appear in the energy graph or the shape graph, we don’t report errors. Also, errors are currently thrown at eval-time (JS runtime) and mostly unchecked, so they result in cryptic messages from the JS runtime. 

**Proposal**: Check correctness of the blocks while making the translation, as well as the whole translation before evaluating it, and give people better errors ahead of time, that refer to source locations.

**Implementation**: This PR adds more static checking to Style and the translation:

1. Check the Style AST to make sure all shape, objective, constraint, and function names exist.
2. Check the initialized translation to make sure that all user-mentioned paths exist.

This prevents a large class of runtime JS errors.

Related issue/PR: #473 #474 #475
Related notes: 2/11/21 meeting notes

Expressions (`Expr`) are folded over recursively to aggregate all errors.

See `compiler/Style.test.ts` and `compiler/venn-3d.sty`.

# Checklist

# Open questions

- [ ] Implement further Style/translation checks described in 2/11/21 meeting notes (e.g. typechecking)
- [ ] Fix any errors that are commented out because a different error is thrown (e.g. `NonexistentNameError`)
- [ ] Report better errors and fixes (use `Identifier`s; reverse-map Substance object translation information to the substitution and Style variable that generated it)
- [ ] Test that multiple errors in a Style program work correctly
- [ ] Make `compiler/venn-3d.sty` a test suite

### DIFF
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -2465,7 +2465,7 @@ const getNum = (e: TagExpr<VarAD> | IFGPI<VarAD>): number => {
 // ported from `lookupPaths`
 // lookup paths with the expectation that each one is a float
 export const lookupNumericPaths = (ps: Path[], tr: Translation): number[] => {
-  return ps.map((path) => findExpr(tr, path)).map(getNum);
+  return ps.map((path) => findExprSafe(tr, path)).map(getNum);
 };
 
 const findFieldPending = (
@@ -2726,7 +2726,7 @@ const computeShapeOrdering = (tr: Translation): string[] => {
 // ---- MAIN FUNCTION
 
 // COMBAK: Add optConfig as param?
-const genOptProblemAndState = (trans: Translation): State => {
+const genOptProblemAndState = (trans: Translation): Result<State, StyleErrors> => {
   const varyingPaths = findVarying(trans);
   // NOTE: the properties in uninitializedPaths are NOT floats. Floats are included in varyingPaths already
   const uninitializedPaths = findUninitialized(trans);
@@ -2737,6 +2737,13 @@ const genOptProblemAndState = (trans: Translation): State => {
   const transInitFields = initFields(varyingPaths, trans);
   // sample varying vals and instantiate all the non - float base properties of every GPI in the translation
   const transInit = initShapes(transInitFields, shapePathList);
+
+    // CHECK TRANSLATION
+    // Have to check it after the shapes are initialized, otherwise it will complain about uninitialized shape paths
+  const transErrs = checkTranslation(transInit);
+  if (transErrs.length > 0) {
+      return err(transErrs);
+  }
 
   const shapeProperties = findShapesProperties(transInit);
   const [objfnsDecl, constrfnsDecl] = findUserAppliedFns(transInit);
@@ -2787,7 +2794,7 @@ const genOptProblemAndState = (trans: Translation): State => {
     varyingMap: {} as any, // TODO: Should this be empty?
   };
 
-    return initState;
+    return ok(initState);
 };
 
 //#endregion
@@ -3028,14 +3035,12 @@ export const compileStyle = (
   }
 
     // TODO(errors): `findExprsSafe` shouldn't fail (as used in `genOptProblemAndState`, since all the paths are generated from the translation) but could always be safer...
-  const initState = genOptProblemAndState(trans);
+    const initState: Result<State, StyleErrors> = genOptProblemAndState(trans);
   log.info("init state from GenOptProblem", initState);
 
-    // Have to check it after the shapes are initialized, otherwise it will complain about uninitialized shape paths
-  const transErrs = checkTranslation(initState.translation);
-  if (transErrs.length > 0) {
-    return err(toStyleErrors(transErrs));
-  }
+    if (initState.isErr()) {
+        return err(toStyleErrors(initState.error));
+    }
 
-  return ok(initState);
+    return ok(initState.value);
 };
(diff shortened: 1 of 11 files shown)


hypotext:
This is ready for review, lmk what you think @wodeni. I'm hoping to merge this tomorrow (Thurs) night, so it's available before the session with our first domain expert on Fri.

packages/core/src/compiler/Style.ts
```diff
@@ -1640,12 +1662,195 @@ const translateSubstsBlock = (
...
+  ) {
+    throw Error("internal error: expected function");
+  }
+
+  const fnDict = FN_DICT[expr.tag];
+  if (!fnDict) {
+    throw Error("internal error: unexpected tag");
+  }
+
+  const fnNames: string[] = Object.entries(fnDict).map((e) => e[0]); // Names of built-in functions of that kind
```
wodeni, line 1731: Can use this one instead: https://lodash.com/docs/4.17.15#keys

packages/core/src/compiler/Style.ts
```diff
@@ -1640,12 +1662,195 @@ const translateSubstsBlock = (
...
+
+  const givenFnName: Identifier = expr.name;
+
+  if (!fnNames.includes(givenFnName.value)) {
+    const fnErrorType = FN_ERR_TYPE[expr.tag];
+    if (!fnErrorType) {
```
wodeni, line 1737: Maybe you could avoid excessive checking like this by narrowing down the type of `expr` first to something like `ObjFn | ConstrFn | CompApp`. It's okay if that's not possible.

packages/core/src/compiler/Style.ts
```diff
@@ -1640,12 +1662,195 @@ const translateSubstsBlock = (
...
+  // NOT CHECKED as this requires more advanced env-building work: At path construction time, check that every Substance object exists in the environment of the block + selector, or that it's defined as a local variable
+
+  const res: StyleResults = block.statements.reduce(
+    (acc: StyleResults, stmt: Stmt): StyleResults =>
+      checkLine(selEnv, stmt, acc),
+    emptyErrs()
+  );
+
+  // TODO(errors): Return warnings (non-fatally); currently there are no warnings though
+  if (res.warnings.length > 0) {
```
wodeni, line 1844: Looking at the comment: should this one throw, or you are supposed to do something nonfatal here?

packages/core/src/compiler/Style.ts
```diff
@@ -2633,6 +2853,105 @@ export const disambiguateFunctions = (env: Env, subProg: SubProg) => {
...
+      const res: Path[] = findPathsExpr(fexpr.contents.contents);
+      return acc.concat(res);
+    } else {
+      return acc;
+    }
+  } else if (fexpr.tag === "FGPI") {
+    // Get any exprs that the properties are set to
+    const propExprs: Expr[] = Object.entries(fexpr.contents[1])
+      .map((e) => e[1])
+      .filter((e: TagExpr<VarAD>): boolean => e.tag === "OptEval")
```
wodeni, line 2933: You can do this in one single lambda function, so TS would know the type of `e` is changed to `OptEval` (in an `if` statement, for instance).

packages/core/src/engine/Evaluator.ts
```diff
@@ -72,79 +73,79 @@ const dummyIdentifier = (name: string): Identifier => {
...
+
+    // Clone translation to use in this top-level call, because it mutates the translation while interpreting the energy function in order to cache/reuse VarAD (computation) results
+    const trans = clone(transWithVarying);
+
+    // Find out all the GPI expressions in the translation
+    const shapeExprs: IFGPI<VarAD>[] = s.shapePaths.map(
+        (p: Path) => findExprSafe(trans, p) as IFGPI<VarAD>
+    );
+
+    console.log("shapePaths", s.shapePaths.map(prettyPrintPath));
```
wodeni, line 106: Use logger

packages/core/src/utils/Error.ts
```diff
@@ -20,460 +20,504 @@ const {
...
+
+        case "TaggedSubstanceError": {
+            return showError(error.error); // Substance error
+        }
+
+        // --- BEGIN BLOCK STATIC ERRORS
+
+        case "InvalidGPITypeError": {
+            return `Got invalid GPI type ${error.givenType.value}.`;
+            // COMBAK: GPI type suggestions, or just print the list
```
wodeni, line 190: I think printing out the list of GPI types is fine. Should be easy to do?

wodeni (changes requested):
Overall looks good to me! Only some minor suggestions below.

packages/core/src/compiler/Style.ts
hypotext: Not sure how I would do this in a lambda in this context, as I would need to be doing both a map and a filter in it. But can do this in the future.

packages/core/src/compiler/Style.ts
hypotext: Yeah, it should throw as a reminder to deal better with warnings in the future. There are currently no warnings in the system, so this is fine. But we should think about how to aggregate/accumulate them in the future, as currently there are only errors, which are reported as if everything is fatal.

packages/core/src/utils/Error.ts
hypotext: Did this, and the properties too

hypotext:
Thanks @wodeni! Addressed changes in f2657d8. LMK when it's good to merge.

packages/core/src/compiler/Style.ts
wodeni: Just saw the `filter`. Don't really know a great alternative :(.

wodeni (approved):

PR #490, opened 2021-02-19
feat: add Polygon and Polyline to the shape library

Dsm0:
I've added one lone shape: the polygon, loosely according to the svg specification

https://www.w3.org/TR/SVG2/shapes.html#PolygonElement


joshsunshine:
@Dsm0 Please remove Nimo's code so we can approve this PR.

PR #505, opened 2021-03-03
chore: add registry.json

maxkrieger:
We have a new `examples/registry.json` which is more human readable than the existing `styleLibrary.json` etc.

It has set theory stuff but needs more domains supported.


wodeni:
Included in #507. Closing to minimize admin overhead.

PR #506, opened 2021-03-03
build: core+browser-ui use estrella for `watch` jobs

maxkrieger:
# Notes
*  For sourcemaps to work, `core/build/` and `browser-ui/build/` dirs must have no `.map` files, so that chrome doesn't get confused and mess the paths up. Instead we use esbuild's `sourcemaps: "internal"` which puts them inline. **This makes the bundle size HUGE, make sure not to publish it**. Since `.map` files are only produced now for building production, they should rarely be around (maybe `npm publish` should remove the `build/` folder afterwards, to get rid of stale `.map` files.
   * "WHY WOULD YOU DO THIS MAX??"
   * because of our monorepo/linking structure, chrome can't find `localhost:3000/core/src/xxx` (which is referenced **relatively** in sourcemaps) from `browser-ui`'s web server. However, somehow, chrome **does** know where these files are if we have the sourcemap and the source code in the same file. An alternative approach may have been to convince `esbuild` to include [`sourcesContent`](https://esbuild.github.io/api/#sources-content) in the `.map` files
* `esm` and `cjs` are both necessary for `browser-ui` to resolve `@penrose/core`
* `error: Could not resolve "src" (use "./src" to reference the file "src/index.ts")` is actually non-fatal, and not sure the cause

## Old notes

Known issues:

- [ ] `esm` and `cjs` are both necessary for `browser-ui` to resolve `@penrose/core`
- [ ] `error: Could not resolve "src" (use "./src" to reference the file "src/index.ts")` is actually non-fatal, and not sure the cause
- [x] Sourcemaps point to the right line number but chrome can't retrieve the files
    * https://stackoverflow.com/q/60172083/10833799
    * https://github.com/rsms/estrella/issues/30
- [x] `path`, `crypto`, etc cannot be resolved for `platform: "browser"` so we use `platform: "node`". This results in:
- [x] `> ../../node_modules/consola/dist/consola.js:1:52608: warning: Indirect calls to "require" will not be bundled (surround with a try/catch to silence this warning)1 │ ...ack_require__ __non_webpack_require__:require;class zu{constructor(u {if(u&&u.log).. ╵                                          ~~~~~~~` (non-fatal)
  - [x] Nimo suggestion: make sure consola actually works
- [x] Some other `require` warning in the browser console

These errors are all quite irritating, but can potentially be hackily silenced. And builds are faster.

The question is if we want to do the same for `browser-ui`, as I sense that's the main chokepoint for build times.

Lemme know if you think it's worth using in its current state @hypotext !


maxkrieger:
- [ ] browser-ui estrella

maxkrieger:
@wodeni @hypotext everything works deliciously, try it out 🍰

maxkrieger:
Workarounds for sourcemap issue:

* Replace `serve-http` with some custom server that serves the `core/` route
* Overwrite the sourcemap json in `core` to not use relative paths
* Make `sourcesContent` actually work https://esbuild.github.io/api/#sources-content

wodeni (approved):
LGTM!

PR #519, opened 2021-03-08
build: emit typescript declarations in watch

maxkrieger:
Oops!

### DIFF
--- a/packages/core/tsconfig.json
+++ b/packages/core/tsconfig.json
@@ -24,7 +24,6 @@
     "typeRoots": ["src/types", "node_modules/@types"],
     "esModuleInterop": true,
     "isolatedModules": true,
-    "noEmit": true,
     "strict": true,
     "declaration": true,
     "declarationMap": true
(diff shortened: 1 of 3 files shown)


packages/core/tsconfig.json
```diff
@@ -24,7 +24,6 @@
     "typeRoots": ["src/types", "node_modules/@types"],
     "esModuleInterop": true,
     "isolatedModules": true,
-    "noEmit": true,
```
wodeni, line 27: Does this affect the existing build scripts?

maxkrieger: good question, I did a quick check with rollup and I believe it overrides this flag anyways. But I am indeed not sure why it was there to begin with, perhaps an artifact of CRA.

hypotext:
This does appear to resolve the cryptic circular build dependency crash I reported earlier. Now, when I add the line `import { shapedefs } from "renderer/ShapeDef";` in `EngineUtils`, the system seems to build and run fine (not reporting any errors either).

hypotext (approved):
WFM

Issue #521, opened 2021-03-08
deal with aliasing in translation


k-qy:
Currently some operations don't work WRT path aliasing. For example, something like this doesn't work:

```
o.v1 = (?, 2.)
o.shape = Line {
start: o.v1
}
override o.shape.start[0] = 0.
```

Presumably, you want then to look up the reference `o.v1` and set its first element to 0.

samestep:
I think this should be closed as of #984, because we no longer support `override` on access paths.

Issue #515, opened 2021-03-06
overriding accesspaths, and setting partial vectors, doesn't work


hypotext:
**Describe the bug**

Overriding accesspaths crashes the system in different ways. See test case below.

**To Reproduce**

On branch `vector-debug`, see `accesspath-debug.sty`.

```
-- (2a) DOES NOT WORK
-- Cannot read property 'tag' of undefined -- in `addPath` in `translateLine`

Object o {
    shape o.shape = Line {
        color: rgba(0., 0., 0., 1.)
        thickness: 5.
        rightArrowhead: true
    }

    o.shape.start[0] = 0.
}

-- (2b) DOES NOT WORK
-- internal error: invalid varying path (in `getNum, in lookupNumericPaths`)

-- Object o {
--     shape o.shape = Line {
--         start: (0., ?)
--         color: rgba(0., 0., 0., 1.)
--         thickness: 5.
--         rightArrowhead: true
--     }
-- }

-- (2c) WORKS

-- Object o {
--     o.y = ?
--     shape o.shape = Line {
--         start: (0., o.y)
--         color: rgba(0., 0., 0., 1.)
--         thickness: 5.
--         rightArrowhead: true
--     }
-- }
```

**Expected behavior**
Setting or overriding accesspaths works (sets the individual variable).

Probably related to issues in #501.

hypotext:
Resolved in 5eba2ed203f6912613934c3247503563e77f6f0c

Issue #513, opened 2021-03-05
Polygon/Polyline optimizing points


Dsm0:
There's an issue with the optimization of point values for polygons and polylines.
The issue can be examined on the polygon branch, with the `shape-spec` example.

The optimization for number-literal values behaves as it should, (although it cannot be dragged).
```
forall Polygon j {
	j.icon = Polygon { 
        points: ((-80.0, 0.0), (100.0, 0.0), (10.0, 50.0))
        strokeWidth: 200
	}
}
```

When point values are declared with question-marks, the style compiler crashes.
```
forall Polygon j {
	j.icon = Polygon { 
        points: ((?, ?), (100.0, 0.0), (10.0, 50.0))
        strokeWidth: 200
        pathLength :2000
	}
}
```

However, when point values are declared with variables, and those variables are question marks, the style compiles, but there's some strange dragging behavior.

```
forall Polygon j {
    x1 = ?
    y1 = ?
    x2 = ?
    y2 = ?
    x3 = ?
    y3 = ?
	j.icon = Polygon { 
        points: ((x1,y1), (x2,y2), (x3,y3))
        strokeWidth: 200
	}
}
```

hypotext:
Hey @Dsm0 I'm happy to look into this. The issue is probably that we never tested polygon optimization.

Can you include a minimal program triple (or commit on a branch) that I can run to reproduce the problem? Can you also include the full Style errors you are getting, as well as more info about the dragging behavior? Thanks!

hypotext:
Ah, I missed your information here: 

> The issue can be examined on the polygon branch, with the shape-spec example.

Will investigate when I get a chance -- polygon optimization doesn't currently seem super high-priority, but let me know if this becomes a blocker for you.

wodeni:
Cannot repro this anymore. Tried the following and it seems to be just fine:

```
canvas {
    width = 100
    height = 100
}

global {
    Polygon {
        points: [(0, 0), (?, ?), (20, 20)]
    }
}
```

PR #522, opened 2021-03-08
Fix vector initialization, as well as overriding elements of vector-typed shape properties

### LINKED ISSUE #501: optimizing a vector as a field in Style doesn't work
There is no gradient for the very simple program triple that @srossd sent to us today, though there is one constraint, and the generated energy appears to be correct. Maybe an issue with finding `AccessPath`?

energy:
```
(function anonymous(x0,x1,x2,x3,x4
) {
const x5 = 0;
const x6 = 0;
const x7 = 252.1585483546197;
console.log("x1 (var x7) | value: ", x7, "during energy evaluation");
const x8 = 75.30582642063939;
console.log("x2 (var x8) | value: ", x8, "during energy evaluation");
const x9 = x7 - x8;
console.log("energy (var x9) | value: ", x9, "during energy evaluation");
const x10 = 0;
const x11 = Math.max(x9, x10);
const x12 = x11 * x11;
const x13 = x6 + x12;
const x14 = 100000;
const x15 = x14 * x0;
const x16 = x13 * x15;
const x17 = x5 + x16;
return x17;
})
```

grad: (wrong)
```
generated f ƒ anonymous(x0,x1,x2,x3,x4
) {
const x5 = 0;
const x6 = 0;
const x7 = 0;
const x8 = 0;
return [x5, x6, x7, x8];
}
```

Now sure why/how the energy would be correct but not the gradient.

Oddly, a very similar function appears to work in the `graph-domain` branch: https://github.com/penrose/penrose/blob/graph-domain/examples/graph-domain/graph-theory.sty#L110

It appears to be linked to the use of fields, or perhaps path aliases/references in general. This Style program works when `o.v1` and `o.v2` are removed, and it doesn't seem to matter whether `start` and `end` are declared explicitly as varying `vec2s` or not.

```
Object o {
    -- Remove these lines
    -- vec2 o.v1 = (?, ?)
    -- vec2 o.v2 = (?, ?)

    shape o.shape = Line {
        -- Optional: start: (?, ?) -- remove this: o.v1
        -- Optional: end: (?, ?) -- remove this: o.v2
        color: rgba(0., 0., 0., 1.)
        thickness: 5.
        rightArrowhead: true
    }

    ensure rightwards(o.shape) -- ensure lines move from left to right
}
```

hypotext:
# Description

Related issue/PR: #501 #515 #521

This PR fixes several bugs in handling vectors and accesspaths.

* add `VaryAD` case to `AnnoFloat` in `Expr`, so `Expr`s can hold `VarAD`s. this prevents `VarAD`s from getting coerced to floats and losing their gradient info (which was a possible bug noted earlier in a comment). fixes #501
* add PropertyPath as a kind of AccessPath to initialize in genOptProblem, and then initialize the shapes before the accesspaths in genOptProblem so that properties can be properly set. this resolves part of #515
* initialize shape properties that are vector-typed by default in `insertExpr`, so overrides like `s.shape.start[0]` can work even if `s.shape.start` is not set in Style. also encode working programs as tests and break resample into separate file to avoid circular import with `EngineUtils` #515

Working program for opt (#501): `roger watch test.sub monoidal-test.sty monoidal.dsl`

Tests for #515 are added to the `"Correct Style programs"` test suite in `Style.test.ts`.

# Checklist

# Open questions

- Deal with aliasing #521


hypotext:
@wodeni It would be awesome to get a once-over on this before noon tomorrow (since I'm working with Ross tomorrow afternoon). Thanks!

hypotext:
I know 5947132 isn't technically directly related to this PR but it's probably good to have some of the new domain in `main` by now, for discussion/testing with our domain expert user.

packages/core/src/compiler/Style.ts
```diff
@@ -2516,9 +2512,15 @@ const isFieldOrAccessPath = (p: Path): boolean => {
...
 // NOTE: Mutates translation
-const initFields = (varyingPaths: Path[], tr: Translation): Translation => {
-  const varyingFields = varyingPaths.filter(isFieldOrAccessPath);
-  const sampledVals = randFloats(varyingFields.length, canvasXRange);
+const initFieldsAndAccessPaths = (
+  varyingPaths: Path[],
+  tr: Translation
+): Translation => {
+  const varyingFieldsAndAccessPaths = varyingPaths.filter(isFieldOrAccessPath);
+  const sampledVals = randFloats(
```
wodeni, line 2520: For a temporary solution, this seems to be fine. Ideally, I'd prefer if we have a way to operate on `Shape[]` instead of varying paths directly. This is somewhat connected to #513. The main problem we have seems to be: if we only look at shapes, we have no idea if a property came from a field, e.g.
```
Set X {
  val = ?
  X.icon = Circle { r: val }
}
```
In the translation, `X.icon.r` is a `LocalVar`. In `state.varyingPaths`, `val` is some local field path. So when dragging a shape, there's no obvious way to "trace" the origin of `X.icon.r` and back propagate the changes. 

I guess another possible solution is to maintain some data structure that points from shape properties to varying paths. We can (1) sample shapes only, pull prop  values out, and insert them into `varyingValues` and (2) update shapes on front-end updates (e.g. drag) and do the same thing to get the next `varyingValues`.

wodeni (approved):
LGTM. I had a high-level comment about shape vs. varying values updates, which is related to #513.

packages/core/src/compiler/Style.ts
hypotext: Hmm. I will give this more careful thought and get back to you tonight.

hypotext: > So when dragging a shape, there's no obvious way to "trace" the origin of `X.icon.r` and back propagate the changes.

Why not look in the translation for `X.icon.r`? That would give you `X.val` which you can then lookup in `varyingPaths`/`varyingValues` to confirm it's a varying value and update it with the results of the drag. Not sure if I'm missing something about this method.

wodeni: Here's a screenshot of the example above:

image

hypotext: Ok, this is indeed a bug in the Style compiler. All the local vars should be substituted out, ideally. Thanks for reporting it! Issue tracked here, will aim to fix today. https://github.com/penrose/penrose/issues/523

Issue #520, opened 2021-03-08
Move `types.d.ts` to respective independent type `.ts` files


maxkrieger:
This will make exported types richer, particularly functions that take in/return `State`. May induce merge conflcits.

PR #526, opened 2021-03-11
fix: #520 (moving types out, fixing soundness)

maxkrieger:
# Description

Related issue/PR: #520 

# Checklist

- [ ] New and existing tests pass locally using `yarn test`

# Open questions


maxkrieger:
@wodeni should probs merge asap if it lgty

packages/core/src/types/shapeEvalTypes.ts
```diff
@@ -0,0 +1,477 @@
+import { VarAD } from "./adTypes";
```
wodeni, line 1: Name change:

packages/core/src/types/shapeEvalTypes.ts
```diff
@@ -0,0 +1,477 @@
...
+  tag: "Done";
+  contents: Value<T>;
+}
+
+export interface IPending<T> {
+  tag: "Pending";
+  contents: Value<T>;
+}
+
+export type Expr =
```
wodeni, line 85: This belongs to style AST technically

packages/core/src/types/shapeEvalTypes.ts
```diff
@@ -0,0 +1,477 @@
...
+  tag: "StringLit";
+  contents: string;
+}
+
+export interface IBoolLit extends ASTNode {
+  tag: "BoolLit";
+  contents: boolean;
+}
+
+// COMBAK: This seems to be unused, delete
```
wodeni, line 129: Can remove this and the one below

packages/core/src/types/styleASTTypes.ts
```diff
@@ -0,0 +1,189 @@
+//#region Style AST
+// TODO: unify type name convention (e.g. stop using `I` for interfaces and drop some of the Haskell ported types)
+
+import { ASTNode, Identifier } from "./ASTTypes";
+import { StyVar, BindingForm, Expr, Path } from "./shapeEvalTypes";
```
wodeni, line 5: These should all be defined in this file instead

packages/core/src/types/substanceASTTypes.ts
```diff
@@ -0,0 +1,116 @@
+import { LabelMap } from "compiler/Substance";
+import { ASTNode, Identifier } from "./ASTTypes";
+import { TypeConstructor } from "./domainASTTypes";
+import { IStringLit } from "./shapeEvalTypes";
```
wodeni, line 4: Move this to ASTTypes

packages/core/src/types/substanceASTTypes.ts
```diff
@@ -0,0 +1,116 @@
+import { LabelMap } from "compiler/Substance";
```
wodeni, line 1: Move this type to this file

packages/core/src/types/types.d.ts
```diff
@@ -1,1614 +1,2 @@
 // TODO: find a way to put this in index.d.ts
```
wodeni, line 1: Remove this file after resolving this TODO item?

wodeni (changes requested):
Overall LGTM. Need to move some things around as noted below. Also can you check if you have any circular deps (maybe by https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-cycle.md?). Circular imports themselves aren't problematic, but it shows illogical module organization quite often.

maxkrieger:
Good point, will cycle check tomorrow

packages/core/src/types/shapeEvalTypes.ts
maxkrieger: ?

packages/core/src/types/shapeEvalTypes.ts
maxkrieger: the whole thing? all the sum type members r in there

packages/core/src/types/styleASTTypes.ts
maxkrieger: these meaning https://github.com/penrose/penrose/pull/526#discussion_r592051644 ?

packages/core/src/types/substanceASTTypes.ts
maxkrieger: IStringLit?

wodeni (approved):

Issue #511, opened 2021-03-04
Top-level API for energy evaluation/analysis


wodeni:
Our top-level API does not allow clients to evaluate constraints/objectives in custom ways. Here are some possible use cases:

* Evaluate the current energy value of a `State`:
```ts
const evalEnergy = (s: State): number;
```
* Evaluate a list of constraints/objectives: this will be useful if a user want to apply a subset of constrs/objs on a `State`
```ts
const evalFns = (fn: Fn[], s: State): number[]; // TODO: constraint/objective eval might be different?
```

Use case:

```ts
const s: State = compileTrio(dsl, sub, sty);
// get the overall energy of the state
const eng: number = evalEnergy(s); 
// get a list of `contains` instances
const containFns: Fn[] = s.constrFns.filter(c => c.fname === "contains");
// only eval `contains` in `s` and ignore all other constraints
const containsEngs: number[] = evalFns(containFns, s); 
// assuming there are more constraints in `s`, the (squared?) sum of `containsEngs` should be smaller than `eng`
```

wodeni:
Other functions per discussion in our team meeting (related: #528)

```ts
// return the gradients for the state, cached if `s` is already initialized and run `prepareState` if not.
evalGradients(s: State): number[]; 
// 
```

hypotext:
Resolved in #512 #527

PR #527, opened 2021-03-11
feat: [layout] add `evalFns` for individual opt fns, and compile each opt function

hypotext:
# Description

Related issue/PR: #511 #512

We need more granular codegen + evaluation for improving our solver, and for other uses like diagram generation.

Changes: Add a new API function `evalFns` to evaluate individual functions on the state (with no weights). Also, precompile each obj/constrfn and cache its energy and gradient in the state in `prepareState`.

- add a new API function for `evalFns`
- add `genFn` based on `genOptProblem`
  - write a new version of `evalEnergyOnCustom` without the weight node stuff, and only on a single objective or constraint
  - change existing `energyAndGradCompiled` so it accepts `MaybeVal<WeightInfo>`
- write a `genFns` wrapper that looks at all opt functions in the state and caches their info in the state, in an `{ fnName(args) => { energy, gradient } }` map
- write `evalFns` on multiple functions (and their gradients) that uses the cache if it exists
- add a test that verifies that running `evalFns` on a simple triple's state before and after optimization yields minimized objective energies and satisfied constraints

In `__tests__`, run `npm test -- -t "Run individual functions"`.

# Checklist

# Open questions

- I didn't test that each individual generated function is actually correct (that is, I didn't inspect the generated code). The tests give a pretty high degree of assurance that it's overall correct, but there could still be bugs.
- Return individual gradients, for use in more sophisticated layout algorithms.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -8,7 +8,7 @@ import {
 } from "compiler/Substance";
 import consola, { LogLevel } from "consola";
 import { evalShapes } from "engine/Evaluator";
-import { genOptProblem, initializeMat, step } from "engine/Optimizer";
+import { genOptProblem, genFns, initializeMat, step } from "engine/Optimizer";
 import { insertPending } from "engine/PropagateUpdate";
 import RenderStatic, {
   RenderInteractive,
@@ -19,6 +19,7 @@ import { PenroseError } from "types/errors";
 import * as ShapeTypes from "types/shapeTypes";
 import { collectLabels } from "utils/CollectLabels";
 import { andThen, Result, showError } from "utils/Error";
+import { prettyPrintFn } from "utils/OtherUtils";
 import { bBoxDims, toHex } from "utils/Util";
 
 const log = consola.create({ level: LogLevel.Warn }).withScope("Top Level");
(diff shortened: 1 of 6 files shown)


packages/core/src/index.ts
```diff
@@ -223,6 +225,41 @@ export const evalEnergy = (s: State): number => {
...
+ */
+export const evalFns = (fns: Fn[], s: State): number[] => {
+  const { objFnCache, constrFnCache } = s.params;
+
+  // NOTE: if `prepareState` hasn't been called before, log a warning message and generate a fresh optimization problem
+  if (!objFnCache || !constrFnCache) {
+    log.warn(
+      "State is not prepared for energy evaluation. Call `prepareState` to initialize the cached objective/constraint functions first."
+    );
+    const newState = genFns(s);
```
wodeni, line 241: Can this be extended to allow not just a subset of existing obj/constrs? Say the user constructs a new constraint and passes it as a part of `fns`. Is `genFn` enough for including this new constraint in the opt problem?

wodeni (commented):
LGTM overall! Just had one small Q about `genFns` (see detailed comments).

packages/core/src/index.ts
hypotext: No, it only generates the function for that individual function. If you want to include a new constraint/objective in the overall opt problem (including having an overall gradient), it will need to be properly weighted and added to the rest of the evaluated functions, so you should regenerate the whole function using `evalEnergyOnCustom`. 

I suppose if you really wanted to, you could compile the individual function and evaluate it, then do the same weighting-adding process to its results using pure JS math, and then manually change the gradient to include the individual function ... but that seems very prone to bugs.

PR #530, opened 2021-03-15
feat: a more general front-end interaction update algorithm 

wodeni:
# Description

Related issue/PR: #513 

The current semantics of drag behaviors assumes the following:
1. Varying paths can only be shape properties (e.g. `x.icon.y`)
2. Front-end interaction will only update positional properties

When a user interacts with shape in the frontend, `state.shapes` gets updated, and the backend updates the varying paths given the new data in `state.shapes`. The current algorithm walks through all varying paths, finds the corresponding values in `state.shapes`, and updates varying values.

However, this model breaks down when 1 and/or 2 are not true. For instance, if a property is defined as path pointing to a varying field like so:

```
Circle E {
    x = ?
    y = 50
    E.icon = Circle {
        center: (x, y)
    }
}
```

This should be the desired behavior on drag, but the current algorithm will cause the circle to snap back all the time:
drag-partial

In this PR, I implement a more general algorithm for back-propagating front-end interaction updates.
* When generating the `State`, pre-compute mappings from "origin paths" to all paths that use them.
* Codify positional properties of shapes by declaring them in shape definitions
* In `dragUpdate`, move all positional properties, instead of moving a list of hard-coded properties
* When back-propagating shape updates, look up each varying path and retrieve values from their descendants.  

`npx roger watch shapes.sub shape-spec.sty shape-spec.dsl`

# Checklist

- [ ] New and existing tests pass locally using `yarn test`

# Open questions

### DIFF
--- a/examples/shape-spec/shape-spec.sty
+++ b/examples/shape-spec/shape-spec.sty
@@ -11,7 +11,11 @@ Square D {
 }
 
 Circle E {
-    E.icon = Circle { }
+    x = ?
+    y = 50
+    E.icon = Circle {
+        center: (x, y)
+    }
 }
 
 Rectangle F {
--- a/examples/shape-spec/shapes.sub
+++ b/examples/shape-spec/shapes.sub
@@ -1,8 +1,9 @@
 -- Path A
-Arrow C
-Square D
+-- Arrow C
+-- Square D
 Circle E
-Rectangle F
-Line G
-Text H
-Image I
+-- Rectangle F
+-- Line G
+-- Text H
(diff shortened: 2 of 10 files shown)


joshsunshine:
What do you mean by "This should be the desired behavior on drag"?

wodeni:
> What do you mean by "This should be the desired behavior on drag"?

Since `y` is fixed and `x` is optimized, the correct behavior should be the one shown in the gif. In `main`, the circle wouldn't be draggable at all.

PR #533, opened 2021-03-25
feat: process prelude values in substance checker

wodeni:
# Description

Related issue/PR: N/A

Currently, the system would parse prelude values, but doesn't really include them in the rest of the compilation pipeline. This PR adds support for processing prelude values with nullary types in the Substance checker.

* Prelude declarations are retrieved from the Domain environment and translated to Substance statements by `toSubDecl``
* In `compileSubstance`, prelude values are appended to the parsed AST _before_ the checker runs (so that the checker can include metadata like varying mappings in the Substance environment)
* We do _not_ handle prelude values with dependent types for now.

* In `Substance.test.ts`, I added `value X: Set` to the common Domain program, and a test checks for prelude values.

# Checklist

# Open questions

- [ ] Error locations in Domain/Substance with prelude line added (see [comment](https://github.com/penrose/penrose/pull/533#issuecomment-815133150))

### DIFF
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -8,6 +8,8 @@ import {
   addWarn,
   defaultLbfgsParams,
   dummyASTNode,
+  dummyIdentifier,
+  dummySourceLoc,
   findExpr,
   findExprSafe,
   initConstraintWeight,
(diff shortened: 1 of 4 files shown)


hypotext:
LGTM, thanks for doing this!

Minor comment: I wonder if errors involving the prelude will have the right source locations (and if prepending the prelude will affect other source locations). 

If, for example, a variable name in Substance clashes with the prelude name, will the error be reported for a nonexistent Substance line (to the end-user)? Or if the error is reported for the Domain line, will the end-user know that it originates from the Domain file? If there is a non-prelude-related Substance error, will the line numbers still be correct?

I think this PR is fine (will merge so we can use it) but if error issues arise we should address them later. :)

wodeni:
> Minor comment: I wonder if errors involving the prelude will have the right source locations (and if prepending the prelude will affect other source locations).

I don't think it'll affect the source locs. The appending is at the tree-level, not string-level. So it's basically a dummy node.

PR #534, opened 2021-03-29
feat: opt tab for inspector; localStorage settings

maxkrieger:
*Autostep and show inspector are now togglable across refreshes via localStorage. By default autostep's off and inspector's on.

* We added a simple sortable table with energies of functions. More features to come

image

### DIFF
--- a/packages/browser-ui/src/inspector/views/Opt.tsx
+++ b/packages/browser-ui/src/inspector/views/Opt.tsx
@@ -40,7 +40,7 @@ const Opt: React.FC<IViewProps> = ({ frame, history }: IViewProps) => {
         highlightOnHover={true}
         striped={true}
         columns={[
-          { name: "Expression", sortable: true, selector: "name" },
+          { name: "Expression", selector: "name", sortable: true },
           { name: "Energy", selector: "energy", sortable: true },
           { name: "Type", selector: "type", sortable: true },
           { name: "Satisfied?", selector: "sat", sortable: true }
--- a/packages/browser-ui/src/inspector/views/Opt.tsx
+++ b/packages/browser-ui/src/inspector/views/Opt.tsx
@@ -14,11 +14,21 @@ const Opt: React.FC<IViewProps> = ({ frame, history }: IViewProps) => {
   const constrInfos = zip(
     frame.constrFns.map(prettyPrintFn),
     evalFns(frame.constrFns, frame)
-  ).map(([name, energy]) => ({ name, energy, type: "constraint" }));
+  ).map(([name, energy]) => ({
+    name,
+    energy,
+    type: "constraint",
+    sat: energy! <= 0 ? "yes" : "no"
+  }));
   const objInfos = zip(
     frame.objFns.map(prettyPrintFn),
     evalFns(frame.objFns, frame)
-  ).map(([name, energy]) => ({ name, energy, type: "objective" }));
+  ).map(([name, energy]) => ({
+    name,
+    energy,
+    type: "objective",
+    sat: energy! <= 0 ? "yes" : "no"
+  }));
 
   // TODO: hyperlink the shapes
   return (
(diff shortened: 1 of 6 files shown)


packages/browser-ui/src/inspector/views/Opt.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <div>
+      <DataTable
+        data={[...constrInfos, ...objInfos]}
+        title={"Optimization"}
+        dense={true}
+        highlightOnHover={true}
+        striped={true}
+        columns={[
+          { name: "Expression", sortable: true, selector: "name" },
```
hypotext, line 33: change order of `selector` and `sortable` to be consistent?

hypotext (changes requested):
Thanks for this! @wodeni did you want to take a look before we merge?

Some comments:

* Not sure why the text disappears from the diagram when I click on an inspector frame. Would be good to fix this first.
Before:


After: (note selected frame)

https://ci6.googleusercontent.com/proxy/exedg8Ev6mD5XUFcB7pAaF20aiCKw0V9ikLdYpxQqCpi7pzoJRABFLmTOu_Tz3SO6xBzDzgYtrdJMmbX_hNF5Ivw0y1UpT-OIIPOC4vYU_2mv0mrcs9m8rNwVAvAxRpRwMJ7W6g_F7jaR8iJ88ygm-rgn7UnLF8=s0-d-e1-ft#https://user-images.githubusercontent.com/2762356/112903461-89c80700-90b5-11eb-84cb-522eb4d035a9.png

* If `autostep` is on, I am confused by the `opt` values shown for each frame (e.g. the past frames). They do seem to be changing. Are those the actual values for the past step in the optimization? Is the inspector auto-saving the history of values, and if so, how/where is it doing that? If not, what are those values from?

May be confusing to existing users if we have `autostep` off by default. Maybe we should just let them know.

wodeni:
> Thanks for this! @wodeni did you want to take a look before we merge?

Already read through the changes. Looks great. Feel free to merge whenever you are ready 💯 !

maxkrieger:
Text disappearance was always the case not sure if it's accentuated now. I think it's an ID conflict... or maybe global symbols..

maxkrieger:
> * If `autostep` is on, I am confused by the `opt` values shown for each frame (e.g. the past frames). They do seem to be changing. Are those the actual values for the past step in the optimization? Is the inspector auto-saving the history of values, and if so, how/where is it doing that? If not, what are those values from?

The history of the values is in the History object. It's just the states. The past frames are visually different states, no? It may be adding more than one frame due to some sort of eventing quirk... Hmm

Autostep is off by default btw.

hypotext:
> The history of the values is in the History object. It's just the states.

Ok, trying to state my question more clearly:

* I thought the `opt` tab only calculates the `opt` info for the current frame (via `evalFns`, etc) and displays them. It doesn't store the values anywhere.
* However, when I click on past frames, I see different opt values in the tab `opt`.
* Where are those past frames` opt values coming from? 

I guess if the text thing isn't a regression then we don't have to fix it now, though it would be good to have an issue for it.

hypotext:
I also noticed that the `opt` tab doesn't work in conjunction with the `mod` tab, i.e. if I use the `mod` tab to change some values, the opt values are not re-updated. That makes harder to test energies, though we can get to that in a second pass.

maxkrieger:
image

packages/browser-ui/src/inspector/views/Opt.tsx
maxkrieger: done :D

hypotext (approved):
LGTM! These features have already been massively useful for me on the `graph-domain` branch. (Another TODO is to, at some point, document the opt tab on the wiki.)

PR #536, opened 2021-03-30
svg command

Dsm0:
This PR _will_ include an svg option to pass to roger. This option won't connect to any active server, instead compiling penrose files to an svg locally. It would be incredibly useful for quick local testing of language features.  

The code is almost identical to the code in `watch.ts`, just with some calls replaced.

There are currently some dependency issues, involving importing "@penrose/core" and "global-jsdom". Currently, it does not compile properly. But once those are resolved, it should be trivial to write compiled svg to a local file.

### DIFF
--- a/packages/roger/README.md
+++ b/packages/roger/README.md
@@ -23,7 +23,7 @@ $ npm install -g @penrose/roger
 $ roger COMMAND
 running command...
 $ roger (-v|--version|version)
-@penrose/roger/1.0.0 darwin-x64 node-v12.19.0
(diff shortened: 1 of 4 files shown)


Dsm0:
While I thought this would be a simple feature, the `mathjax-full` module for `@penrose/core` makes reference to a document object that doesn't exist unless the javascript is running inside a browser.

Given that there's not much new functionality being added, I'm putting this on the back-burner until we develop plans for a more robust command-line interface that doesn't just add one already-present feature.

Issue #504, opened 2021-03-03
disjoint constraint (non-continuous fn on rectangle) causes big optimizer slowdown


hypotext:
**Describe the bug**
As reported by @srossd: 

> I was trying to use the disjoint constraint on a rectangle and a line for my domain. I noticed the constraint didn't quite work - if I understand the old code correctly, it was keeping the line outside of a circle centered on the rectangle with diameter equal to the rectangle width, which doesn't do much for the tall and thin rectangles I'm using. So I tried rewriting this branch of disjoint: https://github.com/srossd/penrose/commit/4a8574a4bcad94d5773e4a7c8dfc70c180cdbc46#diff-e3f90ad59d8a07aa564e67ef3a[…]7f28e946545a651aac3e8da8227dR352
> 
> This does seem to work for the triple simple.sub/monoidal.sty/monoidal.dsl in that same commit. But when the line and rectangle are initially intersecting, the optimization step takes a noticeably long time. For anything more complicated, the optimizer crashes.
> 
> For my domain I don't really need a generic disjoint, so I'll just put that constraint back the way it was and write what I need a different way, but I was curious why this is happening?

**To Reproduce**

Run the code from this [commit](https://github.com/srossd/penrose/commit/4a8574a4bcad94d5773e4a7c8dfc70c180cdbc46#diff-e3f90ad59d8a07aa564e67ef3a[…]7f28e946545a651aac3e8da8227dR352) on the three monoidal programs in the `monoidal-domain` branch and folder.

**Expected behavior**
Line and rectangle are disjoint.

**Screenshots**
None yet.

**Additional context**
I also ran into this problem when writing the initial Style for `graph-domain`. This function (also discrete, and trying to accomplish basically the same thing) also caused the optimizer to hit a massive slowdown: https://github.com/penrose/penrose/blob/main/packages/core/src/contrib/Functions.ts#L327 (see also 2/25/21 meeting notes)

One thing to note is that this is a non-continuous function, and the optimizer hasn’t been tested super thoroughly with functions that involve discrete choices & conditionals (my guess is that it’s generating a very large computational graph, slowing down the process).

TODO: Read the code that Ross sent more carefully

hypotext:
Resolved in 2549fc3

Issue #496, opened 2021-02-25
write tests to check correctness of derivatives for discontinuous/boolean operations


k-qy:
Not super clear whether the generated gradient code properly differentiates through boolean and control flow statements:

* `ifCond` -- presumably the derivative branches and is either the derivative of left or right
* `min`, `max` -- presumably the derivative branches and is either the derivative of left or right
* `and`, `or`, `gt`, `lt`, `eq` -- presumably no derivative?

So far though, not a lot of code uses `ifCond` (which I believe had the wrong derivative earlier).

wodeni:
Bump: is this still relevant after #907?

samestep:
According to a conversation I just had with @keenancrane, the way we're currently doing this (including the new stuff I did in #906) is correct.

PR #538, opened 2021-04-07
feat: Get disjoint AABB rect/line constraints to work; make autodiff more correct; add varying initialization to Style

hypotext:
# Description

This PR makes the disjoint rect/line constraints work properly and account for rectangle dimensions, since our domain expert reported that this was not working for the diagrams they wanted to make. 

It also adds a new Style feature: varying variables that can be initialized by a custom value. This is helpful for testing the optimization with custom initial states.

Related issue/PR: #504 #496

Disjoint constraints are fixed by fixing the autodiff. Specifically, AABB (axis-aligned bounding box) rect-rect and rect-line disjoint and containment constraints now work. The bug was the `noGrad` should be `0`, not `1` which caused subtly wrong gradients for any non-differentiable op (`ifCond`, etc).

I added new examples that show that it scales reasonably well and works for the domain expert's application.

## Changes to implement new feature: specify the initial value of a varying variable in Style

- add `VaryingInit(i)` to the grammar as part of `AnnoFloat`
- in `substituteBlockExpr`, substitute out occurrences of `VARYING_INIT(i)` (the computation) for `VaryingInit(i)` (the `AnnoFloat`)
  - note that this is a hack; instead of shoehorning it into `substituteBlockExpr`, it should be done more cleanly as a compiler pass on the Style block AST at some point. doesn't really matter when this is done as long as it's before the varying float initialization in `genState`
- in `genState`, for shape properties, `VaryingInit(i)` is initialized as `i` (not sampled) and its path is included as a varying path
  - otherwise, for fields and access paths, it is initialized as the const val

`roger watch graph-domain/small-graph.sub graph-domain/disjoint-rects.sty graph-domain/graph-theory.dsl`

image

`roger watch graph-domain/small-graph.sub graph-domain/disjoint-rect-line-horiz.sty graph-domain/graph-theory.dsl`

image

(Added to both registry and wiki page, and the overall tests pass)

Also, the domain expert's example works with some amount of poor local minima which is solved by resampling. But this is my Style program, not theirs, so I would wait to see how they use it.

`roger watch monoidal-category-domain/test.sub monoidal-category-domain/monoidal-test.sty monoidal-category-domain/monoidal-test.dsl`

image

# Checklist

# Open questions

- [ ] Disjoint rect/line constraints should be unified as currently there are a bunch of different ones for different cases
- [ ] Move autodiff tests into test suite (will do this next)
- [ ] Implement rect-rect disjoint by penalizing area of overlap (as Keenan suggested)
- [ ] Get graph theory domain to work

## Possible issues/improvements for the varying initialization

- better if the parser parses a special syntax for this (like `? at i`) so we don't have to do the hacky `substituteBlockExpr` replacement
- `i` can only be a constant float, not an expression
- `i` could be outside the range that the variable is usually sampled in, resulting in incorrect or unsolvable problems
- not checking existence of VARYING_INIT_FN_NAME is a hack, would be removed w/ special syntax
- `VARYING_INIT(i)` will not behave properly if used in expressions like `VARYING_INIT(i) + VARYING_INIT(j)` because we don't look for varying paths in exprs. The evaluator will encounter it and throw an error.
- not super thoroughly tested
- maybe initialize in a range with a distribution, to be more general


packages/core/src/__testfixtures__/transformtest.input.ts
```diff
@@ -1,90 +1,90 @@
 export const objDict = {
```
wodeni, line 1: This could be the reason for CI failure. Consider reverting the changes in both files in this folder, unless they're intentional?

packages/core/src/contrib/Functions.ts
```diff
@@ -332,8 +332,10 @@ export const compDict = {
   },
 
   /**
-   * Figure out which side of the rectangle `[t1, s1]` the `start->end` line is hitting, assuming that `start` is located at the rect's center and `end` is located outside the rectangle, and return the size of the relevant side. Also assuming axis-aligned rectangle. This is used for arrow placement in box-and-arrow diagrams.
-   */
+     * Figure out which side of the rectangle `[t1, s1]` the `start->end` line is hitting, assuming that `start` is located at the rect's center and `end` is located outside the rectangle, and return the size of the OTHER side. Also assuming axis-aligned rectangle. This is used for arrow placement in box-and-arrow diagrams.
```
wodeni, line 335: Maybe mark it with `@deprecated`. Not super sure if tsdoc will catch it

wodeni (approved):
LGTM overall. Just need to fix the CI error. I believe @strout18 had the same issue in her PR. 

Given that this varying with initial value feature is still for debugging only, I think we can merge this branch for now and make it first-class if we decide to introduce it as an end-user construct later.

packages/core/src/__testfixtures__/transformtest.input.ts
hypotext: I reverted both files to Stella's original commit using this command: `git checkout db17b278bebe04f70d8e8c8135189d403f67aa2e -- transformtest.output.ts` and added them to the commit 933ac6ec62fdc6f4bbfab854f9d2aabfc55d0359.

The tests also all pass locally.

However, after the pre-commit hook, it appears that there are no changes made. My guess is that the pre-commit hook has been auto-formatting these files and adding whitespace that may interfere with the tests (though I'm not sure why the tests would work locally and not on CircleCI).

I could try to turn off the hook, but probably it would auto-format everyone else's commits for these files as well.

How should I proceed here?

packages/core/src/contrib/Functions.ts
hypotext: Done, thanks

packages/core/src/__testfixtures__/transformtest.input.ts
wodeni: Yep that sounds about right. Just add this folder to `prettierignore`

hypotext: Looks like it works in CircleCI now.

PR #539, opened 2021-04-07
feat: add Polygon and Polyline to the shape library

Dsm0:
I've added `Polygon`, `Polyline`, and `FreeformPolygon`, all up to Polygon/Polyline svg spec.

https://www.w3.org/TR/SVG2/shapes.html#PolygonElement


wodeni (changes requested):
LGTM. Remove `package.json` and we are good to go!

wodeni (approved):

PR #556, opened 2021-05-12
feat: Add callout shapes

hypotext:
# Description

Add callout shapes as requested by P3 in domain expert study.

Related issue/PR: none 

A callout supports the following parameters:

image

| Property      | Type    |
| ------------- | ------- |
| `anchor` | `vec2` (the black dot) |
| `center` | `vec2` (center of the contents) |
| `w` | `scalar` (width of the contents) |
| `h` | `scalar` (height of the contents) |
| `color` | `color` (fill color) |
| `padding` | `scalar` (padding around the contents, in pixels) |
| `strokeWidth` | `scalar` |
| `strokeColor` | `color` |

For best results, make sure that the anchor distance is less than ~`300` from the center. You can use a constraint to do so:

```
dist = norm(n.callout.center - n.anchor.center)
ensure lessThan(dist, 300.)
```

Added to shape docs [here](https://github.com/penrose/penrose/wiki/Shape-library#callout).

Callout shapes are drawn using a polygon (with `toScreen` transform) in the renderer, which uses all native JS types (lists of numbers) to do the calculations.

The polygon consists of a rectangle around the contents (with padding), plus three points that represent the anchor point, plus the intersection of a line from the anchor point to the center of the contents (extruded with a width that represents the width of the stem) as it intersects with the rectangle around the contents.

The two intersection points are calculated via segment-segment intersection.

Points are finally sorted by their angle from the center of the contents in order to give an ordered list of points for the polygon.

Repro: `roger watch shape.sub shape2.sty shape.dsl`

Visual documentation: `roger watch shape.sub shape.sty shape.dsl`

Examples:

image

# Checklist

Actually there are new warnings during the build step, not sure what to do with this:

```
@penrose/core:  > ../../node_modules/@thi.ng/api/assert.js:12:16: warning: Define "process.env.NODE_ENV" when bundling for the browser
@penrose/core:     12 │         return (process.env.NODE_ENV !== "production" ||
@penrose/core:        ╵                 ~~~~~~~~~~~~~~~~~~~~
@penrose/core:  > ../../node_modules/@thi.ng/api/expose.js:20:28: warning: Define "process.env.NODE_ENV" when bundling for the browser
@penrose/core:     20 │                     return (process.env.NODE_ENV !== "production" ||
@penrose/core:        ╵                             ~~~~~~~~~~~~~~~~~~~~
@penrose/core: ———————————————————————————————————————————————————————————
@penrose/core: TS: OK
@penrose/core: ———————————————————————————————————————————————————————————
@penrose/core: 2 warnings
```

# Open questions

- [ ] add rounded corners for callouts (use a path?)
- [ ] make bbox or bounding polyline available in backend for optimization
- [ ] draw the polyline in the backend? (requires `ifCond` that works on `VecAD`)
- [ ] shape looks bad on occasional samples. deal more gracefully with degenerate cases and parameters


hypotext:
hey @wodeni this is ready for a quick review! I know there are some things I need to clean up in the code or improve about it (e.g. using rect padding in the callout). Will make minor improvements later tonight. I'll also fill in the PR README later tonight. Lmk what you think, would like to merge by 10a tomorrow :-)

hypotext:
done with cleanup and documentation pass (cc @wodeni)

hypotext:
Oddly, when I try to do a clean squash and build with yarn, I run into this problem where it cannot find a library I imported (`@thi.ng/vectors`) despite it being present in `package.json`. Any idea why this might be happening?

```
yarn run v1.22.10
$ yarn build:parsers
$ nearleyc src/parser/Domain.ne > src/parser/DomainParser.ts && nearleyc src/parser/Substance.ne > src/parser/SubstanceParser.ts && nearleyc src/parser/Style.ne > src/parser/StyleParser.ts
$ rollup -c

./src/index.ts → build/dist/penrose.min.js...
[!] (plugin rpt2) Error: /Users/katherineye/Desktop/Code/penrose2/packages/core/src/renderer/Callout.ts(5,42): semantic error TS2307: Cannot find module '@thi.ng/vectors' or its corresponding type declarations.
src/renderer/Callout.ts
Error: /Users/katherineye/Desktop/Code/penrose2/packages/core/src/renderer/Callout.ts(5,42): semantic error TS2307: Cannot find module '@thi.ng/vectors' or its corresponding type declarations.
    at error (/Users/katherineye/Desktop/Code/penrose2/node_modules/rollup/dist/shared/rollup.js:5279:30)
    at throwPluginError (/Users/katherineye/Desktop/Code/penrose2/node_modules/rollup/dist/shared/rollup.js:18240:12)
    at Object.error (/Users/katherineye/Desktop/Code/penrose2/node_modules/rollup/dist/shared/rollup.js:18847:24)
    at Object.error (/Users/katherineye/Desktop/Code/penrose2/node_modules/rollup/dist/shared/rollup.js:18409:38)
    at RollupContext.error (/Users/katherineye/Desktop/Code/penrose2/packages/core/node_modules/rollup-plugin-typescript2/src/rollupcontext.ts:37:18)
    at /Users/katherineye/Desktop/Code/penrose2/packages/core/node_modules/rollup-plugin-typescript2/src/print-diagnostics.ts:41:11
    at arrayEach (/Users/katherineye/Desktop/Code/penrose2/packages/core/node_modules/rollup-plugin-typescript2/node_modules/lodash/lodash.js:516:11)
    at Function._.each [as forEach] (/Users/katherineye/Desktop/Code/penrose2/packages/core/node_modules/rollup-plugin-typescript2/node_modules/lodash/lodash.js:9368:14)
    at printDiagnostics (/Users/katherineye/Desktop/Code/penrose2/packages/core/node_modules/rollup-plugin-typescript2/src/print-diagnostics.ts:9:2)
    at Object.transform (/Users/katherineye/Desktop/Code/penrose2/packages/core/node_modules/rollup-plugin-typescript2/src/index.ts:244:5)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

packages/core/src/renderer/Callout.ts
```diff
@@ -0,0 +1,405 @@
...
+
+import { angleBetween2, normalize } from "@thi.ng/vectors";
+import { zipWith, reduce, sortBy } from "lodash";
+
+// TODO: Move some of the below functions to a numeric utils for frontend?
+
+/**
+ * Some vector operations that can be used on lists.
+ */
+export const ops = {
```
wodeni, line 13: I guess these functions are here because there's no way to reuse the autodiff ops (e.g. adding polymorphic behavior to those functions are too costly)? In general,a separate module would be good for these functions. I imagine other complex shapes will need these operations too.

packages/core/src/renderer/Callout.ts
```diff
@@ -0,0 +1,405 @@
+import { toScreen } from "utils/Util";
+import { attrFill, attrStroke, attrTitle } from "./AttrHelper";
+import { ShapeProps } from "./Renderer";
+
+import { angleBetween2, normalize } from "@thi.ng/vectors";
```
wodeni, line 5: Seems like these functions are not used. If not, consider removing this package from the dependency list.

packages/core/package.json
```diff
@@ -31,6 +31,7 @@
     "prepack": "yarn build"
   },
   "dependencies": {
+    "@thi.ng/vectors": "^6.0.0",
```
wodeni, line 34: Can you say a bit more about why this dependency is needed?

packages/core/src/renderer/Callout.ts
```diff
@@ -0,0 +1,405 @@
...
+
+  // [ ] <- rect center to middle of right segment is zero angle
+  const zeroAngleVec = ops.vdiv(
+    ops.vadd(cornersR.bottomRight, cornersR.topRight),
+    2
+  );
+
+  // Sort points by their angle to the center of the callout
+  const ptsSorted = sortBy(ptsR.concat(stemPts), (pt) =>
+    angleBetween2(
```
wodeni, line 371: I see the `vectors` dependency is used here (and only here?). Does the library have more vector ops, so you don't have to implement them above?  Currently, this file is a mixture of library and hand-written functions.

packages/core/src/contrib/Constraints.ts
```diff
@@ -890,6 +899,33 @@ const intersects = (
...
+// https://stackoverflow.com/questions/9043805/test-if-two-lines-intersect-javascript-function
+// https://stackoverflow.com/posts/58657254/revisions
+export const intersectionSegSeg = (s1: VarAD[][], s2: VarAD[][]): VarAD[] => {
+  const [l1_p1, l1_p2, l2_p1, l2_p2] = [s1[0], s1[1], s2[0], s2[1]];
+  const [[a, b], [c, d]] = [l1_p1, l1_p2];
+  const [[p, q], [r, s]] = [l2_p1, l2_p2];
+
+  // Divide by 0 error...
+
+  const det = sub(mul(sub(c, a), sub(s, q)), mul(sub(r, p), sub(d, b)));
```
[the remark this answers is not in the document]
hypotext: What do you mean by the "autodiff module"? Do you mean the functions module, so it's callable from Style?

packages/core/package.json
hypotext: I added it because I originally thought I was going to use it, but it turns out its API was so annoying that it was easier to port my own vector ops (you can't write `add(v1, v2)`, you have to write `add(result, v1, v2)`). What do you think, do we want to keep a real vector math library around in the frontend or just roll our own? Caveat, whatever I'm doing is probably not amazing for complicated numerics but it's just fine for the current use case.

packages/core/src/renderer/Callout.ts
hypotext: Yes, the library has more vectors, but I don't use it because (see prev comment). Probably the easier thing to do is to remove the dependency and just implement `angleBetween2`.

packages/core/src/renderer/Callout.ts
hypotext: Yeah, there's no way to make such ops nicely polymorphic because they have to call different functions depending on AD or not (e.g. AD vector `normalize` calls `add` whereas normal vector `normalize` calls the native JS `+`). I tried doing it all in AD but ran into #561, so had to do it in pure JS. I would be happy to move these into the utils module in the frontend.

packages/core/src/contrib/Constraints.ts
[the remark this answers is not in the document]
wodeni: I meant wherever `ops` are defined. I thought it was in `Autodiff.ts`?

packages/core/src/renderer/Callout.ts
wodeni: > I would be happy to move these into the utils module in the frontend.

Yeah that would be great!

packages/core/package.json
wodeni: I see. These impure APIs can be annoying and error prone. I was mainly being careful about adding dependencies. If there are only one or two functions to write, I'm more inclined to (a) find a better library and use it widely or (b) drop this library and write your own functions

packages/core/src/contrib/Constraints.ts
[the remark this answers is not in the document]
hypotext: `intersectionSegSeg` is a function, so it wouldn't be in `Autodiff` which is for "fundamental" operations on scalars and vectors (addition, etc)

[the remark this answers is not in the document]
wodeni: The line I commented on was where `det` was. I meant computing the determinant.

packages/core/src/renderer/Callout.ts
wodeni: Yep, that sounds good!

packages/core/src/contrib/Constraints.ts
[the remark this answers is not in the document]
hypotext: Well, this is a 2D determinant so it doesn't seem like it's a fundamental autodiff op either... but I can add a "determinant" of two vectors (i.e. the determinant of the matrix of the stacked column vectors) to the `ops` module.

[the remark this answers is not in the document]
wodeni: If you have the bandwidth for that it would be great!

hypotext:
@wodeni Ok, so I've addressed the PR comments in the last commit, but for some reason my pre-commit hook is not running (i.e. I just made my last commit with no checks). I'm in `packages/core`. I ran `git config --unset core.hookspath` and tried making a dummy commit, but it didn't turn the commit hook back on. Any idea how to fix this?

wodeni:
> @wodeni Ok, so I've addressed the PR comments in the last commit, but for some reason my pre-commit hook is not running (i.e. I just made my last commit with no checks). I'm in `packages/core`. I ran `git config --unset core.hookspath` and tried making a dummy commit, but it didn't turn the commit hook back on. Any idea how to fix this?

Just looked at the diff. Looks good to me overall. Not sure why husky is off all of the sudden. I can't repro this on my end. Try `yarn clean` and `yarn` again? If nothing works, my husky debugging knowledge basically came from places like this: https://github.com/typicode/husky/issues/326.

hypotext:
@wodeni What version of husky and yarn should we be using? Seems like a lot of the fixes involve upgrading/downgrading husky to different versions. 

Since this commit looks good to you, I'm going to merge for now, and look more into the husky problem later, as the issue you linked looks pretty lengthy.

Issue #570, opened 2021-05-26
update shape docs wiki page or autogenerate them


k-qy:
Currently they're very old.

Reported by domain expert P3 on 5/21/21

PR #580, opened 2021-06-03
refactor: shape property definitions and autodoc generate them

### LINKED ISSUE #558: Centralize shapeDef related objects
Simplifies the process of adding new shape definitions. Make a common interface that replaces IPropModel's values with more expressive optional `description`, `inputType`, `range`, etc keys. Deprecates the `mod` `.json` and viewBox switch statements/mappings so that the inspector works in a generic way.

maxkrieger:
closes #558, #570

- [x] Create new interface

```
export interface IProp {
  propType: PropType;
  sampler: Sampler;
  inspection?: IInputProps;
  description?: string;
}
```
- [ ] Port inspector `.json` properties
- [ ] Autodoc (need to find a good generator - maybe a gatsby or jekyll thing like https://pmarsceill.github.io/just-the-docs/ or https://github.com/rundocs/jekyll-rtd-theme or https://primer.style/doctocat/getting-started)

Note:

Typescript doesn't seem to catch the "old" way of accessing properties' samplers (eg `gpi[property][1]`), so runtime errors were thrown. I may have missed some old code paths where this is still the case - may need help finding them.

### DIFF
--- a/packages/browser-ui/src/inspector/views/mod/LabeledInput.tsx
+++ b/packages/browser-ui/src/inspector/views/mod/LabeledInput.tsx
@@ -22,7 +22,7 @@ type InputType =
 
 interface IInputProps {
   inputType: InputType;
-  showValue?: "true" | "false";
+  showValue?: boolean;
   min?: string;
   max?: string;
   minX?: string;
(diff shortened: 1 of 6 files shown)


maxkrieger:
Closing this due to irrelevance with new architecture. Using it as a reference though!

Issue #542, opened 2021-04-10
optimizer converges in diagrams where constraints are not fully satisfied


k-qy:
Investigate if these are local optima, or if not, what is a sensible policy here (are the constraints actually satisfiable?)

For example, in `roger watch monoidal-category-domain/test.sub monoidal-category-domain/monoidal-test.sty monoidal-category-domain/monoidal-test.dsl` on `graph-domain`, the optimizer converges here:

![image](https://user-images.githubusercontent.com/2762356/114254809-b9152880-997f-11eb-823a-50f9ca4d4aec.png)

You can see that several constraints are not satisfied (one is not pictured in the screenshot).

In other samples, all constraints are satisfied.

Optimizer converging on bad examples (maybe local minima, maybe not) is also a phenomenon Jiri reported.

PR #582, opened 2021-06-04
fix: Unify BBox Computation in `Constraints.ts` and Make `Square` Rect-Like

joshpoll:
# Description

Though `Square` is a rect-like shape, it does not expose the same interface of `w` and `h` like other rect-likes, which causes many constraint implementations to fail on squares.

This PR enhances the `bbox` function to compute `w`, `h`, and `center`.

Additionally, this PR moves the implementation of line-like bboxes into `bbox`. This also consolidates `disjointRectLineAAVert` and `disjointRectLineAAHoriz` into one constraint `disjointRectLineAA`.

# Checklist

### DIFF
--- a/packages/core/src/contrib/Constraints.ts
+++ b/packages/core/src/contrib/Constraints.ts
@@ -941,6 +941,9 @@ export const areDisjointBoxes = (a: BBox, b: BBox): VarAD => {
 };
 
 /**
+ * Preconditions:
+ *   If the input is line-like, it must be axis-aligned.
+ *   Assumes line-like shapes are longer than they are thick.
  * Input: A rect- or line-like shape.
  * Output: A new BBox
  * Errors: Throws an error if the input shape is not rect- or line-like.
@@ -957,7 +960,10 @@ export const bboxFromShape = (t: string, s: any): BBox => {
   if (t == "Square") {
     w = s.side.contents;
   } else if (isLinelike(t)) {
-    w = max(absVal(sub(s.start.contents[0], s.end.contents[0])), constOf(2));
+    w = max(
+      absVal(sub(s.start.contents[0], s.end.contents[0])),
+      s.thickness.contents
+    );
   } else {
     w = s.w.contents;
   }
@@ -966,7 +972,10 @@ export const bboxFromShape = (t: string, s: any): BBox => {
   if (t == "Square") {
     h = s.side.contents;
   } else if (isLinelike(t)) {
-    h = max(absVal(sub(s.start.contents[1], s.end.contents[1])), constOf(2));
+    h = max(
+      absVal(sub(s.start.contents[1], s.end.contents[1])),
+      s.thickness.contents
+    );
   } else {
     h = s.h.contents;
   }
--- a/packages/core/src/contrib/Constraints.ts
+++ b/packages/core/src/contrib/Constraints.ts
@@ -421,11 +421,7 @@ export const constrDict = {
     const overlapY = overlap1D(box.yRange, line.yRange);
 
     // Push away in both X and Y directions
-    return ifCond(
-      areDisjointBoxes(box, line),
-      constOf(0),
-      mul(overlapX, overlapY)
-    );
+    return mul(overlapX, overlapY);
   },
 
   /**
@@ -458,11 +454,7 @@ export const constrDict = {
       const overlapY = overlap1D(inflatedBox1.yRange, box2.yRange);
 
       // Push away in both X and Y directions, and account for padding
-      return ifCond(
-        areDisjointBoxes(inflatedBox1, box2),
-        constOf(0),
-        mul(overlapX, overlapY)
-      );
+      return mul(overlapX, overlapY);
     } else {
       // TODO (new case): I guess we might need Rectangle disjoint from polyline? Unless they repel each other?
       throw new Error(`${[t1, t2]} not supported for disjoint`);
--- a/packages/core/src/contrib/Constraints.ts
+++ b/packages/core/src/contrib/Constraints.ts
@@ -29,7 +29,7 @@ import { linePts } from "utils/OtherUtils";
 import { canvasSize } from "renderer/ShapeDef";
 import { VarAD, VecAD, Pt2 } from "types/ad";
 import { every } from "lodash";
-import { BBox } from "engine/BBox";
+import * as BBox from "engine/BBox";
 
 // Kinds of shapes
 /**
@@ -336,8 +336,8 @@ export const constrDict = {
 
       // TODO: There are a lot of individual functions added -- should we optimize them individually with a 'fnAnd` construct?
       return add(
-        constrDict.contains1D(box1.xRange, box2.xRange),
-        constrDict.contains1D(box1.yRange, box2.yRange)
+        constrDict.contains1D(BBox.xRange(box1), BBox.xRange(box2)),
+        constrDict.contains1D(BBox.yRange(box1), BBox.yRange(box2))
       );
     } else if (t1 === "Square" && isLinelike(t2)) {
       const [[startX, startY], [endX, endY]] = linePts(s2);
@@ -400,8 +400,8 @@ export const constrDict = {
 
     // Contains line both vertically and horizontally
     return add(
-      constrDict.contains1D(box.xRange, line.xRange),
-      constrDict.contains1D(box.yRange, line.yRange)
+      constrDict.contains1D(BBox.xRange(box), BBox.xRange(line)),
+      constrDict.contains1D(BBox.yRange(box), BBox.yRange(line))
     );
   },
 
@@ -417,8 +417,8 @@ export const constrDict = {
     const box = bboxFromShape(t1, s1);
     const line = bboxFromShape(t2, s2);
 
-    const overlapX = overlap1D(box.xRange, line.xRange);
-    const overlapY = overlap1D(box.yRange, line.yRange);
+    const overlapX = overlap1D(BBox.xRange(box), BBox.xRange(line));
+    const overlapY = overlap1D(BBox.yRange(box), BBox.yRange(line));
 
     // Push away in both X and Y directions
     return mul(overlapX, overlapY);
@@ -450,8 +450,8 @@ export const constrDict = {
       const box2 = bboxFromShape(t2, s2);
       const inflatedBox1 = BBox.inflate(box1, constOfIf(offset));
 
-      const overlapX = overlap1D(inflatedBox1.xRange, box2.xRange);
-      const overlapY = overlap1D(inflatedBox1.yRange, box2.yRange);
+      const overlapX = overlap1D(BBox.xRange(inflatedBox1), BBox.xRange(box2));
+      const overlapY = overlap1D(BBox.yRange(inflatedBox1), BBox.yRange(box2));
 
       // Push away in both X and Y directions, and account for padding
       return mul(overlapX, overlapY);
@@ -761,9 +761,15 @@ const pointInBox = (p: any, rect: any): boolean => {
  * compute the positive distance squared from point `p` to box `rect` (not the signed distance).
  * https://stackoverflow.com/questions/5254838/calculating-distance-between-a-point-and-a-rectangular-box-nearest-point
  */
-const dsqBP = (p: any, rect: any): VarAD => {
-  const dx = max(max(sub(rect.minX, p.x), constOf(0.0)), sub(p.x, rect.maxX));
-  const dy = max(max(sub(rect.minY, p.y), constOf(0.0)), sub(p.y, rect.maxY));
+const dsqBP = (p: any, rect: BBox.BBox): VarAD => {
+  const dx = max(
+    max(sub(BBox.minX(rect), p.x), constOf(0.0)),
+    sub(p.x, BBox.maxX(rect))
+  );
+  const dy = max(
+    max(sub(BBox.minY(rect), p.y), constOf(0.0)),
+    sub(p.y, BBox.maxY(rect))
+  );
   return add(squared(dx), squared(dy));
 };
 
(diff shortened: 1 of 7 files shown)


joshpoll:
One question I have is whether my use of nested ternaries here is best practice. I used it b/c it allows me to use `const` initialization, but it seems to sacrifice some readability.

packages/core/src/contrib/Constraints.ts
```diff
@@ -971,41 +936,42 @@ export const overlap1D = (
...
-    maxX: pts[0][0],
-    minY: pts[2][1],
-    maxY: pts[0][1],
-  };
+  // TODO: Compute the bbox of the line in a nicer way
 
-  return rect;
-};
+  // initialize center, w, and h depending on whether the input shape is line-like or rect/square-like
+  const center = isLinelike(t)
```
joshpoll, line 952: ternaries here. is it worth weakening/possible to weaken `const` to `let` and replace with if-else to improve readability?

packages/core/src/contrib/Constraints.ts
```diff
@@ -162,12 +168,10 @@ export const objDict = {
       );
 
       // entire equation is (mx - lx) ^ 2 + (my + 1.1 * text.h - ly) ^ 2 from Functions.hs - split it into two halves below for readability
-      const lh = squared(sub(mx, text.center.contents[0]));
```
hypotext, line 165: I would avoid using `bbox` here as well, since the text center and dimensions are available from the shape itself. In general, I would avoid unnecessary computation in an objective/constraint as it adds extra ops to the energy/gradient evaluation (which is the "hottest" path in the system).

packages/core/src/contrib/Constraints.ts
```diff
@@ -303,18 +308,20 @@ export const constrDict = {
       const res = sub(d, o);
       return res;
     } else if (t1 === "Circle" && isRectlike(t2)) {
-      const d = ops.vdist(fns.center(s1), fns.center(s2));
-      const textR = max(s2.w.contents, s2.h.contents);
+      const s2BBox = bbox([t2, s2]);
```
hypotext, line 311: Avoid bbox use?

packages/core/src/contrib/Constraints.ts
```diff
@@ -564,7 +529,8 @@ export const constrDict = {
     padding = 10
   ) => {
     if (isRectlike(t1) && t2 === "Circle") {
-      const textR = max(s1.w.contents, s1.h.contents);
+      const s1BBox = bbox([t1, s1]);
```
hypotext, line 532: Avoid using bbox if not necessary

packages/core/src/contrib/Constraints.ts
hypotext: Doesn't matter here, as using native JS ternaries is not an autodiff operation, and won't show up in the graph or during the optimization in runtime. (This is fine because a shape's type doesn't change during optimization.) I would go for the more readable option.

packages/core/src/contrib/Constraints.ts
```diff
@@ -971,41 +936,42 @@ export const overlap1D = (
...
+    t === "Square"
+      ? s.side.contents
+      : isLinelike(t)
+      ? max(absVal(sub(s.start.contents[1], s.end.contents[1])), constOf(2))
+      : s.h.contents;
 
-/**
- * Return the bounding box (as 4 segments) of an axis-aligned box-like shape given by `center`, width `w`, height `h` as an object with `top, bot, left, right`.
- */
-export const bboxSegs = (center: VecAD, w: VarAD, h: VarAD): any => {
```
hypotext, line 1002: Not totally sure about this design decision. I try to avoid catch-all utility functions as they do unnecessary computation, which gets added to the hot path of optimization. This amount of extra computing is probably fine, but in the future, I would prefer to split up functions into several small ones that can be called by constraint/objectives as needed.

hypotext (changes requested):
Thanks for doing this! Definitely, it's good to unify the rectlike shapes and simplify the operations on them. 

I would recommend making sure that your new functions still work in the first two Style programs here: https://github.com/penrose/penrose/wiki/Example-diagrams#disjoint-tests (especially WRT the max/min thing I pointed out) 

Also, I would avoid unnecessary computation (e.g. bbox).

packages/core/src/contrib/Constraints.ts
```diff
@@ -411,61 +411,25 @@ export const constrDict = {
...
 
+    // Push away in both X and Y directions
     return ifCond(
       areDisjointBoxes(box, line),
       constOf(0),
-      overlap1D(
-        [box.minX, box.maxX],
-        [s2.start.contents[0], s2.end.contents[0]]
-      )
+      min(overlapX, overlapY)
```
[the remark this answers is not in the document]
joshpoll: Yup this looks like a bug. Tried to figure out why I wrote that, and I actually copied it from the implementation of `disjoint`! Pushing a commit to fix both this function and `disjoint` (which now also incorporates offsets correctly). I'm changing both objectives to `add`, since this allows optimization in both the X and Y directions simultaneously and `overlapX` and `overlapY` are always non-negative by definition.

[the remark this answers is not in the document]
joshpoll: Incidentally this seems to fix the third disjoint test!! I can't seem to get it into the bad local minima it would often fall into before.

packages/core/src/contrib/Constraints.ts
joshpoll: This instance is necessary since `Square` does not have a `w` or `h` attribute.

packages/core/src/contrib/Constraints.ts
```diff
@@ -483,16 +447,17 @@ export const constrDict = {
       return sub(addN(o), d);
     } else if (isRectlike(t1) && isLinelike(t2)) {
       const [text, seg] = [s1, s2];
-      const centerT = fns.center(text);
+      const textBB = bbox([t1, s1]);
```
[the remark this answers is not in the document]
joshpoll: This is necessary because `w` is used on L453.

packages/core/src/contrib/Constraints.ts
joshpoll: This is necessary b/c Square does not have `w` or `h`.

packages/core/src/contrib/Constraints.ts
joshpoll: I refactored `BBox` into an immutable class that lazily computes `corners` as needed.

packages/core/src/contrib/Constraints.ts
joshpoll: `h` is used later, so we already need to construct a `BBox`.

packages/core/src/contrib/Constraints.ts
```diff
@@ -120,11 +124,13 @@ export const objDict = {
...
 
     if (isLinelike(t1) && isRectlike(t2) && isRectlike(t3)) {
+      const text1BB = bbox([t2, text1]);
+      const text2BB = bbox([t3, text2]);
       // HACK: Arbitrarily pick the height of the text
       // [spacing * getNum text1 "h", negate $ 2 * spacing * getNum text2 "h"]
-      return centerArrow2(arr, fns.center(text1), fns.center(text2), [
-        mul(spacing, text1.h.contents),
-        neg(mul(text2.h.contents, spacing)),
+      return centerArrow2(arr, text1BB.center, text2BB.center, [
```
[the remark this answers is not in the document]
joshpoll: `h` is used later so the bboxes are necessary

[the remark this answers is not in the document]
joshpoll: I have now renamed the variables to reflect the documentation.

packages/core/src/contrib/Constraints.ts
joshpoll: I replaced the ternaries with if-else.

joshpoll:
This PR will close #542 and #584.

packages/core/src/contrib/Constraints.ts
joshpoll: In the new version of `BBox` in cases where the input is a non-Square rect-like and the only fields used are `w`, `h`, and `center`, the overhead of the BBox is just creating a new BBox object. The constraint graph is not made more complex.

joshpoll:
fyi deleted comment was a copy of the resolved ternary conversation that I couldn't resolve

hypotext:
It looks like there's an ongoing CircleCI build problem that's keeping this PR from being merged (or rather, the CI build is just not running):

> ci/circleci: build Expected — Waiting for status to be reported

@wodeni I checked this page and didn't see JMP's branch there: https://app.circleci.com/pipelines/github/penrose/penrose Could it be that CI doesn't run on PRs for forks? Any idea how we might fix this?

joshpoll:
Seems to be a setting in Circle CI: https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories

hypotext:
Thanks, I just turned that on (assuming you are fine with @wodeni?). But I don't see the branch in CircleCI. 

Is the second question here relevant to you @joshpoll ? https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-

Alternatively, try pushing a no-op commit and see if it triggers it?

joshpoll:
seems to be running now! fingers crossed

wodeni:
> Thanks, I just turned that on (assuming you are fine with @wodeni?). But I don't see the branch in CircleCI.

Yeah, that sounds good. @joshpoll is the plan to keep working on a fork or this happens to be on a forked repo? You should have push access, so feel free to do a branch next time if you prefer.

joshpoll:
I plan to keep working on my fork if that's alright. I prefer to keep the main repo's branch collection as clean as (or cleaner than!) I found it.

packages/core/src/engine/BBox.ts
```diff
@@ -0,0 +1,142 @@
+import { Pt2, VarAD } from "types/ad";
+import { add, constOf, div, neg, ops } from "./Autodiff";
+
+export class BBox {
+  public readonly w: VarAD;
```
hypotext, line 5: Thanks for factoring this out, it is much cleaner! 

If the OO style used just for the purpose of lazily evaluating corners of the bbox? If so, I actually have to backtrack here... on second thought, unnecessary computation is not going to show up in the energy/gradient graph if it is not ultimately used in the energy. (Unnecessary computation is still a problem if it *is* ultimately used in the energy.) 

So this is not a problem, and I think a non-OO style might be cleaner as we don't use it elsewhere in the codebase (rather, we use typedefs paired with functions).

@wodeni curious about your thoughts on introducing OO style into the codebase as here?

hypotext (approved):
Fine to merge as soon as you are able to address my (very minor) comments!

packages/core/src/engine/BBox.ts
joshpoll: Yeah the purpose of OO here was just to have the lazy computation. It can be easily refactored to just a namespace if laziness isn't an issue here.

packages/core/src/contrib/Constraints.ts
```diff
@@ -162,12 +169,10 @@ export const objDict = {
...
       // entire equation is (mx - lx) ^ 2 + (my + 1.1 * text.h - ly) ^ 2 from Functions.hs - split it into two halves below for readability
-      const lh = squared(sub(mx, text.center.contents[0]));
+      const textBB = bboxFromShape(t2, text);
+      const lh = squared(sub(mx, textBB.center[0]));
       const rh = squared(
-        sub(
-          add(my, mul(text.h.contents, constOf(1.1))),
-          text.center.contents[1]
-        )
+        sub(add(my, mul(textBB.h, constOf(1.1))), textBB.center[1])
```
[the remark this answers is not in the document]
joshpoll: My pre-commit hook is running including auto-formatting and tests. I think this changed b/c the `contents` accesses have been removed. (They now happen during BBox creation.)

packages/core/src/engine/BBox.ts
hypotext: Yeah, if you don't mind, I'd prefer a namespace as it's cleaner

packages/core/src/contrib/Constraints.ts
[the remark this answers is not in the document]
hypotext: Ok, that seems fine then

packages/core/src/contrib/Constraints.ts
```diff
@@ -1053,3 +939,45 @@ export const areDisjointBoxes = (a: any, b: any): VarAD => {
...
+      `BBox expected a rect-like or line-like shape, but got ${t}`
+    );
+  }
+
+  // initialize w, h, and center depending on whether the input shape is line-like or rect/square-like
+  let w;
+  if (t == "Square") {
+    w = s.side.contents;
+  } else if (isLinelike(t)) {
+    w = max(absVal(sub(s.start.contents[0], s.end.contents[0])), constOf(2));
```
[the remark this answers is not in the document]
joshpoll: done!

packages/core/src/engine/BBox.ts
joshpoll: This is a very minor point, but as I'm doing the refactor I'm reminded that the other benefit of using a class is that you can write nontrivial getters that look like properties. In the namespace version you have to choose either (i) to pre-compute `corners`, which gives BBox a public `corners` property but confusingly things like `edges` are still functions or (ii) computing `corners` every time, but having a consistent interface for all but the basic three properties of `w`, `h`, and `center`. I'm going to do the second one, because I'm assuming the computation graph execution time dwarfs the execution time of re-constructing corners.

joshpoll: refactored!

hypotext:
Looks great, thanks for making the revisions!

hypotext:
@joshpoll Want to do the squash-and-merge? GitHub will give you a chance to clean up the commit msg if you so wish (optional).

joshpoll:
Awesome! I'm not authorized to merge PRs. Do I need that auth to do the squash-and-merge or can I squash the commits myself and then have you merge it?

hypotext:
No need to squash using git. You should be able to do it straight from GitHub UI. I added you as a repo contributor with write permissions. Are you able to do that now?

joshpoll:
Hmm that didn't seem to work. It looks like there is a branch-specific permission thing going on: https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule#editing-a-branch-protection-rule

hypotext:
Oh yeah, that's something @wodeni added. Let me consult with him.

hypotext:
Ok, @wodeni and I discussed and it seems better for now for us to handle merges to `main`.

PR #587, opened 2021-06-10
fix: arrowheads incorporated in the path length

### LINKED ISSUE #586: Line size does not account for arrowhead size.
**Describe the bug**

Arrowheads intersect other objects when lines are appropriately flush to objects.

Screen Shot 2021-06-10 at 2 33 07 PM copy

**To Reproduce**
This can be seen in the monoidal category example: `roger watch monoidal-category-domain/test.sub monoidal-category-domain/monoidal-test.sty monoidal-category-domain/monoidal-test.dsl`

**Expected behavior**
Arrowheads should not intersect.

joshpoll:
# Description

Closes #586.

I extended the Line renderer in `renderer/Line.ts` to compute the heights of the arrowheads and shrink the main path accordingly. To compute the height of the arrowhead, you multiply the arrowhead style's height by the size specified in Penrose. This value controls the `markerHeight` of the arrowhead. You then have to multiply by the unit for the markerHeight, which is the `stroke-width` of the line. See https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/markerUnits for details.

To compute the amount of distance to shave off the ends of the path I used the formula in this Math Stack Exchange post: https://math.stackexchange.com/a/2045181. It is fairly straightforward: Compute the vector of the difference between the points, then shave off a fraction of that vector from each endpoint as needed.

I used the monoidal category domain with the following files to generate the before and after screenshots.

`monoidal-category-domain/test.sub`
`monoidal-category-domain/monoidal-test.sty`
`monoidal-category-domain/monoidal-test.dsl`

Before:
Screen Shot 2021-06-10 at 5 00 44 PM

After:
Screen Shot 2021-06-10 at 5 28 47 PM

Notice that arrowheads no longer intersect components!

# Checklist

### DIFF
--- a/packages/core/src/contrib/Constraints.ts
+++ b/packages/core/src/contrib/Constraints.ts
@@ -41,7 +41,7 @@ export const isRectlike = (shapeType: string): boolean => {
 };
 
 /**
- * Takes a `shapeType`, returns whether it's linelike. (TODO: Account for arrowhead size)
+ * Takes a `shapeType`, returns whether it's linelike.
  */
 export const isLinelike = (shapeType: string): boolean => {
   return shapeType == "Line" || shapeType == "Arrow";
--- a/packages/core/src/renderer/Arrow.ts
+++ b/packages/core/src/renderer/Arrow.ts
@@ -1,3 +1,4 @@
+import { Shape } from "types/shape";
 import { IFloatV, IStrV, IColorV, IVectorV } from "types/value";
 import { arrowheads, round2, toHex, toScreen } from "utils/Util";
 import { attrFill, attrTitle, DASH_ARRAY } from "./AttrHelper";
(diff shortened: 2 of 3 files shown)


wodeni (commented):
Nice! The implementation looks good to me. BTW the `Arrow` shape is doing something similar, but probably not robust as this. Maybe it's better to use `makeRoomForArrows` in here as well: https://github.com/penrose/penrose/blob/09062eee7bd027396905958cf009305fcc8aa6f6/packages/core/src/renderer/Arrow.ts#L56?

joshpoll:
Wait... why does `Arrow` exist lol it's basically line, but with the arrowhead direction specified, right?

joshpoll:
I will move `makeRoomForArrows` to `Arrow` like `arrowHead` is currently, since consolidating the two shapes is worthy of a separate PR (if it even needs to change).

maxkrieger (approved):

hypotext:
@wodeni good to merge?

wodeni (approved):

PR #590, opened 2021-06-11
fix: type-change in synthesizer was adding improperly formed bind statements to AST

hsharriman:
# Description
Fixes a bug where `typeChange` mutation in the synthesizer module was adding undefined statements to the AST, because the way new `Bind` statements were being created was improper. 

* Refactored `Synthesizer` code to use `generateConstructor`, `generateFunction` and `generatePredicate` functions.
  * Added option for each of these methods to take in a pre-selected declaration instead of randomly picking one (which will also help out with testing later on)
* Refactored so that `Add/Edit/Delete` mutation operations are added by the `Synthesizer` instead of by `SynthesizerContext`.

Validate correct functionality by running `yarn set-example` or `yarn tri-example` from `packages/synthesizer`
# Checklist

# Open questions

### DIFF
--- a/packages/core/src/analysis/SubstanceAnalysis.ts
+++ b/packages/core/src/analysis/SubstanceAnalysis.ts
@@ -1,4 +1,4 @@
-import { pullAt } from "lodash";
+import { pullAt, range } from "lodash";
 import { Identifier } from "types/ast";
 import { Map } from "immutable";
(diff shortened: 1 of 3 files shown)


wodeni (approved):
LGTM! Thanks for pushing the fix :D

PR #591, opened 2021-06-14
refactor: add determinism to synthesizer

wodeni:
# Description

Related issue/PR: #590 

Continuing the effort to localize calls to random generators in #590, I refactored the code base to only use random generators within the `Synthesizer` class. This will be useful for testing and some other future features to come.

* Replaced the default `choice` and `random` functions from `pandemonium` with custom ones using their `createChoice` and `createRandom` constructors. 
* The new custom `choice` and `random` functions can only be called within the `Synthesizer` class
* Added `seed` to the constructor arguments of `Synthesizer`, which defaults to `synthesizerSeed `

In `packages/synthesizer/index.ts`, add a fourth argument to `new Synthesizer(...)` and the results will change depending on your random seed. 

# Checklist

### DIFF
--- a/packages/core/src/synthesis/Synthesizer.test.ts
+++ b/packages/core/src/synthesis/Synthesizer.test.ts
@@ -46,7 +46,7 @@ const initSynth = (
   if (subRes.isOk()) {
     subResult = subRes.value;
   }
-  const synth = new Synthesizer(env, setting, subResult);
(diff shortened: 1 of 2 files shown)


hsharriman (approved):
LGTM! Glad we should be able to test the synthesizer now

PR #600, opened 2021-06-15
feat: existential graph domain

wodeni:
# Description

This PR includes some progress on building the existential graph domain (in collaboration with domain experts). Although we will keep building this domain, we will merge the progress so far into `main` so that some of the functions in `contrib` are available on the IDE.

* `cd examples/existential-graph-domain`
* `roger watch graph.sub style.sty existential-graph.dsl`


packages/core/src/contrib/Constraints.ts
```diff
@@ -116,7 +116,7 @@ export const objDict = {
     [t2, text1]: [string, any],
     [t3, text2]: [string, any]
   ): VarAD => {
-    const spacing = varOf(1.1); // arbitrary
+    const spacing = varOf(2.0); // arbitrary
```
hypotext, line 119: Might this cause regressions/changes in other examples?

hypotext:
Looks fine to merge (just see one comment above for something to check).

packages/core/src/contrib/Constraints.ts
wodeni: Umm this should really be an argument of the function. Anyway, this function is not used in this domain. Reverting this change.

PR #601, opened 2021-06-18
feat: Substance mutations as data + refactored program generator

wodeni:
# Description

This PR introduces the `Mutation` module, which contains all the Substance mutation types and validation functions. The main improvement is that mutations are no longer executed inline in `Synthesizer` anymore. Instead, mutations are encoded as data and can be executed individually (`executeMutation`) or in a batch (`executeMutations`).

Aside from the refactoring, this branch also contains the `Search` module, which are experimental functions for by-example synthesis. This module is currently unused.

* `Mutation` type: every type of mutation (e.g. `Add`, `Delete`, `ChangeStmtType` etc) has its own attributes enough for executing it. For instance, `SwapStmtArgs` has the target statement and the indices to swap. A mutation can also have `additionalMutations`, which are related mutations.
* Guard functions: there are two ways to construct a mutation: (1) if you have exactly the right parameters in the right types, you can directly construct the object of one of the `Mutation` types; (2) if you only have a vague target such as `SubStmt` (which might be something like `AutoLabel All` and not a valid target for any of the mutations), you can use the `check*` guard functions to check if a certain kind of mutation is possible for the given target.
* Callbacks in guard functions: each guard function has a callback function, which normally takes in a validated target and outputs a mutation. When you are calling the guard function to check if you can apply a mutation, you need to also supply a callback for generating the components for the replacement.
* The program generator: now the `Synthesizer` class mutates Substance programs by enumerating `Add`, `Delete`, and `Update` mutations and randomly selecting one of them in each step.
* Refactored `generate*` functions: instead of building up a Substance program through `Synthesizer`'s internal fields opaquely, these generator functions return one or more Substance statements, which is later converted into `Add` mutations by `Synthesizer`. 

TODO

# Checklist

# Open questions

* ~~`generate*` functions: the `SynthesizerContext` passed into them is still mutated instead of returned. Should we make them completely pure instead?~~
  * ~~One known problem with the current approach is: these functions mutate the context when generating mutations that might not selected at the end, so you'll see extra ids even if the mutations are never executed :(.~~
  * - [x] RESOLVED: implemented pure context passing in f68d19e
* `Synthesizer` doesn't enumerate all the mutations. Instead, it's still making some random choices such as picking a particular statement to edit.
* `weights` is still not used in the system for picking among `Add`, `Delete` and `Update`.


packages/core/src/synthesis/Synthesizer.ts
```diff
@@ -375,238 +352,308 @@ export class Synthesizer {
...
-      tag: "Replace",
-      old: oldStmt,
-      new: newStmt,
-      mutationType: "TypeChange",
-    });
+  /**
+   * Pick a random statement in the Substance program and enumerate all the applicable mutations.
+   * @returns a list of mutation groups, each representing a series of `Update` mutations
+   */
+  editStmt = (ctx: SynthesisContext): MutationGroup[] => {
```
hsharriman, line 533: nit: potentially consider renaming this function as well as the `addStmt` and `deleteStmt` functions to clarify that they are generating the list of all possible mutations for the chosen statement, not making any changes to the AST. Something like `editCandidates` or `buildEditGroup` or `buildEditCandidates`, etc.

packages/core/src/synthesis/Mutation.ts
```diff
@@ -0,0 +1,427 @@
...
+): Add[] | undefined => {
+  const stmts: SubStmt[] = newStmts(cxt);
+  return stmts.map((stmt: SubStmt) => ({
+    tag: "Add",
+    stmt,
+    mutate: appendStmtCtx,
+  }));
+};
+
+export const checkAddStmt = (
```
hsharriman, line 197: I'm only seeing this function being used in one place, it seems like it could just be entirely replaced by `checkAddStmts`

packages/core/src/synthesis/Synthesizer.ts
```diff
@@ -624,143 +671,243 @@ export class Synthesizer {
...
+      func.args,
+      ctx
+    );
     const rhs: ApplyFunction = applyFunction(func, args);
+    // find the `TypeDecl` for the output type
     const outputType = func.output.type as TypeConstructor;
+    // NOTE: the below will bypass the candidate list and generate a new decl, to follow the config more strictly:
+    // const outputTypeDecl: TypeDecl | undefined = ctx.candidates.types.find(
+    //   (decl, typeName) => typeName === outputType.name.value
+    // );
```
hsharriman, line 733: nit: in practice, is there a reason to keep this?

packages/core/src/synthesis/Synthesizer.ts
```diff
@@ -624,143 +671,243 @@ export class Synthesizer {
...
+      ctx
+    );
     const rhs: ApplyConstructor = applyConstructor(cons, args);
     const outputType = cons.output.type as TypeConstructor;
-    // TODO: choose between generating vs. reusing
-    const lhs: Identifier = this.generateType(outputType.name).name;
+    // NOTE: the below will bypass the candidate list and generate a new decl, to follow the config more strictly:
+    // const outputTypeDecl: TypeDecl | undefined = this.ctx.candidates.types.find(
+    //   (decl, typeName) => typeName === outputType.name.value
+    // );
```
hsharriman, line 757: nit: same question as above, is there a reason to keep this here?

packages/core/src/synthesis/Search.ts
```diff
@@ -0,0 +1,395 @@
+import {
```
hsharriman, line 1: Since this code is WIP I didn't really take a close look at this file, lmk if you would like me to look at anything in particular!

hsharriman (approved):
LGTM!  These changes have definitely made the code a lot more modular which is great, and I can see that inserting a new mutation will take only a modest amount of work now. Looking forward to trying it out when I implement the new swap-in mutator! I left some comments--basically all nits--on the PR.

packages/core/src/synthesis/Search.test.ts
```diff
@@ -0,0 +1,262 @@
...
+    Set A, B, C, D, E, F, Z
+    IsSubset(B, A)
+    IsSubset(D, A)
+    F := Union(A, B)
+    `;
+    const res1: SubRes = getSubRes(domainSrc, prog1);
+    const ast1: SubProg = res1[0].ast;
+    const ast2: SubProg = getSubRes(domainSrc, prog2)[0].ast;
+    const diffs: StmtDiff[] = diffSubStmts(ast1, ast2);
+    expect(diffs).toHaveLength(2);
```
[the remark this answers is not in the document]
wodeni: Good point. The tests are still kind of WIP at this stage. I'll leave them as-is for now.

packages/core/src/synthesis/Mutation.ts
```diff
@@ -0,0 +1,427 @@
...
+): ChangeExprType | undefined => {
+  if (stmt.tag === "Bind") {
+    const { expr } = stmt;
+    if (
+      expr.tag === "ApplyConstructor" ||
+      expr.tag === "ApplyFunction" ||
+      expr.tag === "Func"
+    ) {
+      const res = getMutations(stmt, expr, cxt);
+      if (res) {
```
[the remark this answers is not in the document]
wodeni: I refactored `ChangeType`, but not `ReplaceName`, as the latter depends on the structure of the old statement more.

packages/core/src/synthesis/Synthesizer.ts
```diff
@@ -375,238 +352,308 @@ export class Synthesizer {
...
+            if (pick.tag === "PredicateDecl") {
+              newStmts = this.generatePredicate(pick, ctx);
+            } else if (pick.tag === "FunctionDecl") {
+              newStmts = this.generateFunction(pick, ctx);
+            } else {
+              newStmts = this.generateConstructor(pick, ctx);
             }
-          }
-          break;
+            const { res, stmts } = newStmts;
```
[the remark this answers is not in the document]
wodeni: Yep! See `generateArgStmt`

packages/core/src/synthesis/Synthesizer.ts
```diff
@@ -375,238 +352,308 @@ export class Synthesizer {
...
+            }));
+            const addOps: Add[] = stmts.map((stmt) => ({
+              tag: "Add",
+              stmt,
+              mutate: appendStmtCtx,
+            }));
+            return {
+              newStmt: res,
+              additionalMutations: [...deleteOps, ...addOps],
+            };
```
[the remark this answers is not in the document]
wodeni: Created action constructors for `Add` and `Delete`. Much cleaner now!

packages/core/src/synthesis/Synthesizer.ts
wodeni: Yeah, they are really enumerating things. Changed the names to reflect this fact.

packages/core/src/synthesis/Mutation.ts
wodeni: Leaving it in, just in case future synthesizers will use this guard function.

packages/core/src/synthesis/Synthesizer.ts
wodeni: Deleting the code as it's outdated anyway. Still keeping the `NOTE:` so we are aware of this and may fix this later.

packages/core/src/synthesis/Search.ts
wodeni: Nothing for now! Will continue this effort in another PR :D.

packages/core/src/index.ts
```diff
@@ -27,7 +27,7 @@ import { andThen, Result, showError } from "utils/Error";
 import { prettyPrintFn } from "utils/OtherUtils";
 import { bBoxDims, toHex } from "utils/Util";
 import { Canvas } from "renderer/ShapeDef";
-import { showOps } from "synthesis/Mutation";
+import { showMutations } from "synthesis/Mutation";
```
hsharriman, line 30:
```
index.ts:4:3 - error TS2305: Module '"../../node_modules/@penrose/core/build/dist"' has no exported member 'showMutations'.

4   showMutations,
```
I'm getting this error message here, so I cannot check the existing example functionality at the moment. I'm not quite sure if it has anything to do with changes made on another branch or not though.

hsharriman (approved):
LGTM overall. I was unable to run the new examples on my own machine because of import issues from penrose, but I'm not sure if that's an issue with my machine at the moment or not.

Issue #602, opened 2021-06-21
add arc command to Path shape spec


hsharriman:
**Is your feature request related to a problem? Please describe.**
Another way to make curves in SVG that doesn't involve beziers is use of the arc command, well-illustrated towards the bottom of the linked document: https://css-tricks.com/svg-path-syntax-illustrated-guide/).  Penrose currently has no way to render an arc with this command. Trying to make a perfectly regular arc programmatically using bezier curves is quite complicated and often results in strange results. Additionally, since all of the other types of SVG curves are implemented besides the arc, it makes sense to add this one to complete the set. 
**Describe the solution you'd like**
Add functionality to the `Path` shape spec to understand and create arcs in SVG.

**Describe alternatives you've considered**
Using bezier curves only, it quickly becomes quite tedious to implement logic for the control points of bezier curves when all that is needed is a slightly rotated arc/semicircle

**Additional context**
For reference I'm trying to create these types of marks in Penrose:
![Screenshot from 2021-06-21 14-08-18](https://user-images.githubusercontent.com/31522997/122840667-72ad4580-d2af-11eb-9206-fef40ec4dd92.png)
I have a PR up that adds the aforementioned arc functionality, so I've only created this issue for additional clarity.

PR #603, opened 2021-06-22
feat: add arc command to path shape object, implement tick marks for triangle congruency

hsharriman:
# Description
Screenshot from 2021-06-21 14-08-18
Added ability to represent angle and segment length equality in Euclidean style within Penrose.
Related issue/PR: #602 
* Feature request to add the SVG arc specification to the existing Path spec in Penrose. This was a necessary prerequisite for getting the angle markers to render.

* At this point in time, I tried to fit my code within the existing system. As such, there are a couple of places where the code seems to take the "hacky" approach (esp in `EngineUtils.ts`). This is a short-term solution that should be amended when a larger refactor is in the works.
* Where possible, the functions added to `Functions.ts` were designed for modularity, so sometimes arguments that could just be constants for my particular use case are passed in as parameters (especially with regards to the `arc` function).
* Instead of having one predicate, `EqualLength` that makes 2 segments equal length and adds a tick mark, I created 2 predicates, `EqualLength` (makes the segments the same length) and `EqualLengthMarker` (adds a tick mark to the 2 segments). This was done for the sake of the synthesizer, where in some cases it will be advantageous to have 2 segments that appear equal but are not actually defined as such (to be defined as equal they would need the markers), and vice versa. The same decision was made for `EqualAngle`, `EqualAngleMarker1` and `EqualAngleMarker2`
* `EqualAngleMarker1` and `EqualAngleMarker2` both exist for the time being because there is currently no way to track how many times a given predicate is called in a given diagram. So you cannot do something like:
```[typescript]
for each Angle a, b
where EqualAngleMarker(a,b) {
  --make count tickmarks
  count++
}
```
In geometry, the first set of angles that are equal get 1 tick, the second set gets 2, and so on. However, since it is uncommon to see more than 2 pairs of equal angles in a standard geometry problem, I temporarily elected to just make 2 separate predicates.
* run `yarn start` from root 
* open `localhost` in your browser
* in a new terminal run `npx roger watch packages/synthesizer/__tests__/congruent.sub packages/synthesizer/__tests__/euclidean.sty  packages/synthesizer/__tests__/geometry.dsl`
# Checklist

# Open questions
* Currently I have these 3 files in a folder inside of the synthesizer package, should I move them to the `examples` folder? I can either add new versions or overwrite the existing `geometry/geometry.dsl` and `geometry/euclidean-simple.sty` files
* This example takes quite some time to optimize (~10-15seconds). @wodeni and I confirmed that it was not due to code-generation, the optimization process takes a while and several of the constraints are unsatisfied:
Screenshot from 2021-06-21 17-39-19


packages/core/src/contrib/Functions.ts
```diff
@@ -357,20 +362,86 @@ export const compDict = {
...
+    };
+  },
+  /**
+   * Find intersection between a circle centered at [x1, y1] with radius r and a line segment between [x1, y1] and [x2, y2].
+   * @param x1, y1: centerpoint of circle, one endpoint of line segment
+   * @param x2, y2: endpoint of line segment
+   * @param r: radius of circle
+   * @returns: vector representation of the point of intersection
+   */
+  angleMarker: (
```
wodeni, line 409: My interpretation is this function computes a point along a line segment `[(x1, y1), (x2, y2)]` that is `r` away from `(x1, y1)`? If so, you can actually find the unit vector of the line segment and use `ops.vmove(start, r, dir)` to find the point.

packages/core/src/engine/EngineUtils.ts
```diff
@@ -201,6 +201,17 @@ function mapElem<T, S>(f: (arg: T) => S, e: Elem<T>): Elem<S> {
       tag: e.tag,
       contents: mapTup3((x) => mapTup2(f, x), e.contents),
     };
+  } else if (e.tag === "Arc") {
+    const l: [[S, S], [S, S, S], [S, S]] = [
+      // TODO sad :(
```
wodeni, line 206: The `map*` functions in this module seem to be very redundant and not easy to maintain. My understanding is that they are only here for converting autodiff numbers to JS numbers? I was wondering if there's a better way to unpack autodiff numbers so we can avoid boilerplate like this @hypotext? Probably some kind of dynamic dispatch?

packages/synthesizer/__tests__/euclidean.sty
```diff
@@ -138,6 +147,110 @@ with Point p; Point q {
...
+where EqualAngleMarker2(a, b) {
+  startA = angleMarker(a.q, a.p, a.radius)
+  endA = angleMarker(a.q, a.r, a.radius)
+  sweepA = arcSweepFlag(a.q, startA, endA)
+
+  startB = angleMarker(b.q, b.p, b.radius)
+  endB = angleMarker(b.q, b.r, b.radius)
+  sweepB = arcSweepFlag(b.q, startB, endB)
+
+  a.angleMark = Path {
```
wodeni, line 212: This is not a change request: it would be great if we can have some concept of grouping or an easy API to compose multiple paths. You could technically pack the two arcs into one `Path` shape by having 2 `Elem`s in the `pathData`, but since there's no native syntax or functions for constructing complex paths, @hsharriman has to repeatedly declare multiple `Path`s for this.

wodeni (approved):

wodeni (approved):

PR #606, opened 2021-06-24
feat: add right angle marker to euclidean style

hsharriman:
# Description
Screenshot from 2021-06-22 14-44-23

* Added ability to render right angle markers for Euclidean angles, using `RightMarked(a)` predicate
* Added ability to make an angle appear as though it is a right angle, without the marker, using the `RightUnmarked(a)` predicate
* Fixed a bug in the `Synthesizer` that was caused by the `Swap` mutation attempting to swap arguments in functions/predicates/constructors that only have one argument
* Added an `unordered_collinear` constraint to `Constraints.ts` as an option to make points collinear without needing to specify their order in substance. for ex the relevant substance for this diagram was `Collinear(B,C,D)`:
Screenshot from 2021-06-23 15-47-06
With the `collinear` constraint function, the points would be ordered as `B,C,D`, with `C` in the middle. However, since I used `collinear_unordered`, the points are ordered as `C,B,D`. We are keeping both options because the original collinear produces reliable results, while the unordered version adds additional interesting challenges/mutations (esp. for program synthesis), and is more true to the geometric definition of collinear.

* `yarn start` from root
* View single example problem with `npx roger watch packages/synthesizer/__tests__/c4p1.sub packages/synthesizer/__tests__/euclidean.sty  packages/synthesizer/__tests__/geometry.dsl`
To run the synthesizer:
* `cd packages/synthesizer` and run `yarn congruent-example`
* `cd packages/automator` and run `yarn generate-site`
* view `packages/automator/browser/index.html` in your browser 
# Checklist

- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] My changes generate no new warnings
- [x ] New and existing tests pass locally using `yarn test`
- [x ] I ran `yarn docs` and there were no errors when generating the HTML site
- [x ] My code follows the style guidelines of this project (e.g.: no ESLint warnings)

# Open questions

### DIFF
--- a/packages/core/src/contrib/Constraints.ts
+++ b/packages/core/src/contrib/Constraints.ts
@@ -658,7 +658,7 @@ export const constrDict = {
   /**
    * Require that the `center`s of three shapes to be collinear. Does not enforce a specific ordering of points, instead it takes the arrangement of points that is most easily satisfiable.
    */
-  collinear_unordered: (
+  collinearUnordered: (
     [, p0]: [string, any],
     [, p1]: [string, any],
     [, p2]: [string, any]
--- a/packages/core/src/contrib/Constraints.ts
+++ b/packages/core/src/contrib/Constraints.ts
@@ -654,6 +654,36 @@ export const constrDict = {
       throw new Error("collinear: all input shapes need to have centers");
     }
   },
+
+  /**
+   * Require that the `center`s of three shapes to be collinear. Does not enforce a specific ordering of points, instead it takes the arrangement of points that is most easily satisfiable.
+   */
+  collinear_unordered: (
+    [, p0]: [string, any],
+    [, p1]: [string, any],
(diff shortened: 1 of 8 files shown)


packages/core/src/contrib/Constraints.ts
```diff
@@ -654,6 +654,36 @@ export const constrDict = {
       throw new Error("collinear: all input shapes need to have centers");
     }
   },
+
+  /**
+   * Require that the `center`s of three shapes to be collinear. Does not enforce a specific ordering of points, instead it takes the arrangement of points that is most easily satisfiable.
+   */
+  collinear_unordered: (
```
wodeni, line 661: Just for uniformity, let's use camel case instead of K&R here.

wodeni (approved):
LGTM!

Issue #42, opened 2017-12-05
Staged diagrams


k-qy:
Add language constructs for specifying the staging, or order, or a diagram, as it might be used in a talk, animation, or presentation.

Comment from Keenan: Would also be good to infer, as much as possible, a default ordering (e.g., based on logical dependencies).

Also we need to allow users to specify diagram display / layering order (so shapes with transparency don't display on top of labels, and so on).

hypotext:
Keenan adds:

> Basically we haven't addressed the temporal aspect of Penrose.  The "worst solution except for all the rest" is WYSIWYG slideshow editors like PowerPoint and Keynote, which enable one to specify an ordering for events, and associate transitions with these orderings (which can also include specific timing directives).  But just like laying out diagrams spatially, "laying out" diagrams temporally takes a lot of tedious, manual labor. [Personal note: I have spent hundreds of hours doing this in Keynote, and it is hell.]  To add insult to injury, the strict ordering required by this paradigm puts artificial limitations on how staged diagrams can be "timed."  A more powerful alternative is the timeline-based editing found in 2D packages like Adobe [Premiere|AfterEffects|Animate] or 3D packages like Blender and Maya.  But this again requires meticulous hand-timing—even more than slideshow editors.  
> 
> So, the main question is how to build the existing utility of Penrose out into the temporal domain.  For instance, it makes a lot of sense to annotate Style rules with transition times and styles (e.g., a new point added to a diagram should fade in for 0.35 seconds).  A very simple temporal ordering scheme could mimic the spatial ordering scheme we already have via the layer above/below keywords (so, perhaps, a show before/after  construct).  But then comes the question of how to deal with more complex synchronous events that can't be expressed via an ordering.  For instance, one could expand these semantics into starts before and ends after.  Etc.  These are all "seed suggestions"—basically the point is that this design really needs to be fleshed out to be both simple but also general.  I.e., it should line up with the same Penrose design goals stated in the paper, but just along the time axis.  A beautiful thing that can happen here is that values like transition times and delay times can be optimized by the Penrose layout engine. Just as we currently might pack a set of graphical primitives onto the canvas, we can "pack" a set of events into a given interval of time, but still allow the programmer to specify which ones should be shorter/longer in a relative sense.  [This is again something that I have tweaked meticulously by hand, and again, it is hell.]

[...]

PR #610, opened 2021-06-25
feat: staged diagram generation in automator

heleaf:
Screen Shot 2021-06-28 at 12 46 56 PM
Screen Shot 2021-06-28 at 12 54 53 PM

# Description

Related issue/PR: #42 

Change Summary: 
- Added a new command to `automator` that allows for command-line SVG generation of a single sub/sty/dsl trio.
- Added an option to `automator` that allows for command-line generation of staged diagrams (as multiple SVG files) of sub/sty/dsl trios. 

Motivation:
- `automator` was missing functionality to generate SVG(s) for a single sub/sty/dsl trio (previously mostly working on batch sub/sty/dsl) 
- Staged diagrams have been requested alongside the final diagram itself by users 

Dependencies: nothing new

- Changes to `automator` can be found in `index.tsx`.
- The functions used to generate states for staged diagrams can be found in `Comic.ts` under `core/src/renderer`.

### Staged diagram generation functions
- The staged diagram states are determined as a post-processing calculation (i.e. after the state of the original sub/sty/dsl has been fully optimized). Each state has certain missing shapes from their shapelist according to the order in which the user declared GPI primitives in their substance program. 
- The main function for creating the list of states is `getListOfStagedStates` in `Comic.ts`. 
- `getListOfStagedStates` creates a list of objects with FGPI properties from a state's `translation.trMap` and uses that list to determine what GPIs should remain in a state's shapelist for a list of staged states. 
- The order in which shapes get "added" state-by-state depends on the order in which the user defines GPI-tagged objects in substance. Note: whether or not substance commands like `Set A` and `IsSubset(B,A)` have GPI properties or not depends on the definitions given in style. For example, if the user's substance program is `Set A, B, C, IsSubset(B,A)` and their style program associates each `Set` with a `Text` label, and `IsSubset` with an `Arrow`,  then the shapelists in the list of generated states would look something like `[[Text for Set A], [Text for Set A, Text for Set B], [Text for Set A, Text for Set B, Text for Set C], [Text for Set A, Text for Set B, Text for Set C, Arrow for IsSubset(B,A)]]`.
Screen Shot 2021-06-28 at 12 47 34 PM

- When called in `automator`'s `index.tsx`, this function is used to generate multiple SVG files, corresponding to each state in the outputted list of states. 

### `automator` changes
- A new command line call (`draw`) has been added that allows the user to generate a single SVG file that corresponds to a single sub/sty/dsl trio. See line 29 of `const USAGE` for details. The call to `draw` is located in the `else` case of the `async ()` call on line 323. 
- A new command line option (`--staged`) has been added that allows the user to generate staged SVG files that correspond to a single sub/sty/dsl trio. See line 36 of `const USAGE` for details. 
Screen Shot 2021-06-28 at 12 46 16 PM

- Function `singleProcess` has a new parameter `staged` and several new cases to account for the `--staged` option. `batchProcess` also has a new parameter `staged` for this option. 

- Run `automator draw tree.sub tree.sty setTheory.dsl out --src-prefix=../../examples/set-theory-domain` in `automator` to generate an example SVG for a sub/sty/dsl trio. 
- Add a `--staged` flag to the above command (i.e. run automator draw tree.sub tree.sty setTheory.dsl out --src-prefix=../../examples/set-theory-domain --staged`) to generate staged SVGs for a sub/sty/dsl trio. 

# Checklist

# Open questions
- With the $LOCAL variables that are created for each matched substance statement + style block, it's difficult to tell which $LOCAL variables correspond to each substance statement (which is not great for controlled storyboarding)
- The generated svg window size is weirdly vertical (how can this be controlled?) (this may also be a result of `State` not having a `canvas` property in `automator`). 
- For comic-like staging, it might be possible to treat each generated SVG as a rectangle, and then do some optimization on how to lay that out.


packages/automator/README.md
```diff
@@ -10,12 +10,14 @@ Penrose Automator.
 Usage:
   automator batch LIB OUTFOLDER [--folders]  [--src-prefix=PREFIX] [--repeat=TIMES] [--render=OUTFOLDER]
   automator render ARTIFACTSFOLDER OUTFOLDER
+  automator draw SUBSTANCE STYLE DOMAIN OUTFOLDER [--folders] [--src-prefix=PREFIX] [--staged]
```
wodeni, line 13: Not sure if we want the `--folder` option here. Consider removing it

packages/automator/README.md
```diff
@@ -39,3 +41,8 @@ In addition to batch-processing Penrose programs, you can also use `automator` t
 - Open `browser/index.html` to view the result.
 
 ![](docs/penrose-artifacts.png)
+
+## 06/25/2021 Update: Staged Diagram Generation
```
wodeni, line 45: Don't have to date this :P. A quick screenshot would be great!

packages/automator/index.tsx
```diff
@@ -8,6 +8,8 @@ import {
   stepUntilConvergence,
 } from "@penrose/core";
 import { renderArtifacts } from "./artifacts";
+import { getListOfStagedStates } from "../core/src/renderer/Staging";
```
wodeni, line 11: Kinda surprised that this worked. Export your functions in `index.tsx` of `core` and import it here from `@penrose/core`. These are separate packages, so relative paths won't work in general.

packages/automator/index.tsx
```diff
@@ -175,8 +207,21 @@ const singleProcess = async (
     // returning metadata for aggregation
     return { metadata, state: optimizedState };
   } else {
-    fs.writeFileSync(out, canvas);
-    console.log(chalk.green(`The diagram has been saved as ${out}`));
+    if (staged) {
+      // write multiple svg files out
+      const writeFileOut = (canvasData: any, index: number) => {
+        let filename = out.slice(0, out.indexOf("svg") - 1);
```
wodeni, line 213: Perhaps use [`path.join`](https://nodejs.org/api/path.html#path_path_join_paths) and templated strings for readability

packages/automator/index.tsx
```diff
@@ -222,13 +268,14 @@ const batchProcess = async (
     // try to render the diagram
     try {
       // Warning: will face id conflicts if parallelism used
-      const res = await singleProcess(
+      let res = await singleProcess(
```
wodeni, line 271: Why the `let` here?

packages/automator/index.tsx
```diff
@@ -222,13 +268,14 @@ const batchProcess = async (
...
         subURI,
         styURI,
         dslURI,
         folders,
         `${out}/${name}-${id}${folders ? "" : ".svg"}`,
         prefix,
+        staged, // make something else to fix singleProcess?
```
wodeni, line 278: Is there something wrong with `singleProcess`?

packages/automator/index.tsx
```diff
@@ -273,27 +320,31 @@ const batchProcess = async (
...
-      await batchProcess(args.LIB, folders, args.OUTFOLDER, prefix);
+      await batchProcess(args.LIB, folders, args.OUTFOLDER, prefix, staged);
     }
     if (browserFolder) {
       renderArtifacts(args.OUTFOLDER, browserFolder);
     }
   } else if (args.render) {
     renderArtifacts(args.ARTIFACTSFOLDER, args.OUTFOLDER);
   } else {
+    // this assumes that yarn start draw was called.
```
wodeni, line 339: Now that I think about this, I think it's probably better if we do `else if (args.draw)` and return an error on `else`.

packages/core/src/renderer/Staging.ts
```diff
@@ -0,0 +1,87 @@
...
+      throw new Error("shape property doesn't have a .");
+    }
+    const shapeName = shapePropPathName.slice(0, dotIndex);
+    return shapeNamesToInclude.includes(shapeName);
+  };
+
+  const newShapeList = state.shapes.filter(includeShape);
+
+  // to be cleaner, i should technically update shapeOrdering as well
+  // but not modifying that doesn't seem to affect the creation of the SVG
```
wodeni, line 80: Something like:
```ts
return {
  ...state,
  shapes: newShapeList
}
```

wodeni (changes requested):
Looking great! Only some minor comments below.

packages/automator/README.md
heleaf: What do you mean by a screenshot? 😅

packages/automator/index.tsx
heleaf: woops I'm not sure why I made a comment there 😅 `singleprocess` is fine (i think that was when I was doing some testing)

packages/automator/index.tsx
heleaf: (changed it back to `const` 😅 )

wodeni (approved):
Thanks for addressing the comments! Good to merge now 🥳

Issue #607, opened 2021-06-24
certain file triples take several seconds to optimize


hsharriman:
**Describe the bug**
The optimization time for some file triples is on the order of 5-10 seconds per resample, and 3-5 seconds with a couple of `repel` constraints removed. After optimization, several of the constraints remain unsatisfied (see screenshot below)

**To Reproduce**
The branch: `main` of Penrose.  
domain: https://github.com/penrose/penrose/blob/main/packages/synthesizer/__tests__/geometry.dsl
substance: https://github.com/penrose/penrose/blob/main/packages/synthesizer/__tests__/congruent.sub
style: https://github.com/penrose/penrose/blob/main/packages/synthesizer/__tests__/euclidean.sty

Noticed that commenting out lines [144-145](https://github.com/penrose/penrose/blob/fec0dd9c7b6e93c64307f676409abb65bf1828ef/packages/synthesizer/__tests__/euclidean.sty#L144) in `euclidean.sty` dropped the optimization time from 5-10 seconds to 3-5 seconds

Steps to reproduce the behavior (from latest version of `main`):
1. `yarn start` from Penrose root
2. Run from Penrose root: 
```
npx roger watch packages/synthesizer/__tests__/congruent.sub packages/synthesizer/__tests__/euclidean.sty  packages/synthesizer/__tests__/geometry.dsl
```
5. Open `localhost:3000` and hit resample or reload the page

**Screenshots**
![Screenshot from 2021-06-21 17-39-19](https://user-images.githubusercontent.com/31522997/123337060-8baa3680-d4fb-11eb-9666-cdf33e1d87f9.png)

hypotext:
Instead of repelling a label from a segment, you could try using a constraint that ensures that a label has some minimum distance from a segment, e.g. by calculating the closest point to the label as a point. (This function already exists as `closestPt_PtSeg`; you'd just need to write a wrapper around it to make it a constraint.) 

My intuition is that this will be faster since the closest point calculation doesn't subsample a segment. But it may also be more numerically unstable since the closest point may "jitter" around as the label is moved during optimization. 

Let me know what happens!

joshpoll:
It looks like the diagram may be getting stuck in a local minimum where triangle CDE is nearly degenerate. Adding some constraints to prevent this might help. For example, encouraging angles to be less than 180 degrees.

[...]

PR #613, opened 2021-06-25
fix: repel objective on segments causes slow optimization time

hsharriman:
# Description
Using `repel` to ensure that text labels are a certain distance from a segment causes optimization times of between 10-15 seconds. Based on the discussion in related issue #607, implemented 2 changes:
* Added a constraint `closestPtToSeg` that ensures a given point is a certain distance away from the **nearest point on a given segment**. As @hypotext mentions in the issue, this is faster because it does not involve sampling the distance at multiple points of the segment. After implementing this objective, the resample time drops from 10-15 seconds to 2-6 seconds.
* Per @joshpoll suggestions, added a constraint `nonDegenerateAngle` to push angles to be within a reasonable range (avoid 0 and 180-degree angles). At the moment, this is implemented with a cutoff, so any angles between 10 and 170 degrees have automatically satisfied this objective.

1. `yarn start`
2. Run `npx roger watch packages/synthesizer/__tests__/congruent.sub packages/synthesizer/__tests__/euclidean.sty  packages/synthesizer/__tests__/geometry.dsl` from penrose root.
# Checklist

# Open questions
This was my first time writing a new constraint, and though everything appears to be working properly I'd definitely appreciate feedback on the implementation.

### DIFF
--- a/examples/geometry-domain/euclidean-simple.sty
+++ b/examples/geometry-domain/euclidean-simple.sty
@@ -150,8 +150,8 @@ with Point p; Point q {
      p.icon above e.icon
      q.icon above e.icon
 
-     encourage closestPtToSeg(e.icon, p.text.center, 30.)
-     encourage closestPtToSeg(e.icon, q.text.center, 30.)
+     encourage pointAtDistFromClosestPt(e.icon, p.text.center, 30.)
+     encourage pointAtDistFromClosestPt(e.icon, q.text.center, 30.)
 
      e.icon above const.plane
 }
--- a/examples/geometry-domain/euclidean-simple.sty
+++ b/examples/geometry-domain/euclidean-simple.sty
@@ -150,8 +150,8 @@ with Point p; Point q {
      p.icon above e.icon
      q.icon above e.icon
 
-     encourage pointLineDist(e.icon, p.text.center, 30.)
-     encourage pointLineDist(e.icon, q.text.center, 30.)
+     encourage pointLineDist(p.text.center, e.icon, 30.)
+     encourage pointLineDist(q.text.center, e.icon, 30.)
 
      e.icon above const.plane
 }
--- a/examples/geometry-domain/euclidean-simple.sty
+++ b/examples/geometry-domain/euclidean-simple.sty
@@ -150,8 +150,8 @@ with Point p; Point q {
      p.icon above e.icon
      q.icon above e.icon
 
-     encourage repel(e.icon, p.text, const.repelWeight)
-     encourage repel(e.icon, q.text, const.repelWeight)
+    encourage closestPtToSeg(e.icon, p.text.center, 30.)
+    encourage closestPtToSeg(e.icon, q.text.center, 30.)
 
      e.icon above const.plane
 }
--- a/examples/geometry-domain/euclidean-simple.sty
+++ b/examples/geometry-domain/euclidean-simple.sty
@@ -150,8 +150,8 @@ with Point p; Point q {
      p.icon above e.icon
      q.icon above e.icon
 
-     encourage pointAtDistFromClosestPt(e.icon, p.text.center, 30.)
-     encourage pointAtDistFromClosestPt(e.icon, q.text.center, 30.)
+     encourage pointLineDist(e.icon, p.text.center, 30.)
+     encourage pointLineDist(e.icon, q.text.center, 30.)
 
      e.icon above const.plane
 }
--- a/examples/geometry-domain/euclidean-simple.sty
+++ b/examples/geometry-domain/euclidean-simple.sty
@@ -150,8 +150,8 @@ with Point p; Point q {
      p.icon above e.icon
      q.icon above e.icon
 
-    encourage closestPtToSeg(e.icon, p.text.center, 30.)
-    encourage closestPtToSeg(e.icon, q.text.center, 30.)
+     encourage closestPtToSeg(e.icon, p.text.center, 30.)
+     encourage closestPtToSeg(e.icon, q.text.center, 30.)
 
      e.icon above const.plane
 }
--- a/packages/core/src/contrib/Constraints.ts
+++ b/packages/core/src/contrib/Constraints.ts
@@ -229,6 +229,45 @@ export const objDict = {
   nearScalar: (c: any, goal: any) => {
     return squared(sub(constOfIf(c), constOfIf(goal)));
   },
+  /**
+   * try to make distance between a point and a segment `s1` = padding.
+   */
+  closestPtToSeg: ([, s1]: [string, any], point: VarAD[], padding: VarAD) => {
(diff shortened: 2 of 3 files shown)


packages/core/src/contrib/Constraints.ts
```diff
@@ -229,6 +229,53 @@ export const objDict = {
...
+      const c1 = fns.center(p1);
+      const c2 = fns.center(p2);
+
+      const l1 = ops.vsub(c0, c1);
+      const l2 = ops.vsub(c2, c1);
+
+      // angles between 10 and 170 degrees do not need to be pushed
+      return ifCond(
+        lt(
+          absVal(ops.vdot(ops.vnormalize(l1), ops.vnormalize(l2))),
```
joshpoll, line 264: factor the cosine calculation into a new local variable and re-use it for L264 and L270

examples/geometry-domain/euclidean-simple.sty
```diff
@@ -150,8 +150,8 @@ with Point p; Point q {
      p.icon above e.icon
      q.icon above e.icon
 
-     encourage repel(e.icon, p.text, const.repelWeight)
-     encourage repel(e.icon, q.text, const.repelWeight)
+    encourage closestPtToSeg(e.icon, p.text.center, 30.)
```
joshpoll, line 153: add back space on L153 and L154

packages/synthesizer/__tests__/euclidean.sty
```diff
@@ -141,8 +141,8 @@ with Point p; Point q {
      p.icon above e.icon
      q.icon above e.icon
 
-     encourage repel(e.icon, p.text, const.repelWeight)
-     encourage repel(e.icon, q.text, const.repelWeight)
+    encourage closestPtToSeg(e.icon, p.text.center, 30.)
```
joshpoll, line 144: add back space on L144 and L145

joshpoll (changes requested):

packages/core/src/contrib/Constraints.ts
```diff
@@ -229,6 +229,48 @@ export const objDict = {
   nearScalar: (c: any, goal: any) => {
     return squared(sub(constOfIf(c), constOfIf(goal)));
   },
+  /**
+   * try to make distance between a point and a segment `s1` = padding.
+   */
+  closestPtToSeg: ([, s1]: [string, any], point: VarAD[], padding: VarAD) => {
```
hypotext, line 235: Looks mostly good, though I assumed you would make this a constraint, not an objective.

The thing to note is that `equalHard` is written as a constraint (as noted in the function comment). So I would either:

- Make `closestPtToSeg` a constraint and see how well that works, OR
- Use `equalHard` squared, and keep `closestPtToSeg` as an objective. That will give it nicer numerical properties as an energy (currently it is linear and looks like this: https://www.wolframalpha.com/input/?i=f%28x%2C+y%29+%3D+max%28x%2C+y%29+-+min%28x%2C+y%29 )

Let me know how well either option works. 

Also, I'd give your function a more descriptive name, like `pointAtDistFromClosestPt`.

hypotext (changes requested):

packages/core/src/contrib/Constraints.ts
hsharriman: I tried both of your suggestions, and it seems like keeping it as an objective is a bit faster and more reliable than making it a constraint. When I made it a constraint, every now and again it would take ~3minutes to optimize and end in a degenerated state such as this:
Screenshot from 2021-06-28 11-38-48
Keeping it as an objective still results in the 2-5 second time between resamples that I described in the original PR comment (as a constraint it is closer to 5 seconds on average)

hypotext (approved):
These changes look fine to me. @joshpoll @wodeni What do you think, is this ready to merge?

packages/core/src/contrib/Constraints.ts
```diff
@@ -229,6 +229,54 @@ export const objDict = {
...
+      const c1 = fns.center(p1);
+      const c2 = fns.center(p2);
+
+      const l1 = ops.vsub(c0, c1);
+      const l2 = ops.vsub(c2, c1);
+
+      const cosine = absVal(ops.vdot(ops.vnormalize(l1), ops.vnormalize(l2)));
+      // angles between 10 and 170 degrees do not need to be pushed
+      return ifCond(
+        lt(cosine, varOf(0.985)), // cos10 deg = .9848, abs(cos170) = .9848
```
wodeni, line 252: It would be good if we can take the min and max angles as arguments, perhaps with defaults `10` and `170`?

wodeni (approved):
Looking at the code, the changes look good to me. Seems like the perf problem is still there (?), but I think it's good to merge as long as @hypotext and @joshpoll think these functions are better alternatives to `repel`

packages/core/src/contrib/Constraints.ts
```diff
@@ -229,6 +229,54 @@ export const objDict = {
...
+    } else {
+      throw new Error(
+        "nonDegenerateAngle: all input shapes need to have centers"
+      );
+    }
+  },
+  /**
+   * try to make distance between a point and a segment `s1` = padding.
+   */
+  pointAtDistFromClosestPt: (
```
joshpoll, line 265: maybe `pointLineDist` or something similar would be a better name? point-line distance is canonically the shortest distance from the point to the line

joshpoll (changes requested):

packages/core/src/contrib/Constraints.ts
hsharriman: I added a `range` variable that defines how many degrees from 0 or 180 are considered unsafe (it's 10 by default, so angles <10 degrees or >170 degrees are in the "unsafe" range)

packages/core/src/contrib/Constraints.ts
```diff
@@ -229,6 +230,51 @@ export const objDict = {
...
+    } else {
+      throw new Error(
+        "nonDegenerateAngle: all input shapes need to have centers"
+      );
+    }
+  },
+  /**
+   * try to make distance between a point and a segment `s1` = padding.
+   */
+  pointLineDist: ([, s1]: [string, any], point: VarAD[], padding: VarAD) => {
```
joshpoll, line 267: Apologies for the back and forth here, but if you could switch the arguments to this objective so the first argument is the point and the second argument is the line that would be awesome. Also please test if the line `isLineLike` and throw an error if it's not. (check the other constraints and objectives to see how this is done)

joshpoll (changes requested):
overall lgtm once this change is incorporated I will approve!

joshpoll (approved):

hypotext:
Forgot @joshpoll didn't have merge permission. Done now, thanks @hsharriman!

PR #623, opened 2021-07-16
feat: predicate aliasing 

heleaf:
# Description
Screen Shot 2021-07-19 at 5 18 33 PM

Related issue/PR: n/a

### Motivation:
- Currently, GPIs + constraints/objective expressions associated with predicates can't be explicitly referenced by other selector blocks or in the translation, since they can only be defined locally within selector blocks. 
- Predicate aliasing allows for predicate-associated objects to be referenced in multiple blocks, which gives greater flexibility to style writers. 
- As a precursor to animation, being able to directly reference the GPIs/constraints/objectives associated with predicate instances in the translation is useful for giving the user greater control over creating staged diagrams.

### Summary of changes:
- Added `as` keyword to the Style language for predicate aliasing (style parser)
- New predicate aliasing in the translation when `as` keyword is used (style compiler)
- Well-formedness checking of the alias keywords based on domain and selector environment keywords (style compiler)
- Added additional properties to certain types (`RelBind`) and a new alias-specific error `SelectorAliasNamingError`, with error message handling 
- Added some new test cases for predicate aliasing to `StyleParser.test.ts` and `Style.test.ts`, and working examples to `__tests__/predicate_aliasing_examples`

### Style Parser Changes
- Deprecation of `as` keyword for namespaces. To avoid ambiguity, the option to use `as` with namespaces in selectors has been removed from `Style.ne`, and test cases in `StyleParser.ts` that formerly used `as` & namespaces have had those components removed. 
- A new case for aliasing has been added to the `rel_pred` matching. 
- A new matching for at least one multiline whitespace character (`__ml`) has been added for use in the `rel_pred` alias case.  The use of `__ml` over `__` is mainly for backwards compatibility in existing test cases where `as` is used. 
- New aliasing test cases have been added to `StyleParser.test.ts`

### Style Compiler Changes 
1. **Alias well-formedness checking**

- Main function: `aliasConflictsWithDomainOrSelectorKeyword`, which checks that the alias name does not conflict with existing domain type/function/constructor declarations and selector environment variable names. 
- The full set of selector environment variables seems to only be available in the selector environment. To make it so that `aliasConflictsWithDomainOrSelectorKeyword` has access to the selector environment, the functions that `aliasConflictsWithDomainOrSelectorKeyword` is nested in (`checkRelPattern`, `checkRelPatterns`,`checkHeader`) now pass in the selector environment as another argument. (Previously, they only passed in the relevant part of the style AST like Header/RelationPattern and the merged domain + selector environment)
- `aliasConflictsWithDomainOrSelectorKeyword` and its helpers are nested under the `Types and code for selector checking and environment construction` region in `Style.ts`. 

2. **Unique alias per predicate instance substitution**

- Main function: `addRelPredAliasSubsts`, which takes in an existing valid substitution and returns a substitution with additional alias substitutions (if any). It only ever adds alias substitutions the existing collection of style variable : substance variable substitutions; it does not modify the input substitutions. 
- This function is called in `findSubstsSel`. Afterwards, the functions that substituted style variables with substance variables in the AST also substitute the style alias variables with distinct "predicate instance" names. The fully substituted AST is then folded into the translation as per usual (no changes made beyond the additional alias substitutions). 
- The "predicate instance" names are created to mirror the instances in which the predicate is declared in substance. Ex. `IsSubset(A,B)` in substance will lead eventually to the `IsSubset_A_B` object name in the translation. Nested predicates like `IsSubset(Union(A,B), C)` will lead to the `IsSubset_Union_A_B_C` object name in the translation. The main function for finding the predicate instance name is `getRelPredAliasInstanceName`.
- `addRelPredAliasSubsts` and its helper functions are nested under the `Types and code for finding substitutions` region in `Style.ts`. 

### New working examples
- In `__tests__/predicate_aliasing_examples`, there are some examples of style programs with aliased predicates. The optimization and end result doesn't seem to have changed, but the organization of GPIs/expressions in the translation is different than style programs without aliased predicates. 

### Misc. changes (other files touched) 
- New `alias` property added to `RelPred` in `types/style.ts`
- New error type `SelectorAliasNamingError` in `types/errors.ts`, with error message handling in `utils/Error.ts`

1. Run `yarn start` in the root folder.
2. Navigate to `packages/core/src/__tests__/predicate_aliasing_examples`
3. Run `npx roger watch` on one of the following trios:

- `tree.sub tree.sty setTheory.dsl` (has GPIs nested under aliases)
- `tree.sub venn.sty setTheory.dsl` (has constraints/objectives and GPIs nested under aliases)
- `tree.sub venn-3d.sty setTheory.dsl` (has aliasing in style, but no constraints/objectives or GPIs nested under aliases)

4. Open localhost, and toggle the objects in `translation/trMap` 

Screen Shot 2021-07-19 at 5 21 30 PM

# Checklist

# Open questions

- I may open another PR where the "merged" domain environment and style environment replaces the variable list in the domain environment with the variable list in the style environment (instead of merging the two lists). This may be more semantically correct, and would allow for `checkRelPattern` and associated functions to only need the domain environment and the AST node being checked to be passed in as arguments, as previously implemented. 
- Based on the toy examples I've run so far in the testing files, and with the files in `predicate_aliasing_examples`, this new method of nesting GPIs/expressions doesn't seem to affect the optimizing process (it appears to just be a new way of organizing GPIs/expression in the translation). I'm not too sure how the optimization problem is created, but double checking that the overall process isn't impacted by predicate aliasing would be good.


packages/core/src/compiler/Style.test.ts
```diff
@@ -300,6 +300,113 @@ describe("Compiler", () => {
 
   // TODO: There are no tests directly for the substitution application part of the compiler, though I guess you could walk the AST (making the substitution-application code more generic to do so) and check that there are no Style variables anywhere? Except for, I guess, namespace names?
 
+  describe("Correct style programs with predicate aliasing", () => {
+    const domainProg = "type Set\npredicate IsSubset : Set s1 * Set s2";
+    const subProg = "Set A\nSet B\nSet C\nIsSubset(B,A)\nIsSubset(C,B)";
+    // TODO: Name these programs
+    const styProgs = [
+      // These are mostly to test setting shape properties as vectors or accesspaths
+      `Set x; Set y where IsSubset(x,y) as foo {}
```
wodeni, line 309: Use multi-line string with the ` ``` ` syntax for better readability.

packages/core/src/compiler/Style.ts
```diff
@@ -397,9 +399,76 @@ const checkDeclPatternsAndMakeEnv = (
   );
 };
 
+/**
+ * Helper fxn for checking well-formedness of selectors with predicate aliases
+ * @param m : Map<string, any> (only the keys of the map are used)
+ */
+const getKeyWordsFromMap = (m: any): string[] => {
```
wodeni, line 406: The short hand for getting an array of keys of a `Map`: 
```ts
const keywords: string[] = [...m.keys()];
```

wodeni: Since the alternative is short, I guess this function is not needed. Just map over the kw maps and do `m => [...m.keys()]`

packages/core/src/parser/StyleParser.test.ts
```diff
@@ -149,18 +149,41 @@ where IsSubset(   );
     const prog = `
 forall Set A, B; Map f
 with Set C, D; Map \`g\`
-where C := intersect ( A, B, Not(f) ) ; IsSubset( A, B ); IsSubset( Union(A,B), C); Intersect (   )
```
wodeni, line 152: Would be a good idea to add some tests for alias expression locations, e.g.:
```
forall Set A, B; Map f
with Set C, D; Map \`g\`
where C := intersect ( A, B, Not(f) ) ; IsSubset( A, B ) as SubsetAB; IsSubset( Union(A,B), C) as nested_Subset; Intersect (   )
```
...and failure cases like:

```
forall Set A, B; Map f as Const -- should fail because decls can't be aliased
```

packages/core/src/compiler/Style.ts
```diff
@@ -608,6 +687,126 @@ const couldMatchRels = (
...
+    throw new Error("unknown tag");
+  }
+};
+
+/**
+ * Helper fxn for @function getRelPredAliasInstanceName
+ *
+ * Returns the substitution for a bindingform
+ */
+const getBindingFormAliasInstanceName = (
```
wodeni, line 754: `getAliasName` would be just fine. Long function names are not great for readability

packages/core/src/compiler/Style.ts
```diff
@@ -608,6 +687,126 @@ const couldMatchRels = (
...
+ *
+ * Ex, IsSubset(x,y) must have 'x' and 'y' exist as keys in @param subst
+ * for the relation to be valid.
+ */
+const isValidRelSubst = (subst: Subst, rel: RelationPattern): boolean => {
+  if (rel.tag === "RelPred") {
+    return rel.args.every((arg) => predArgExistsInSubst(arg, subst));
+  } else if (rel.tag === "RelBind") {
+    return false; // these don't have aliases
+  } else {
```
wodeni, line 744: There're no other possibilities. Why are we throwing here? TypeScript should complain about `boolean | undefined` doesn't match with `boolean` whenever you miss a case.

wodeni: Same comment goes to functions above this

wodeni (changes requested):
Thanks for the PR @heleaf :D. The changes look great to me overall (also really good documentation on the code)! See comments below for some requested changes.

packages/core/src/compiler/Style.test.ts
heleaf: I'm not sure why, but forming strings with ````` causes typescript to complain (?)
Screen Shot 2021-07-27 at 2 38 08 PM

I think it may be because the style program strings are later concatenated with the canvas preamble, and it seems like in general typescript doesn't like concatenating ````` syntax strings? o_o
Screen Shot 2021-07-27 at 2 39 40 PM

packages/core/src/parser/StyleParser.test.ts
heleaf: What do you mean by alias expression locations exactly? @wodeni

packages/core/src/compiler/Style.ts
heleaf: Would `getBindingFormAliasName` and `getRelPredAliasName` be ok? (I have these two as separate functions due to the recursive nature of `RelPred` and `BindingForm` types) @wodeni

packages/core/src/compiler/Style.ts
heleaf: Since my cases are using `if/else if` clauses, typescript complains that I haven't checked for other `else` cases (even though there are none), and that the return type `boolean` should instead be `boolean | undefined`. The error throwing is mostly to make typescript happy about the typing of the overall function (I could also replace the second `else if` case with `else`, but perhaps the function is more clear this way with explicit cases?)

Screen Shot 2021-07-27 at 2 54 59 PM

wodeni (approved):
Thanks for addressing the changes :D. Looks good to me!

wodeni:
Still pending reviews from @hypotext  Do you mind looking through the changes in `Style.ts`?

hypotext:
Yes, I will do that today

hypotext (approved):
This approach all seems fine, great job! I would just ask that you make sure all the existing examples work as here: https://github.com/penrose/penrose/wiki/Example-diagrams

> this new method of nesting GPIs/expressions doesn't seem to affect the optimizing process (it appears to just be a new way of organizing GPIs/expression in the translation)

I don't think there would be problems with the optimization, as long as everything is named and referred to consistently.

packages/core/src/compiler/Style.ts
```diff
@@ -608,6 +676,126 @@ const couldMatchRels = (
...
+  } else throw new Error("unknown tag");
+};
+
+/**
+ * Returns the substitution for a predicate alias
+ */
+// IsSubset(B,A) --> `IsSubset_B_A`
+// IsSubset(Union(B,C),A) --> `IsSubset_Union_B_C_A`
+// TODO: this can be refactored for a more descriptive name for nested cases
+// TODO: adding parentheses into the strings messes with GPIs sometimes?
```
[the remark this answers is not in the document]
heleaf: > // TODO: this can be refactored for a more descriptive name for nested cases
Right now I'm creating predicate names by replacing parentheses and commas with underscores, so if the user declares `IsSubset(Union(B,C),A)` in substance and assigns some objects to this predicate in style, then the key for the predicate instance in the translation's `trMap` will be `IsSubset_Union_B_C_A`. I think this sort of mapping is okay for now, but I'm not sure that it's bijective between the predicate instance (`IsSubset(Union(B,C),A)`) and the translation key (`IsSubset_Union_B_C_A`), which is why a more descriptive mapping that differentiates between parens and commas (esp for nested cases) might be good. 

> // TODO: adding parentheses into the strings messes with GPIs sometimes?
Ideally I would directly use the substance statement `IsSubset(Union(B,C),A)` as the key in the translation, but using parentheses in the key seems to mess with the shape GPIs sometimes (it might just be this one example)

`npx roger watch tree.sub tree.sty setTheory.dsl` in `__tests__/predicate_aliasing_examples`

Using all underscores in the translation keys:
Screen Shot 2021-08-03 at 4 06 55 PM

Using parentheses in the translation keys: (arrows missing their heads?)
Screen Shot 2021-08-03 at 4 07 32 PM

heleaf:
Running existing examples (https://github.com/penrose/penrose/wiki/Example-diagrams): 

`linear-algebra-domain/twoVectorsPerp-unsugared.sub linear-algebra-domain/linear-algebra-paper-simple.sty linear-algebra-domain/linear-algebra.dsl` 
Screen Shot 2021-08-03 at 4 11 10 PM

`set-theory-domain/twosets-simple.sub set-theory-domain/venn-small.sty set-theory-domain/setTheory.dsl`
Screen Shot 2021-08-03 at 4 23 00 PM

`nested.sub venn.sty setTheory.dsl`
Screen Shot 2021-08-03 at 4 24 52 PM

`set-theory-domain/tree.sub set-theory-domain/venn.sty set-theory-domain/setTheory.dsl`
Screen Shot 2021-08-03 at 4 25 46 PM

`set-theory-domain/tree.sub set-theory-domain/tree.sty set-theory-domain/setTheory.dsl`
Screen Shot 2021-08-03 at 4 26 32 PM

 `set-theory-domain/tree.sub set-theory-domain/venn-3d.sty set-theory-domain/setTheory.dsl`
Screen Shot 2021-08-03 at 4 29 19 PM

`set-theory-domain/continuousmap.sub set-theory-domain/continuousmap.sty set-theory-domain/setTheory.dsl`
Screen Shot 2021-08-03 at 4 29 51 PM

`hyperbolic-domain/hyperbolic-example-unsugared.sub hyperbolic-domain/PoincareDisk.sty hyperbolic-domain/hyperbolic.dsl` (needed to add `canvas` selector in style)
Screen Shot 2021-08-03 at 4 31 23 PM

`geometry-domain/pythagorean-theorem-unsugared.sub geometry-domain/euclidean.sty geometry-domain/geometry.dsl`

- Commented in some pre-existing types/functions/predicates in style. 
- The calls to `pointLineDist` in style generate `InvalidObjectiveNameError`s (not sure why). Commenting out the `pointLineDist` calls allows the diagram to be rendered. 
Screen Shot 2021-08-03 at 4 47 11 PM

`set-theory-domain/multisets.sub set-theory-domain/venn-comp-test.sty set-theory-domain/setTheory.dsl`
Screen Shot 2021-08-03 at 4 50 46 PM

`mesh-set-domain/DomainInterop-unsugared.sub mesh-set-domain/DomainInterop.sty mesh-set-domain/DomainInterop.dsl`
Screen Shot 2021-08-03 at 4 51 42 PM

`graph-domain/graph1.sub graph-domain/graph-theory.sty graph-domain/graph-theory.dsl`
Screen Shot 2021-08-03 at 4 52 30 PM

`molecules/glutamine.sub molecules/molecules-basic.sty molecules/molecules.dsl`
Screen Shot 2021-08-03 at 4 53 26 PM

`molecules/glutamine.sub molecules/molecules.sty molecules/molecules.dsl`
Screen Shot 2021-08-03 at 4 57 10 PM

`molecules/glutamine.sub molecules/molecules-elegant.sty molecules/molecules.dsl `
Screen Shot 2021-08-03 at 4 58 11 PM

`molecules/caffeine.sub molecules/molecules-basic.sty molecules/molecules.dsl`
Screen Shot 2021-08-03 at 4 54 43 PM

`molecules/caffeine.sub molecules/molecules.sty molecules/molecules.dsl`
Screen Shot 2021-08-03 at 4 59 48 PM

`molecules/caffeine.sub molecules/molecules-elegant.sty molecules/molecules.dsl`
Screen Shot 2021-08-03 at 5 00 40 PM

`logic-circuit-domain/half-adder.sub logic-circuit-domain/distinctive-shape.sty logic-circuit-domain/logic-gates.dsl`
Screen Shot 2021-08-03 at 5 01 23 PM

`logic-circuit-domain/full-adder.sub logic-circuit-domain/distinctive-shape.sty logic-circuit-domain/logic-gates.dsl`
Screen Shot 2021-08-03 at 5 02 58 PM

`graph-domain/small-graph.sub graph-domain/disjoint-rects.sty graph-domain/graph-theory.dsl`
Screen Shot 2021-08-03 at 5 05 03 PM

`graph-domain/small-graph.sub graph-domain/disjoint-rects-large-canvas.sty graph-domain/graph-theory.dsl`
Screen Shot 2021-08-03 at 5 05 31 PM

`graph-domain/small-graph.sub graph-domain/disjoint-rects-small-canvas.sty graph-domain/graph-theory.dsl`
Screen Shot 2021-08-03 at 5 06 12 PM

`graph-domain/small-graph.sub graph-domain/disjoint-rect-line-horiz.sty graph-domain/graph-theory.dsl`
Screen Shot 2021-08-03 at 5 06 36 PM

`monoidal-category-domain/test.sub monoidal-category-domain/monoidal-test.sty monoidal-category-domain/monoidal-test.dsl`
Screen Shot 2021-08-03 at 5 07 02 PM

packages/core/src/compiler/Style.ts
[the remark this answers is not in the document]
wodeni: > Ideally I would directly use the substance statement `IsSubset(Union(B,C),A)` as the key in the translation, but using parentheses in the key seems to mess with the shape GPIs sometimes (it might just be this one example)

Yeah parens in string keys might not be the greatest idea... Not sure exactly how it'd cause problems tho

joshsunshine:
I think @liangyiliang should pick this back up so it can be merged. I really like this feature and it would be very useful to have it in the system. It's also somewhat related the other style matching work he's done before.

liangyiliang:
Resolved and merged in PR #1066

Issue #498, opened 2021-02-25
Optimization error when running `nested.sub`


wodeni:
**Describe the bug**
Strange autodiff/opt error when running `nested.sub`. This example works on `v1.01.alpha.10` (e94c0627ed9c594fa6b452120b0fe4fca1ae509a) but is broken on `v1.01.alpha.13`.

**To Reproduce**
1. `roger watch nested.sub venn.sty setTheory.dsl`

**Screenshots**
![image](https://user-images.githubusercontent.com/11740102/109217119-65f66600-7783-11eb-999e-1c8d71e72314.png)

hypotext:
FYI the NaN seems to actually appear on 9fbe49f27cd90500f27e20bdb1010692795d171a (the multiline commit), not the one before it (style fixes). 

Maybe something to do with label dimensions? I did a quick check and the labels appear to render correctly with dimensions that make sense, and `nested.sub` doesn't have complicated labels. So it's not obvious to me what would cause this. Happy to look into it more next week.

wodeni:
> FYI the NaN seems to actually appear on [9fbe49f](https://github.com/penrose/penrose/commit/9fbe49f27cd90500f27e20bdb1010692795d171a) (the multiline commit), not the one before it (style fixes).
> Maybe something to do with label dimensions? I did a quick check and the labels appear to render correctly with dimensions that make sense, and `nested.sub` doesn't have complicated labels. So it's not obvious to me what would cause this. Happy to look into it more next week.

Gotcha. Thanks for looking into this! Let's look into it again next week.

joshpoll:
I did some digging and this issue doesn't seem to be directly related to labels. The NaN appears even if the text labels are replaced with fixed-size rectangles. Moreover, with both text and rectangle labels, stepping manually does not generate NaNs and converges properly.

wodeni:
Resolved by #628.

PR #628, opened 2021-07-20
fix: resolved NaN in the nested.sub by increasing the max size of sets shapes #498

joshsunshine:
# Description

Related issue/PR: #498

Resolved NaN in nested.sub example by increasing maximum size of circles representing sets to accommodate the amount of nesting in this example without creating unsatisfiable constraints.

N/A

# Checklist

# Open questions

- Seems brittle; tests fail (the "cross-instance energy evaluation") if `maxSize`'s parameter is decreased to, e.g. `canvasSize/1.5`.

### DIFF
--- a/examples/set-theory-domain/venn.sty
+++ b/examples/set-theory-domain/venn.sty
@@ -14,7 +14,7 @@ forall Set x {
 
     ensure contains(x.icon, x.text)
     ensure minSize(x.icon)
-    ensure maxSize(x.icon, canvas.height / 3.)
+    ensure maxSize(x.icon, canvas.height / 1.)
     encourage sameCenter(x.text, x.icon)
     x.textLayering = x.text above x.icon
 }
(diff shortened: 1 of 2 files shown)


joshsunshine:
@wodeni In my original change I tried setting the maximum size to the canvas size / 1.5. This oddly caused the cross-instance energy evaluation test to fail in overall_tests. When I changed the value to cavas_size/1 it worked. Perhaps cross-instance energy evaluation is brittle?

wodeni (approved):
Tested with values between 1 and 2. No errors found in the test suite. If you were referring to `State is not prepared for energy evaluation`. This is not really an error. I lowered the log level of the message so we don't see it all the time. To explain this a bit more, there was a test in `overall.test` that deliberately re-generate the overall objective functions twice, which is not standard behavior for `evalEnergy`, so we added this message to warn people.

PR #629, opened 2021-07-20
chore: refactor SVG path interface

hsharriman:
# Description
The current way we describe SVG path commands in Penrose is pretty convoluted and spread across the codebase. Additionally, a lot of work is necessary to properly format types and interface data structures within `Functions.ts` that makes it more challenging than it needs to be to define an SVG path. 

Changes made:
* Removed all references to the `IPolygonV` interface as it has been deprecated for a while
* Added a `PathBuilder` class to `EngineUtils.ts` which abstracts the underlying Penrose-specific data structure and lets users chain together path commands.
* Defined a new data structure that requires fewer special cases (instead of 1 interface per path command there is 1 interface for coordinates and 1 interface for flags/values, see the implementation section for further clarification)
* Updated all functions in `Functions.ts` that create SVG path data to use the `PathBuilder` class instead
* Simplified the `Path.ts` shape file to use the new data structure
* Updated the `LabeledInput` component to use the new data structure, which had the unintended benefit of automatically resolving the `polygon pointrange mod unimplmented` limitation (see screenshot).
* Validated that behavior of the mod tab is preserved with these changes
* Validated that current behavior in implemented Penrose triples that use `pathData` are maintained
Screenshot from 2021-07-20 16-14-52
^the mod tab for one of the arcs that makes up the angleMarker. The variable names are a little funny but the functionality is there.
Related issue/PR: # (issue/PR)

The decision to implement a path builder was based on the D3 Path API: https://www.tutorialspoint.com/d3js/d3js_paths_api.htm. This allows us to abstract out technicalities of generic interfaces from the user, and simultaneously provides full control over the SVG spec, so the user only needs to focus on correctly calculating the arguments to pass into each SVG command.

The new interface data structure for `IPathDataV` is informed by the idea that all SVG path commands are a single character, followed by a space-separated list of numbers. The current implementation involves custom nested lists and many custom mapping functions just to convert from VarAD -> number. The new implementation reduces the number of unique interfaces down to a minimal set:
```
IPathData -> IPathCmd -> ISubPath
where
ISubPath = IValueV | ICoordV
```
* `IPathData` interface holds a list of all SVG paths in a penrose object
* `IPathCmd` interface holds a list of the SVG path commands needed to draw that shape
* `ISubPath` type corresponds to either a coordinate or a sequence of values. the distinction is that a coordinate needs to be converted from the global coordinate system to the canvas coordinate system in order to properly render, while values requires no adjustments. Any `(x,y)` coordinates should be defined as `ICoordV` interface, while things such as the radius of an arc or the degrees to rotate an arc should be defined as `IValueV` interface.

With the new interfaces we can avoid nested, customized mapping functions and provide one interface for all SVG path commands (`IPathCmd`). 
 
* check out this branch,  `shape-interface`
* run `yarn start`
* run `npx roger watch packages/synthesizer/__tests__/congruent.sub packages/synthesizer/__tests__/euclidean.sty  packages/synthesizer/__tests__/geometry.dsl` 
* shows functionality maintained for lines and arcs.
* Check the mod tab and click on `a4.mark` on the lefthand side to see that the `LabeledInput` component is working properly and actually allows you to make live adjustments to the points of the arc in addition to the basic style adjustments like strokeWidth and color
Also, you can check out the rewritten functions in `Functions.ts` to see how the new `PathBuilder` class is used.
# Checklist

- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] My changes generate no new warnings
- [x ] New and existing tests pass locally using `yarn test`
- [x ] I ran `yarn docs` and there were no errors when generating the HTML site
- [x ] My code follows the style guidelines of this project (e.g.: no ESLint warnings)

# Open questions

* There are not really examples of each type of path (i.e. cubic beziers and shorthand quadratic/cubic beziers are not implemented in `Functions.ts`) so their corresponding methods in the `PathBuilder` class have not been validated yet. However, their implementation is 95% the same as the commands that have been tested. I can extend test coverage in a future PR.


packages/core/src/renderer/Path.ts
```diff
@@ -5,79 +5,25 @@ import { flatten } from "lodash";
...
-  }
-
-  const [headCommand, ...tailCommands] = commands;
-  return (
-    fstCmdString(headCommand, canvasSize) +
-    tailCommands.map((cmd: any) => toCmdString(cmd, canvasSize)).join(" ")
-  );
-};
-
 const toPathString = (pathData: any[], canvasSize: [number, number]) =>
```
wodeni, line 8: We can probably give a type to `pathData: IPathCmd[]` now?

packages/core/src/renderer/Path.ts
```diff
@@ -5,79 +5,25 @@ import { flatten } from "lodash";
...
+        contents.map((c: any) => {
+          if (c.tag === "CoordV") return toScreen(c.contents, canvasSize);
+          else if (c.tag === "ValueV") return c.contents;
+          else {
+            console.error("WARNING: improperly formed pathData");
+            return;
+          }
+        })
+      ).join(" ");
+      return `${cmd} ${pathStr}`; // TODO need to convert to pt
```
wodeni, line 26: Is the TODO resolved?

packages/core/src/engine/EngineUtils.ts
```diff
@@ -265,6 +216,136 @@ function mapPalette<T, S>(f: (arg: T) => S, v: IPaletteV<T>): IPaletteV<S> {
 
 // Utils for converting types of values
 
+export class PathBuilder {
```
wodeni, line 219: Feels a little weird to put this in utils. Do you think `renderer/ShapeDef` is better for this?

wodeni (approved):
Looking great! This PR really simplified and standardized the `Path` interface 💯 . See some small questions I had while reading the code below.

packages/core/src/engine/EngineUtils.ts
hsharriman: It does seem a bit weird here. ShapeDef looks like it defines the interfaces for each shape that can be used in style. This class doesn't get used in Style programs (it is used in Functions.ts), so maybe another spot? Maybe it could even have its own file since we don't have any other "svg builders" implemented yet?

wodeni: `shapes/PathBuilder.ts`?

wodeni (approved):

PR #643, opened 2021-08-12
feat: added multiple tick mark functionality to euclidean style

hsharriman:
# Description
Screenshot from 2021-08-11 18-18-43
Previously, the `EqualLengthMarker` predicates would render `x` ticks directly on top of each other, making it look like there was only one tick at any given time. This PR makes them render properly.

* Used the new PathBuilder class to make the SVG path easily
* The function that was added in Functions.ts can be easily recycled in the future when we are able to iteratively add tick marks to any segments with `EqualLengthMarked` without explicitly calling `EqualLengthMarker1` `EqualLengthMarker2` `EqualLengthMarker3`

1. check out this branch, `yarn start`
2. `npx roger watch examples/geometry-domain/textbook_problems/ex.sub examples/geometry-domain/euclidean.sty  examples/geometry-domain/geometry.dsl`
3. This example will often lay the triangles directly on top of each other but you can still see the various 1,2, and 3 mark variants
# Checklist

# Open questions

### DIFF
--- a/packages/core/src/contrib/Functions.ts
+++ b/packages/core/src/contrib/Functions.ts
@@ -481,6 +481,41 @@ export const compDict = {
     };
   },
   /**
+   * Create equally spaced tick marks centered at the midpoint of a line
+   * @param pt1: starting point of a line
+   * @param pt2: endping point of a line
+   * @param spacing: space in px between each tick
+   * @param numTicks: number of tick marks to create
+   * @param tickLength: 1/2 length of each tick
+   */
+  ticksOnLine: (
+    pt1: VarAD[],
+    pt2: VarAD[],
+    spacing: VarAD,
+    numTicks: VarAD,
+    tickLength: VarAD
+  ) => {
+    const path = new PathBuilder();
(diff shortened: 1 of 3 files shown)


packages/core/src/contrib/Functions.ts
```diff
@@ -480,6 +480,41 @@ export const compDict = {
...
+    // from the original pt1->pt2 line
+    const [x1p, y1p] = ops.vmove(mid, tickLength, normalDir);
+    const [x2p, y2p] = ops.vmove(mid, tickLength, ops.vneg(normalDir));
+
+    multipliers.map((multiplier) => {
+      const [sx, sy] = ops.vmove([x1p, y1p], multiplier, unit);
+      const [ex, ey] = ops.vmove([x2p, y2p], multiplier, unit);
+      path.moveTo([sx, sy]).lineTo([ex, ey]);
+    });
+    return path.getPath();
```
wodeni, line 516: Oh pretty interesting! So the tick marks are basically just a `<path>` with multiple move-and-line-to commands

wodeni (approved):
LGTM!

PR #646, opened 2021-08-18
feat: added swap-in mutator

hsharriman:
# Description
Added a new mutator to the synthesizer called the swap-in mutator. For any predicate, function, or constructor, this mutator will replace one of the statement's arguments with another random variable of the same type that is in scope. For instance:
```
Set, A, B, C
IsSubset(B, A)
```
A valid Swap-in mutation would be `IsSubset(B, A) -> IsSubset(B, C)`. Note that `IsSubset(B, B)` is not a valid swap because the argument is already included.

Builds upon Nimo's refactor of the synthesis module. 

* checkout branch, `yarn start` from root
* `yarn set-example` from `packages/synthesizer`
* `yarn generate-site` from `packages/automator`
* load the generated `index.html` in your browser
# Checklist

# Open questions


packages/core/src/analysis/SubstanceAnalysis.ts
```diff
@@ -236,6 +236,30 @@ export const signatureArgsEqual = (a: Signature, b: Signature): boolean => {
...
+ * Find all declarations with the same type as a given Identifier in the current program
+ * NOTE: When the statement doesn't exist in the program, `removeStmt` returns the original program without errors.
+ *
+ * @param id a Substance program
+ * @param prog the current program
+ * @returns a new Substance program with the statement removed
+ */
+export const identicalTypeDecls = (
+  ids: Identifier[],
+  prog: SubProg
```
wodeni, line 249: Perhaps a more straightforward way is to use `env`: `env.vars` is a symbol table of type `Map<string, TypeConsApp>`. I think you can simply check if the `TypeConsApp`s are equivalent.

wodeni (changes requested):
Overall looks great to me! Just a minor point about finding IDs of the same type below. 

Hopefully the refactored version made things easier to implement :D.

packages/core/src/analysis/SubstanceAnalysis.ts
hsharriman: If I need to pass in Env from `checkSwapInArgs` in `Mutation.ts` into this function, how would I do that?

wodeni: I think the original `Env` is stored in the `SynthesisContext` passed into the guard functions (i.e. `check*`)

hsharriman: So it seems that `Identifier` is not directly equatable to `TypeConsApp`. Is there a way to convert between the two/compare the two?

wodeni: I think what I meant was:

* instead of a `SubProg`, you can take in an `Env`. The reason is `env.vars` is a map from `string` ids to their corresponding `TypeConsApp` already, and that seems to be the thing you needed?
* You can find out the types of `ids` by looking up `vars` and find variables of the same type by filtering `vars`, too (if I understood this function correclty).

wodeni (approved):
The new changes look good to me! Please feel free to merge whenever you think this is ready.

Issue #581, opened 2021-06-03
Vector-Scalar Multiplication Causes Runtime Error in EngineUtils


joshpoll:
**Describe the bug**
Multiplying a scalar by a vector is fine, but multiplying a vector by a scalar results in the following runtime error:
```
Uncaught (in promise) TypeError: Cannot read property 'map' of undefined
    at mapVector (EngineUtils.ts:132)
    at mapValueNumeric (EngineUtils.ts:279)
    at valueAutodiffToNumber (EngineUtils.ts:306)
    at Evaluator.ts:225
    at lodash.js:13427
    at lodash.js:4925
    at baseForOwn (lodash.js:2990)
    at Function.mapValues (lodash.js:13426)
    at evalShape (Evaluator.ts:212)
    at Evaluator.ts:109
```

**To Reproduce**
I am using default penrose on the main branch.

Domain:
```
type Scalar
type Vector
function sMul : Scalar * Vector -> Vector
```

Substance:
```
Scalar a
Vector v
Vector mul := sMul(a, v)
```

Style:
```
forall Vector u {
    u.vector = (1., 1.)

    u.shape = Arrow {
        start: (0., 0.)
        end: u.vector
    }
}

forall Scalar a {
    a.val = 1.
}

forall Vector u; Vector v; Scalar a
where u := sMul(a, v) {
    override u.vector = v.vector * a.val
}
```

If the last line is `override u.vector = a.val * v.vector` this runs without error.

**Expected behavior**
Either this program should not error or Penrose should throw a more helpful error message at compile time when processing the Style file.

**Screenshots**
n/a

**Desktop:**
 - OS: macOS
 - Browser Chrome
 - Version 90

wodeni:
The root cause is probably https://github.com/penrose/penrose/blob/main/packages/core/src/engine/Evaluator.ts#L913, where scale-vector multiplication is only implemented one way.

wodeni:
Fixed by #650

PR #650, opened 2021-09-23
fix:  allow vectors to be multiplied by a scalar

Mystic5pirit:
Related issue/PR: #581

Allows vectors to be multiplied by a scalar


wodeni:
Awesome. Thanks for fixing this, @Mystic5pirit!

Issue #654, opened 2021-09-29
Substance checker not catching name collisions across types


wodeni:
**Describe the bug**
The program in #218 causes an unexpected error in Style rather than Substance:
```
Set A
Point A
AutoLabel All
```
Error message:
```
Overriding path 'A.icon' without override flag set
```

**To Reproduce**
https://panes.penrose.ink/gist/942f40dfd96510847f02b9455fe173c6

PR #657, opened 2021-10-05
fix: add duplicate name check in Substance checker

wodeni:
# Description

Related issue/PR: #654 

The Substance checker doesn't check for duplicated names (since the Haskell days!?). This PR adds duplicate name checks for `Decl`s.

* In `checkStmt`, check if the name exists right after the type constructor check. 
* Fix all tests in the existing suite that had duplicate names (surprise!)
* Add a test of the example in #654 in the suite.


samestep (approved):
lgtm other than a couple random questions

PR #662, opened 2021-10-12
chore: rename `repel` to `notTooClose` #300 

### LINKED ISSUE #300: Rename repel (and other "operational" encourage statements)
In Style we currently have lines of the form

```
encourage repel( x.shape, y.shape )
```

The name `repel` is counter to the idea of Style: Style programs are supposed to specify **what the diagram should look like** rather than **how the diagram should be produced**.  I.e., the name `repel` suggests something about how the layout algorithm is supposed to proceed.  A better name would be

```
encourage notTooClose( x.label, y.label )
```

This statement says, _"in the final diagram, I want these things to be separated; I don't care how you make that happen."_  Moreover, it frees up the Penrose layout engine to make different choices about how to achieve this objective.  (E.g., future versions are free to adopt a better layout strategy without betraying the intent of the Style programmer.)

Probably there are other "operational" objectives beyond `repel`; we should think about names that we can use in their stead.  (Hopefully this will be easy from an implementation point of view—"just" a global name change…)

wodeni:
# Description

Closes #300 

* Used "Find in files" in VSCode to replace all instances.

### DIFF
--- a/packages/core/src/contrib/Objectives.ts
+++ b/packages/core/src/contrib/Objectives.ts
@@ -123,12 +123,12 @@ export const objDictGeneral = {
   /**
    * Try to repel shapes `s1` and `s2` with some weight.
    */
-  repel: (
+  notTooClose: (
     [t1, s1]: [string, any],
     [t2, s2]: [string, any],
     weight = 10.0
   ): ad.Num => {
-    // HACK: `repel` typically needs to have a weight multiplied since its magnitude is small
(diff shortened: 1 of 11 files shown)


wodeni:
> Some of it might still be on gitbooks

Umm does that mean the source files are not in this GH org at all? Just somewhere on the gitbooks storage?

samestep:
We changed `repel` from a constraint to an objective in #751.

PR #663, opened 2021-10-12
fix: Throw error on invalid labels 

### LINKED ISSUE #492: throw error on invalid labels
**Describe the bug**
If a label is invalid, no error is thrown in the console or reported in the inspector. Instead, the desired label simply does not render. The rest of the labels (if valid) show up fine.

**To Reproduce**
For any Substance program with labels, write an invalid label, e.g. `Label e1 $\frac{1]{e^x$` (note nonterminated curly brace).

**Expected behavior**
The diagram does not render, and an error is thrown, and reported in the inspector, that indicates the incorrect label string, and perhaps the location that was incorrect (as reported by MathJax).

**Additional context**
Happens on latest `main`

samestep:
Fixes #492 using techniques from the following:

- https://github.com/mathjax/MathJax-demos-node/issues/3#issuecomment-497524041
- https://github.com/mathjax/MathJax-demos-node/issues/25#issuecomment-711247252

Example:

error message

Subtasks:

- [x] make `collectLabels` return a `Result`
- [x] statically type the return value of `tex2svg`
- [x] correct `SVGSVGElement` to `HTMLElement` (thanks @wodeni!)
- [x] actually make it work
- [x] give good error messages


packages/core/src/utils/CollectLabels.ts
```diff
@@ -90,34 +116,44 @@ export const retrieveLabel = (
...
-          contents: height === Infinity ? 0 : (height as number),
+          contents: height === Infinity ? 0 : height,
         },
-        rendered: body as HTMLElement,
+        rendered: body,
       };
       labels.push([shapeName, label]);
     }
   }
-  return Promise.all(labels);
```
wodeni, line 122: I'm not 100% sure if the async behavior will be the same now that `Promise.all` is gone. I guess since you are validating and pushing to the cache per step, it's kind of the same as before?

wodeni (approved):
Looks good! As noted below, I'm not super sure if there will be any performance change at all and didn't really run anything to test it. I'll rely on your experience on this. Good to go for me!

packages/core/src/utils/CollectLabels.ts
samestep: Yep, that's my understanding; I think it's fine since it passed the typechecker?

PR #215, opened 2019-05-07
Geometry domain

hypotext:
# Description

This PR adds support for the following:
* Geometry domain in the Euclidean style and 3D spherical style, as documented on the [wiki page](https://github.com/penrose/penrose/wiki/The-geometry-domain)
* New Style syntax for the TOG paper submission (with parser, semantics, and ported Style programs)

Related issues: #219 #201

# Implementation strategy and design decisions

See the geometry wiki page.

# Examples with steps to reproduce them

`runpenrose geometry-domain/teaser.sub geometry-domain/euclidean.sty geometry-domain/geometry.dsl`
`runpenrose geometry-domain/teaser.sub geometry-domain/spherical.sty geometry-domain/geometry.dsl`

See #219 for Style parser examples.

# Open questions

Questions that require more discussion or to be addressed in future development:

- [ ] Geometry style refinements
- [ ] Hyperbolic style

(See geometry wiki page)

PR #693, opened 2021-11-15
feat: port the SIGGRAPH Euclidean geometry example

wodeni:
# Description

Related issue/PR: #215

Port the SIGGRAPH Euclidean geometry examples using our new system. 

image

The example is added to the registry: `siggraph-teaser-euclidean-teaser`


wodeni:
Update: progress so far

image

The markers for `PerpendicularBisector` were computed by Haskell function back then, so it's a little hard to port. Other things in the Style program were pretty low effort to port. 

In general, spherical and hyperbolic examples seem very hard to port because they rely on hyper-specific computation functions written in Haskell.

Issue #642, opened 2021-08-06
more efficient `if` evaluation in energy/gradient compilation


hypotext:
Currently, energy/gradient compilation evaluates both branches of an `if`. To be more efficient, it should just evaluate one branch of an `if`. Energy and gradient are "hot paths" of the system so this optimization is probably worth it.

Details: for an `if` statement `ifCond(e, a, b)`, since everything is uniformly compiled into expressions, it becomes `const y = eval(e) ? eval(a) : eval(b)`. This could be clearly made way more efficient (albeit less functional) if compiled into:

```
let y;
if (eval(e)) {
  y = eval(a);
else {
  y = eval(b);
}
```

samestep:
It seems like this relates to autodiff expressiveness, not just efficiency: sometimes it would be useful to use a conditional to avoid doing computations that would throw errors, such as the `or(eq(rxRaw, constOf(0)), eq(ryRaw, constOf(0)))` check that #698 would like to do but doesn't.

wodeni:
Would #1636 solve this issue @samestep?

samestep:
@wodeni not currently, because Rose currently only supports and not branching (since supporting differentiation of the latter would require sum types); but it will help with #561.

PR #698, opened 2021-11-22
feat: Define bounding box function for every shape

### LINKED ISSUE #696: Make bounding box mandatory for every shape
#582 introduced the `bboxFromShape` function, but it only works for a subset of our shapes. As we introduce more general constraints like `onCanvas` (#694), reliable and general bounding box computation will be vital to the system. Just like any other graphics system, every shape should have a `bbox` function. The current proposal is:

* Introduce a `bbox` function for every shape defined in the `ShapeDef` module. 
* Ideally, the `bbox` function should be co-located with the rest of the shape definition for a particular shape. For instance, `circDef` should include a function for computing its bounding box.
  * `Path` will be a tricky shape to handle, since it could contain `Arc`, `CubicBezier`, and `QuadraticBezier`. Tips from @keenancrane:

> bboxFromShape for a polygon should be just a max/min over the vertex coordinates
> For a Path I think it's the same: assuming it's a Bézier curve, the curve is contained in the convex hull of the control points.  The bounding box of the control points in turn contains the convex hull.  So again you just have to take a max/min over the control point coordinates.
> (for arcs) use a conservative bound for now, such as the bounding box of the circle whose radius is max(rx,ry) (or work out an exact solution).
 
* Refactor `bboxFromShape` to use shape-type-specific `bbox` functions

Related effort on `ShapeDef`: #558.

samestep:
# Description

Resolves #696.

This PR replaces the implementation of the `bboxFromShape` function (in `contrib/Constraints.ts`) with a call to the `bbox` method on the shape definition for the given shape type. That is, the `IShapeDef` interface (in `renderer/ShapeDef.ts`) is given a `bbox` field, and each shape definition in that file is given its own bbox implementation to use for that field.

The bbox implementations in this PR for the following shapes are precise modulo stroke width (as far as I am aware):

- `Circle`
- `Ellipse`
- `Rectangle`
- `Callout` (modulo some rendering weirdness)
- `Polygon` (see #709)
- `FreeformPolygon`
- `PathString` (I just use the given `center`, `w`, and `h`)
- `Polyline` (see https://github.com/penrose/penrose/issues/709#issuecomment-989203210)
- `Image` (modulo #712)
- `Text` (depending on your definition of "precise")
- `Line` and `Arrow` (modulo arrowheads)

The following are not:

- `Square` (when `rotation` and `rx` are both nonzero)
- `Path` (as per @keenancrane's suggestions, quoted in #696)
  - for Bézier curves, we use the bounding box of the control points
  - for arcs, we use the bounding box of the circle with radius `max(rx, ry)` whose center is the same as that of the ellipse

<details>
<summary>(expand for examples/visualizations of imprecise bboxes)</summary

square bbox

Bézier bbox

arc bbox
</details>

This PR adds tests for most of the bbox functions; the exception is `Text`, for which the testing strategy (see below) doesn't work. Still, the `bboxFromRectlike` function which `Text` uses is itself tested via the tests for `PathString` and `Image`, so this doesn't really result in lack of coverage.

Thanks to #710, the parameter for each bbox function is typed `Properties<VarAD>` instead of `any`. However, as mentioned in #701, this still isn't precise enough to distinguish individual shape types, so each bbox function contains a lot of checks on `tag`s and on lengths of vectors (using the `isPt2` function which this PR adds to `types/ad.ts`) to make TypeScript happy. These could be made somewhat less verbose by replacing them with unsound `as` casts, but probably a better solution would be to just wait for #701 to be resolved, which would remove the need for them entirely.

Several of the bbox functions use the `bboxFromPoints` helper function. This was originally written just for `Polygon`, but it also turned out to be convenient for handling rotation on rectangles, as well as `Line`s and `Arrow`s. However, it is also used for `Path`, although I feel like for that particular case, there's probably a more elegant way to compute the bbox instead of by first assembling an array of points.

The bbox implementation for arcs in paths is particularly loose: as described in comments in the code (and earlier in this PR description), it simply returns the bbox of the smallest circle containing the entire ellipse of which that arc is a part. That comment links to https://stackoverflow.com/a/65441277 as an example of a much more precise ellipse arc bbox implementation, but that is difficult to translate into autodiff code because of things like #642.

As mentioned in #700, there is currently no easy way to programmatically instantiate shapes, so this PR tests the bbox functions by compiling little Style programs and then extracting the shapes from them. This required some minor modifications to the current evaluation pipeline, so #710 needed to be merged first.

Because of this method for constructing shapes, this PR needed to add a couple new functions (`quadraticCurveFromPoints` and `cubicCurveFromPoints`) to `contrib/Functions.ts` to allow for testing more Bézier commands in `Path`. As #716 mentions, it would probably be a good idea to replace these with a more general, principled set of `Path` construction functions.

In the tests themselves, we call `numOf` (from `engine/Autodiff.ts`) to get JavaScript numbers from each of the four `VarAD`s from the computed bbox, and then use Jest's [`toBeCloseTo`](https://jestjs.io/docs/expect#tobeclosetonumber-numdigits) matcher to compare those to the expected values. This gets the job done, but is a bit annoying: because a common workflow is to implement or modify a bbox function, run it using a test case, and then use https://panes.penrose.ink/ to visually check that the bbox is correct (see below); but since each test has four different assertions instead of just one (e.g. using [`toEqual`](https://jestjs.io/docs/expect#toequalvalue)), it can take up to four sequential runs of the test to extract all the computed values.

All the tests in `renderer/ShapeDef.test.ts` serve as examples. To reproduce most of them:

- substitute a shape Style string from `renderer/ShapeDef.test.ts` into the `makeSty` template at the top of that file
- go to https://panes.penrose.ink/
- paste that Style program into the **sty** box
- add a `global.bbox` which is a `Rectangle` using the expected `w`, `h`, and `center` from the appropriate test case
- click **compile**

The exceptions are the ones using the newly introduced `quadraticCurveFromPoints` and `cubicCurveFromPoints` functions, for which you instead need to check out this PR locally and use `yarn start` and `npx roger watch`.

# Checklist

# Open questions

- Should the tag/length checks be replaced with unsound `as` casts to reduce verbosity?
- Should the bbox implementation for `Path` arcs (or for that matter, Bézier curves and rounded squares) be made precise?

### DIFF
--- a/packages/core/src/contrib/Constraints.ts
+++ b/packages/core/src/contrib/Constraints.ts
@@ -1011,7 +1011,6 @@ export const areDisjointBoxes = (a: BBox.BBox, b: BBox.BBox): VarAD => {
  *   (shape-specific)
  * Input: A shape.
  * Output: A new BBox
- * Errors: Throws an error if called on an unsupported shape type.
  */
 export const bboxFromShape = (t: string, s: any): BBox.BBox => {
   return findDef(t).bbox(s);
--- a/packages/core/src/contrib/Constraints.ts
+++ b/packages/core/src/contrib/Constraints.ts
@@ -27,7 +27,7 @@ import {
 } from "engine/Autodiff";
 import * as _ from "lodash";
 import { linePts } from "utils/OtherUtils";
-import { isLinelike, isRectlike } from "renderer/ShapeDef";
+import { findDef, isLinelike, isRectlike } from "renderer/ShapeDef";
 import { VarAD } from "types/ad";
 import { every } from "lodash";
 import * as BBox from "engine/BBox";
@@ -1008,51 +1008,11 @@ export const areDisjointBoxes = (a: BBox.BBox, b: BBox.BBox): VarAD => {
 
 /**
  * Preconditions:
- *   If the input is line-like, it must be axis-aligned.
- *   Assumes line-like shapes are longer than they are thick.
- * Input: A rect- or line-like shape.
+ *   (shape-specific)
+ * Input: A shape.
  * Output: A new BBox
- * Errors: Throws an error if the input shape is not rect- or line-like.
+ * Errors: Throws an error if called on an unsupported shape type.
  */
 export const bboxFromShape = (t: string, s: any): BBox.BBox => {
-  if (!(isRectlike(t) || isLinelike(t))) {
-    throw new Error(
-      `BBox expected a rect-like or line-like shape, but got ${t}`
-    );
-  }
-
-  // initialize w, h, and center depending on whether the input shape is line-like or rect/square-like
-  let w;
-  if (t == "Square") {
-    w = s.side.contents;
-  } else if (isLinelike(t)) {
-    w = max(
-      absVal(sub(s.start.contents[0], s.end.contents[0])),
-      s.thickness.contents
-    );
-  } else {
-    w = s.w.contents;
-  }
-
-  let h;
-  if (t == "Square") {
-    h = s.side.contents;
-  } else if (isLinelike(t)) {
-    h = max(
-      absVal(sub(s.start.contents[1], s.end.contents[1])),
-      s.thickness.contents
-    );
-  } else {
-    h = s.h.contents;
-  }
-
-  let center;
-  if (isLinelike(t)) {
-    // TODO: Compute the bbox of the line in a nicer way
-    center = ops.vdiv(ops.vadd(s.start.contents, s.end.contents), constOf(2));
-  } else {
-    center = s.center.contents;
-  }
-
-  return BBox.bbox(w, h, center);
+  return findDef(t).bbox(s);
 };
--- a/packages/core/src/contrib/Functions.ts
+++ b/packages/core/src/contrib/Functions.ts
@@ -218,7 +218,7 @@ export const compDict = {
     return {
       tag: "ColorV",
       contents: {
-        tag: "NONE"
+        tag: "NONE",
       },
     };
   },
@@ -559,7 +559,7 @@ export const compDict = {
 
   /**
            * Figure out which side of the rectangle `[t1, s1]` the `start->end` line is hitting, assuming that `start` is located at the rect's center and `end` is located outside the rectangle, and return the size of the OTHER side. Also assuming axis-aligned rectangle. This is used for arrow placement in box-and-arrow diagrams.
-        
+
        @deprecated Don't use this function, it does not fully work
            */
   intersectingSideSize: (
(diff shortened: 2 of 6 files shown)


samestep:
TODO: before merging, make functions to construct the other Bezier curves, and test

joshsunshine:
> The` bbox implementations in this PR for the following shapes are precise modulo stroke width.
What would it take to include stroke width? @keenancrane is it normal to exclude stroke width in bbox computations? If not, what is usually done?

joshsunshine:
The PR should include examples of imprecision for squares and arcs. I assume for bezier curves the bbox is precise, but I'm wrong, add an example there too.

samestep:
> What would it take to include stroke width? @keenancrane is it normal to exclude stroke width in bbox computations? If not, what is usually done?

If we're considering that level of detail then I guess we would probably also have to think about things like [linecaps](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Fills_and_Strokes).

samestep:
> The PR should include examples of imprecision for squares and arcs. I assume for bezier curves the bbox is precise, but I'm wrong, add an example there too.

OK! Do you mean show some pictures of this imprecision? I can add those to the PR description if that would be helpful.

joshsunshine:
> OK! Do you mean show some pictures of this imprecision? I can add those to the PR description if that would be helpful.

Yes, I mean for you to include pictures of the imprecision.

joshsunshine:
> This gets the job done, but is a bit annoying: because a common workflow is to implement or modify a bbox function, run it using a test case, and then use https://panes.penrose.ink/ to visually check that the bbox is correct (see below); but since each test has four different assertions instead of just one (e.g. using toEqual), it can take up to four sequential runs of the test to extract all the computed values.

I don't understand why this takes four runs. How exactly are you extracting the values using panes.penrose.ink?

samestep:
> I don't understand why this takes four runs. How exactly are you extracting the values using panes.penrose.ink?

I'm not; I'm extracting the values by running the test case and looking at what it returns when the assertion fails.

joshsunshine:
> I'm not; I'm extracting the values by running the test case and looking at what it returns when the assertion fails.
Does our testing infrastructure allow only one assertion to fail per test? Why can't we collect information from all four failures?

samestep:
> Does our testing infrastructure allow only one assertion to fail per test? Why can't we collect information from all four failures?

As far as I know, yes; perhaps there's a way to disable the fail-fast behavior though?

samestep:
> Yes, I mean for you to include pictures of the imprecision.

Added :+1:

keenancrane:
image

Stroke width would not be hard to add.  In some vector graphics environments (like Illustrator) strokes can also be drawn inside/outside/centered on the path, which adds additional complexity.  However, SVG "fortunately" does not support this feature, and just draws all strokes centered.

So, let's consider what happens to the size of a shape if you stroke it by a stroke of width w.  Giving a closed path a width w is equivalent to sweeping a circle of radius w/2 along that path.  Hence, no matter which points are the extreme points along the path, the bounding box will need to grow by w/2 in all directions (top/bottom/left/right).  So, just add w to both the bounding box width and height.  Easy.

@samestep I wouldn't worry about linecaps _for now_, since this will get us pretty close.  But for a very large stroke width, our current bounding boxes may be way off.

Issue #392, opened 2020-10-24
Support empty fill and stroke style


keenancrane:
Currently to draw curves in Style we set the fill color to an RGBA color with zero opacity, e.g., `shape.color = rgba(0.0, 0.0, 0.0, 0.0)`.  A shape with opacity zero is however different from a shape with an empty fill style, specified in SVG (and CSS) via `fill: none;`. With an empty fill style, the interior does not get rasterized, and likewise, the interior is not treated as a selection mask.  Empty shapes are important for, e.g., drawing empty circles as outlines, rather than solid disks.  Similarly for strokes.

**Proposal.** In Style, define a type that has both `rgba` and the flag `none` as subtypes (say, `fillType`).  Then one can write either

```
x.shape = Circle {
   fill: rgba( red, green, blue, alpha )
}
```

or

```
x.shape = Circle {
   fill: none
}

Note that having a richer type `fillType` opens the door to (eventually) supporting other standard SVG features, such as [filling with a pattern](http://tutorials.jenkov.com/svg/fill-patterns.html).

cmumatt:
Resolved in main via merged PR# 699

PR #699, opened 2021-11-23
fix: Support empty fill and stroke style #392

cmumatt:
# Description

Related issue: #392 

Implement SVG compliant paint type 'none' for object fill and stroke.    This allows true lack of fill or stroke, which has different semantics than transparent fill (i.e., drag, etc).  Previously full transparency (i.e., rgb(0,0,0,0) was the closest we could accommodate fill=none)

Implemented computation function, none(), similar to rgba() and hvsa().  Note that the no-paint structure lacks color.

Substance:
```
Set A, B, C, D, E
-- Set A must contain B
IsSubset(B,A)

-- Set C is disjoint with D
IsDisjoint(C,D)

--Set E must contain A, B, C, D
IsSubset(A,E)
IsSubset(B,E)
IsSubset(C,E)
IsSubset(D,E)

IsEmptyFill(D)
IsEmptyFill(E)

IsEmptyStroke(B)
IsEmptyStroke(C)
AutoLabel All
```

Add to venn.sty:
```
forall Set x 
where IsEmptyFill(x) {
    override x.icon.color = none()     
}

forall Set x 
where IsEmptyStroke(x) {
     override x.icon.strokeColor = none()     
}
```

Add to setTheory.dsl
```
predicate IsDisjoint : Set s1 * Set s2
```

Compile.  Inspect output HTML and SVG source to observe fill=none on D, E and stroke=none on B, C.  Observe other shapes have normal coloring, as expected.

# Checklist

- [X] I have commented my code where appropriate, particularly in hard-to-understand areas

# Subsequent items

While reviewing code for this change, observed some opportunities for improvement as we move from the idea of "color" to "paint," in closer alignment with SVG's paint spec:
- Update all Color ‘none’ declarations in Style examples to ‘none()’ vs. the current rgba(0,0,0,0)
- Rename ‘color’ variables/types as ‘paint’ to better align with SVG and make the link to the standard clearer


packages/core/src/contrib/Functions.ts
```diff
@@ -211,6 +211,20 @@ export const compDict = {
...
 
+  /**
+   * Return a paint of none (no paint)
+   */
+  none: (): IColorV<VarAD> => {
+    return {
+      tag: "ColorV",
+      contents: {
+        tag: "NONE",
+        // Irrelevant dummy color -- paint=none is a special case in the renderer
```
wodeni, line 222: Is the dummy color needed here? Maybe the type `INone` shouldn't have `contents` in the first place?

wodeni: Also, I think `variableAD` signals to the autodiff engine that this is a variable, not a constant. `constOf(0)` is probably the better function to call if you want to create a constant.

packages/core/src/contrib/Functions.ts
```diff
@@ -669,12 +683,12 @@ export const compDict = {
    * Set the opacity of a color `color` to `frac`.
    */
   setOpacity: (color: Color<VarAD>, frac: VarAD): IColorV<VarAD> => {
-    const rgb = color.contents;
+    const props = color.contents;
     return {
       tag: "ColorV",
-      contents: {
-        tag: "RGBA",
-        contents: [rgb[0], rgb[1], rgb[2], mul(frac, rgb[3])],
```
wodeni, line 677: Since we don't have typechecking for computations, there are two near-term choices here: do nothing if the color is `INone`, or throw a runtime error.

packages/core/src/types/value.ts
```diff
@@ -223,7 +223,7 @@ export interface IHMatrix<T> {
   dy: T;
 }
 
-export type Color<T> = IRGBA<T> | IHSVA<T>;
+export type Color<T> = IRGBA<T> | IHSVA<T> | INONE<T>;
```
wodeni, line 226: A better name is perhaps `INone` or `INoPaint`. Also the `<T>` generics is probably not needed here. Some of the value types are parametrized mainly because of autodiff. See [autodiff guide](https://github.com/penrose/penrose/wiki/Autodiff-guide) for why.

packages/core/src/types/value.ts
```diff
@@ -235,6 +235,11 @@ export interface IHSVA<T> {
   contents: T[];
 }
 
+export interface INONE<T> {
+  tag: "NONE"
+  contents: T[];
```
wodeni, line 240: The `contents` field is not needed here.

wodeni (changes requested):
Looks great overall! Seems like you covered the necessary places to change. See my comments for some small suggestions.

packages/core/src/contrib/Functions.ts
cmumatt: Tend to agree: the color quad is vestigial in this case as the renderer will ignore it.  

I was similarly tempted to remove this and have an alternate local implementation to that effect, but it carries some downsides.  As we implement more aspects of SVG paint, we will have more "paint" information flowing through the middle of the system -- image and gradient paints might be future examples -- and it's a valid question about how closely we need to manage that data structure before it gets to the renderer. I.e., do we only include the relevant data in the structure (i.e., the structure for color, none, image, etc. are all different) or do we put the paint information we have in one structure and let the renderer decide what it needs, probably based on the tag.

After thinking through both implementations, it seemed simpler (short and long-term) to keep the interface structure consistent as a "paint" structure; varying it by tag triggers a fair amount of if/then/else and casting in the middle of the pipeline to deal with the structure variances based on what the type checker flagged.  I know a couple places where color/paint is typed 'any,' which leads me to think there might be other places that need to be altered to avoid runtime errors the static checker couldn't find automatically.  

The net effect either way seems to be the same: the renderer will have to decide how to apply paint at the end of the pipeline, probably based on the tag plus whichever fields in the structure it decides it needs.  Varying the structure of the type before it gets to the renderer, on balance, seemed to generate more code just to the structural difference without bringing us much benefit short or long term.  

I still have that alternate implementation and can do some more testing to try and tease out any runtime exceptions and then send it as a PR -- I am fine with that if it is the best thing for the project.  With either implementation, we can always feel free to change our minds later.

packages/core/src/contrib/Functions.ts
cmumatt: Added branch to do nothing in the no paint situation.

cmumatt: Added a branch in this routine in the updated PR to do nothing if provided the no-paint data structure.

packages/core/src/types/value.ts
cmumatt: Renamed & de-parameterized type in updated PR.

packages/core/src/types/value.ts
cmumatt: Removed in updated PR & added downstream logic where needed.

packages/core/src/engine/EngineUtils.ts
```diff
@@ -183,20 +183,17 @@ function mapPathData<T, S>(f: (arg: T) => S, v: IPathDataV<T>): IPathDataV<S> {
...
+      tag: v.tag,
+      contents: mapTuple(f, (v as any).contents),
     };
+  } else if (v.tag === "NONE") {
+    return {  
+      tag: v.tag,
+    }
   } else {
-    throw Error("unexpected color tag");
+    throw Error("unexpected color tag: " + v['tag']);
```
wodeni, line 196: More idiomatic: `v.tag`

packages/core/src/utils/Util.ts
```diff
@@ -229,20 +230,33 @@ export const hsvToRGB = (
...
+    const hsv = color.contents;
+    const rgb = hsvToRGB([hsv[0],hsv[1],hsv[2]]);
     return toHexRGB(rgb);
+  } else if (color.tag === "NONE") {
+    return "none";
   } else {
-    console.error("color type", color.tag, "unimplemented");
+    console.error("color type", color['tag'], "unimplemented");
     return "";
   }
```
wodeni, line 243: Now that you typed the input, I think this branch can be removed.

packages/core/src/utils/Util.ts
```diff
@@ -229,20 +230,33 @@ export const hsvToRGB = (
...
 };
 
+export const toSvgOpacityProperty = (color: Color<number>): number => {
+  if (color.tag === "RGBA" || color.tag === "HSVA") {
+    return color.contents[3];
+  } else if (color.tag === "NONE") {
+    return 1;
+  } else {
+    console.error("color type", color['tag'], "unimplemented");
+    return 1
```
wodeni, line 256: Similarly, this branch can go away as well.

wodeni (changes requested):
A few cosmetic changes, and we're ready to merge!

wodeni:
BTW seems like circleci now gets triggered by PR branches. The issue is it doesn't know `cmumatt:main` is not `penrose:main` :(. Going forward, I'll suggest all contributors to branch off in their fork and open a PR here (if that fixes this issue for now).

packages/core/src/engine/EngineUtils.ts
```diff
@@ -183,17 +183,9 @@ function mapPathData<T, S>(f: (arg: T) => S, v: IPathDataV<T>): IPathDataV<S> {
...
-    };
-  } else if (v.tag === "NONE") {
-    return {  
-      tag: v.tag,
-    }
-  } else {
-    throw Error("unexpected color tag: " + v['tag']);
+  switch (v.tag) {
+    case "NONE": return {tag: v.tag};
+    default:     return {tag: v.tag, contents: mapTuple(f, (v as any).contents)};
```
wodeni, line 188: I think you can explicitly write out the cases for readability:
```ts
case "RGBA": // this case will fall through
case "HSVA":
   return {tag: v.tag, contents: mapTuple(f, (v as any).contents)}; 
```
BTW do you have [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) installed? If not, I'd suggest installing it. Also a bit surprised that the pre-push hook didn't fix it for you.

packages/core/src/utils/Util.ts
```diff
@@ -231,29 +231,17 @@ export const hsvToRGB = (
...
-    return toHexRGB(rgb);
-  } else if (color.tag === "NONE") {
-    return "none";
-  } else {
-    console.error("color type", color['tag'], "unimplemented");
-    return "";
+  switch (color.tag) {
+    case "RGBA": return toHexRGB(         [color.contents[0],color.contents[1],color.contents[2]] );
+    case "HSVA": return toHexRGB(hsvToRGB([color.contents[0],color.contents[1],color.contents[2]]));
+    default:     return "none"; // NONE
```
wodeni, line 237: Similarly, `case "None"` will do the trick too

packages/core/src/utils/Util.ts
```diff
@@ -231,29 +231,17 @@ export const hsvToRGB = (
...
-  if (color.tag === "RGBA" || color.tag === "HSVA") {
-    return color.contents[3];
-  } else if (color.tag === "NONE") {
-    return 1;
-  } else {
-    console.error("color type", color['tag'], "unimplemented");
-    return 1
+  switch (color.tag) {
+    case "NONE": return 1;
+    default:     return color.contents[3];
```
wodeni, line 244: Write out the cases for readability.

wodeni (changes requested):

wodeni (approved):

PR #710, opened 2021-12-03
refactor: make `evalShapes` number type agnostic

wodeni:
# Description

Related issue/PR: #698 

The current implementation of `evalShapes` has unclear semantics: it takes in a `State` and return another. Instead, this PR refactors it (and related functions) s.t. it returns a list of shapes instead. Converstion from autodiff props to number props was done explicitly everywhere (e.g. `prepareState`).


samestep (approved):
lgtm

Issue #724, opened 2021-12-28
Polygon points are relative to top-left of canvas, not origin


samestep:
While working on https://github.com/penrose/penrose/pull/698#issuecomment-1002015939, I realized that our current bounding boxes for `Polygon`, `FreeformPolygon`, and `Polyline` are very incorrect.

<img width="1134" alt="Screen Shot 2021-12-28 at 8 38 01 AM" src="https://user-images.githubusercontent.com/8246041/147572050-4d737bb6-5a34-4e04-876b-c0ae06d771c8.png">

This is because, for some reason, the `points` and `scale` properties for those shapes are relative to the top-left corner of the canvas, not the origin `(0, 0)`.

Related:

- #709
- #713

keenancrane:
I also just hit this bug.  The polygon is given the same vertices as the three black dots i, j, k:

![image](https://user-images.githubusercontent.com/3604525/148081357-0e5060b9-458d-4bee-bee7-0f44e1843adb.png)

PR #725, opened 2021-12-28
feat: Handle strokeWidth in bbox

samestep:
# Description

Mostly resolves https://github.com/penrose/penrose/pull/698#issuecomment-1002015939

This PR accounts for the `strokeWidth` property when computing bounding boxes, as per @keenancrane's suggestion:

> So, let's consider what happens to the size of a shape if you stroke it by a stroke of width w. Giving a closed path a width w is equivalent to sweeping a circle of radius w/2 along that path. Hence, no matter which points are the extreme points along the path, the bounding box will need to grow by w/2 in all directions (top/bottom/left/right). So, just add w to both the bounding box width and height.

The only shape for which this PR does not use that strategy is `Square`, because in that case it is more correct to add half the `strokeWidth` in all directions _before_ doing the rotation instead of after.

The following shapes are accounted for:

- `Circle`
- `Ellipse`
- `Rectangle`
- `Callout`*
- `Square`

The one(s) marked with an asterisk * have newly introduced imprecision:

Screen Shot 2021-12-28 at 9 12 25 AM

The following shapes did not need to be accounted for:

- `Image`
- `Text`
- `Line`
- `Arrow`

And finally, the following shapes should ideally be accounted for, but are not in this PR for various reasons:

- `Polygon`/`FreeformPolygon`/`Polyline` because https://github.com/penrose/penrose/issues/724
- `PathString` because currently its bbox implementation just uses its `w` and `h` properties and ignores `data`
- `Path` because currently its bbox implementation already has a lot of imprecision for Bézier curves and for arcs

# Checklist

### DIFF
--- a/packages/core/src/renderer/ShapeDef.test.ts
+++ b/packages/core/src/renderer/ShapeDef.test.ts
@@ -37,6 +37,8 @@ const getShape = async (shapeSty: string): Promise<ShapeAD> => {
 const circleSty = `Circle {
   r: 100
   center: (42, 121)
+  strokeWidth: 50
+  strokeColor: rgba(0, 0, 0, 1) -- Circle strokeColor is noPaint by default
 }
 `;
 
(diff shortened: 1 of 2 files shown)


keenancrane (approved):
Looks ok to me!

PR #735, opened 2021-12-30
feat: Allow subtypes inline

### LINKED ISSUE #722: Allow subtypes inline
Currently in a Domain schema, subtype relationships have to be given on separate lines from the type declaration.  For instance, you have to write
```
type BaseType
type SubType
SubType <: BaseType
```
This can get hard to follow if you have many types, with many subtype relationships.  Basically you have to choose whether to interleave subtype statements like this:
```
type BaseType
type SubType1
SubType1 <: BaseType
type SubType2
SubType2 <: BaseType
type SubType3
SubType3 <: BaseType
```
or give them all after the initial type declarations, like this:
```
type BaseType
type SubType1
type SubType2
type SubType3
SubType1 <: BaseType
SubType2 <: BaseType
SubType3 <: BaseType
```

Would be nice to be able to give the subtype relationships on the same line as the type declaration.  E.g.,
```
type BaseType
type SubType1 <: BaseType
type SubType2 <: BaseType
type SubType3 <: BaseType
```
If there are multiple subtyping relationships, one might write
```
type BaseType1
type BaseType2
type SubType1 <: BaseType1, BaseType2
type SubType2 <: BaseType, BaseType2
type SubType3 <: BaseType, BaseType2
```

NA

**Additional context**
NA

samestep:
# Description

Resolves #722.

The first commit in this PR renamed the `type` `statement` rule to `type_decl`, because previously it was being combined with the other `type` rule below, which would have caused ambiguous parsing after adding the syntax extension this PR adds.

I'm not sure why `SubTypeDecl`'s fields are `Type` instead of `TypeConstructor` since the compiler throws an error if they aren't the latter, but I've kept that the same in this PR.

This PR updates most things in `examples/` to use the more concise syntax.

# Checklist

# Open questions

- Should we even keep around `SubTypeDecl` at all? (Note: a few places in `examples/` still use it.)
- Currently this PR only allows the comma syntax (for specifying multiple supertypes) in `TypeDecl`, not `SubTypeDecl`; should that syntax be allowed in the latter as well?
- Should the `subType`, `superType`, and `superTypes` fields have their types changed from `Type` to `TypeConstructor`?

### DIFF
--- a/packages/core/src/parser/Domain.ne
+++ b/packages/core/src/parser/Domain.ne
@@ -92,13 +92,14 @@ statement
   |  subtype     {% id %}
 
 # not to be confused with `type`, defined below
-type_decl -> "type" __ identifier (_ "(" _ type_params _ ")"):? {%
-  ([typ, , name, ps]): TypeDecl => {
+type_decl -> "type" __ identifier (_ "(" _ type_params _ ")"):? (_ "<:" _ sepBy1[type, ","]):? {%
+  ([typ, , name, ps, sub]): TypeDecl => {
     const params = ps ? ps[3] : [];
+    const superTypes = sub ? sub[3] : [];
     return { 
-      ...nodeData([name, ...params]),
+      ...nodeData([name, ...params, ...superTypes]),
       ...rangeBetween(typ, name),
-      tag: "TypeDecl", name, params
+      tag: "TypeDecl", name, params, superTypes
     };
   }
 %}
--- a/packages/core/src/parser/Domain.ne
+++ b/packages/core/src/parser/Domain.ne
@@ -83,7 +83,7 @@ statements
     |  _ statement _c_ nl statements {% d => [d[1], ...d[4]] %}
 
 statement 
-  -> type        {% id %}
+  -> type_decl   {% id %}
   |  predicate   {% id %}
   |  function    {% id %}
   |  constructor_decl {% id %}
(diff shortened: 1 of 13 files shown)


keenancrane (approved):
@wodeni will be able to better comment on how things get implemented, but the new syntax in the examples looks great!

I'm not too worried about doing stuff like
```
SubType1, SubType2 <: BaseType
```
but maybe we'll discover down the line (with further examples) that this would be nice.  So far, it's really the ability to declare a type _and_ specify its base types on the same line that seems helpful for readability (since then you get a picture of what this type is all about all in one line).

samestep:
@keenancrane Sounds good!

I'll wait for @wodeni to approve before merging this.

packages/core/src/parser/Domain.ne
```diff
@@ -83,21 +83,23 @@ statements
...
-  ([typ, , name, ps]): TypeDecl => {
+# not to be confused with `type`, defined below
+type_decl -> "type" __ identifier (_ "(" _ type_params _ ")"):? (_ "<:" _ sepBy1[type, ","]):? {%
+  ([typ, , name, ps, sub]): TypeDecl => {
     const params = ps ? ps[3] : [];
+    const superTypes = sub ? sub[3] : [];
     return { 
-      ...nodeData([name, ...params]),
+      ...nodeData([name, ...params, ...superTypes]),
       ...rangeBetween(typ, name),
```
samestep, line 101: @wodeni Is it OK that this range remains the same, or do I need to change it? I see that it already doesn't account for `params`, so I assumed it also shouldn't account for `superTypes`, but I'm not entirely sure that's correct.

wodeni (approved):
LGTM! I do like adding the `superTypes` field vs. some kind of syntactic sugar. I made this mistake when parsing `decl_list`, i.e. parse it into multiple `Decl` nodes. It's generally best when we can un-parse the AST into identical source.

packages/core/src/parser/Domain.ne
samestep: Makes sense; thanks for reviewing!

PR #737, opened 2021-12-31
feat: Use C-style syntax for Domain arglists

### LINKED ISSUE #312: Revise syntax for named parameters in Domain functions/predicates
Domain currently allows types to have names, e.g., instead of writing just

`predicate From: LinearMap * VectorSpace * VectorSpace`

one can write

`predicate From: LinearMap * VectorSpace domain * VectorSpace codomain`

However, this syntax is fairly awkward in terms of legibility.  E.g., seeing `domain * VectorSpace` juxtaposed may be jarring for someone used to a language like Haskell.  On the other hand, omitting these names altogether feels like a bad solution, since they essentially provide useful comments that give semantics to functions and predicates.

Jonathan proposes a syntax like

`predicate From(LinearMap, domain:VectorSpace, codomain:VectorSpace)`

I'm on board with this, or any variant.  To me (and I suspect others), the most natural thing would be a C-like prototype:

`predicate From(LinearMap, VectorSpace domain, VectorSpace codomain)`

But we should think about what makes most sense in Penrose.

samestep:
# Description

Resolves #312. That issue only gives examples for `predicate`, and only mentions that and `function`, but this PR also updates the syntax for `constructor`.

A few things I noticed:

- Previously we had a special case for zero args, where no colon would be used. Now that we delimit the arglist with parentheses, I removed that special case for greater consistency.
- I'm a bit confused about the syntax for what look like generics? They seem to use round parentheses in the `type` declaration, then square brackets when listing them before the arglist for a `constructor`, then round parentheses again in the actual arglist. Those latter parentheses are a bit less readable with the new arglist syntax in my opinion, but not terrible; still, it seems like it may be a better idea to make them all square brackets?
- While most (all?) of `examples/` seems to be checked by `packages/core/src/__tests__/overall.test.ts`, which is run as part of `yarn test` from the repo root, the same does not seem to be true of `packages/synthesizer/__tests__/*.dsl`.
- I made use of local search-and-replace to do these example replacements a bit faster, but it was still a fairly manual process. I seem to remember @keenancrane previously mentioning in a conversation with @wodeni that it would be nice to have tools to systematically update all our examples when we update syntax.

The `sepBy1` and `sepBy` macros previously had some bugs, which this PR fixes:

- `sepBy1` previously forbid whitespace between the last item and its trailing separator
- `sepBy` previously forbid a trailing separator
- `sepBy` previously allowed a leading separator
- `sepBy` previously allowed leading whitespace, which could cause ambiguity

They were also duplicated across all three of our preexisting nearley files, so this PR extracts their definitions out into a new `packages/core/src/parser/macros.ne` file. Previously Substance was the only language to use `sepBy` (the other two languages only used `sepBy1`), so it had a fix for a bug in that macro definition which would have caused a compile error if the macro were used in either of the other languages. Other than that, the duplicated macro definitions were already all the same.

# Checklist

# Open questions

- Do we actually _want_ `sepBy1` and `sepBy` to allow trailing separators?

### DIFF
--- a/packages/core/src/compiler/Domain.test.ts
+++ b/packages/core/src/compiler/Domain.test.ts
@@ -82,11 +82,11 @@ type Real
 type Interval
 type OpenInterval
 type ClosedInterval
-constructor Cons ['X] : 'X head * List('X) tail -> List('X)
-constructor Nil['X] -> List('X)
-constructor CreateInterval: Real left * Real right -> Interval
-constructor CreateOpenInterval: Real left * Real right -> OpenInterval
-constructor CreateClosedInterval: Real left * Real right -> ClosedInterval
+constructor Cons ['X] ('X head, List('X) tail) -> List('X)
+constructor Nil['X]() -> List('X)
+constructor CreateInterval(Real left, Real right) -> Interval
+constructor CreateOpenInterval(Real left, Real right) -> OpenInterval
+constructor CreateClosedInterval(Real left, Real right) -> ClosedInterval
     `;
     const res = compileDomain(prog);
     const types = [
@@ -111,17 +111,17 @@ constructor CreateClosedInterval: Real left * Real right -> ClosedInterval
 type Map
 type Set
 type Point
-predicate Not : Prop p1
-predicate From : Map f * Set domain * Set codomain
-predicate Empty : Set s
-predicate Intersecting : Set s1 * Set s2
-predicate IsSubset : Set s1 * Set s2
-predicate PointIn : Set s * Point p
-predicate In : Point p * Set s
-predicate Injection : Map m
-predicate Surjection : Map m
-predicate Bijection : Map m
-predicate PairIn : Point * Point * Map
+predicate Not(Prop p1)
+predicate From(Map f, Set domain, Set codomain)
+predicate Empty(Set s)
+predicate Intersecting(Set s1, Set s2)
+predicate IsSubset(Set s1, Set s2)
+predicate PointIn(Set s, Point p)
+predicate In(Point p, Set s)
+predicate Injection(Map m)
+predicate Surjection(Map m)
+predicate Bijection(Map m)
+predicate PairIn(Point, Point, Map)
     `;
     const res = compileDomain(prog);
     const types = ["Map", "Set", "Point"];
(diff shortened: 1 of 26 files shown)


packages/core/src/compiler/Domain.test.ts
```diff
@@ -168,13 +168,13 @@ type Set
   });
   test("Type not found", () => {
     const prog = `
-constructor Cons ['X] : 'X head * List('X) tail -> List('X)
+constructor Cons ['X] ('X head, List('X) tail) -> List('X)
```
wodeni, line 170: Now that I think about it, perhaps a more conventional `constructor Cons<'X> ('X head, List<'X> tail) -> List<'X>` is more readable?

wodeni: Also funny that everything is just looking more like typescript 😛

packages/core/src/parser/macros.ne
```diff
@@ -0,0 +1,13 @@
+sepBy1[ITEM, SEP] -> $ITEM (_ $SEP _ $ITEM):* (_ $SEP):? {%
+  ([first, rest]) => {
+    const restNodes = rest.map((ts: any[]) => ts[3]);
+    return concat(first, ...restNodes);
+  }
+%}
+
+sepBy[ITEM, SEP] -> null {% d => [] %} | $ITEM (_ $SEP _ $ITEM):* (_ $SEP):? {%
```
wodeni, line 8: Start each of the two cases in separate lines for readability?

wodeni: Also super great to see redundant code extracted! Somehow `@include` used to give me troubles. Maybe I never used it properly? Glad that it works now!

wodeni (approved):
💯

> Do we actually want `sepBy1` and `sepBy` to allow trailing separators?

I'm in favor of trailing separators in general. They just make formatting more flexible.

packages/core/src/compiler/Domain.test.ts
samestep: > Now that I think about it, perhaps a more conventional `constructor Cons<'X> ('X head, List<'X> tail) -> List<'X>` is more readable?

Yeah, I'd like that a lot better too I think; should I make an issue about this so it can be done in a followup PR?

packages/core/src/parser/macros.ne
samestep: > Start each of the two cases in separate lines for readability?

Done!

samestep:
> I'm in favor of trailing separators in general. They just make formatting more flexible.

Yep I agree in general. I think it might be a bit weird to have trailing separators when the separator is something like `*`, but it looks like after this PR is merged, the separator is always `,` or `;` so this seems good.

packages/core/src/compiler/Domain.test.ts
wodeni: Sounds good!

samestep: Done: #742

PR #739, opened 2021-12-31
refactor: Retype ASTs

samestep:
# Description

This PR adds types to `packages/core/src/parser/ParserUtil.ts` to get rid of all warnings in that file. Doing this required a few additional changes beyond just adding types to `ParserUtil`:

- new `SourceRange` type in `types/ast`
- slight change to the type predicate used in `rangeOf`
- rewrite of `tokenEnd` to fix bugs
- new test for `tokenEnd`

and also a substantial reworking of our AST types:

- every AST node type is now generic in whether or not it extends `SourceRange`
- `types/ast` now provides `A` and `C` types to put in that generic slot, corresponding to "abstract" or "concrete"
- every usage of those types has been updated to provide that generic type parameter
- most ended up being `A`, because even in our compilers we generate a lot of synthetic nodes
- for other types holding nodes, everything that could be `C` was made `C`, but some were forced to be `A` due to above
- error types use `A` instead of `C` just for convenience
- this rewrite found a few missing `nodeType` and `children` fields in synthetic nodes, which were fixed

# Checklist


samestep:
Oh... the build is failing mostly because all the parsers pass `ASTNode`s to these functions instead of `ConcreteNode`s, so their types say that the `start` and `end` fields are optional; I don't know how to fix this so I'm turning this PR back to a draft for now, but @wodeni if you have any ideas of how to deal with this then let me know.

packages/core/src/types/ast.ts
```diff
@@ -4,9 +4,14 @@ export interface SourceLoc {
...
 
+export interface SourceRange {
+  start: SourceLoc;
+  end: SourceLoc;
+}
+
 type Optional<T, K extends keyof T> = Omit<T, K> & Partial<T>;
 
-export type ASTNode = Optional<ConcreteNode, "start" | "end">;
+export type ASTNode = Optional<ConcreteNode, keyof SourceRange>;
```
wodeni, line 14: Now I'm seeing the problem after some testing. I think we need to think about how to design the node types again. The current types aren't super useful. Perhaps the right way to go is generics? 
* `ASTNode<T>` contains all the meta properties and data of type `T`.
```ts
interface ASTNode<T extends NodeData> {
  data: T,
  // ...other meta props
}
```
* For convinience, we also need a common super type `NodeData` for all the node data types (`Decl`, `StringLit` etc) so we can do things generically over node data. Seems like the only thing they have in common is `tag`? 
```ts
interface NodeData { 
  tag: string 
}
interface Decl extends NodeData { 
  // ...concrete data here
}
```
* When typing the parsers, we can use `ASTNode<T>` for the output of each rule

wodeni: This is probably a big change. So I guess we could do another PR for this if you think it's a good idea? Lmk if you have any suggestions!

samestep: Yep this makes a lot of sense to me; it's more or less what I was thinking about doing eventually after mulling over it for a little while. Probably should be done in a separate PR as you suggest (i.e. as a prerequisite to this PR).

wodeni (approved):

PR #743, opened 2022-01-03
refactor: Reorganize shapes

### LINKED ISSUE #700: No easy way to programmatically instantiate shapes
Working on #698 revealed that it would be nice to have an easy way to programmatically instantiate shapes. The specific use case there is unit tests, but @wodeni mentioned that there are also other contexts in which this would be useful.

In the status quo, one can programmatically generate a shape by compiling/evaluating a sub/sty/dsl trio and extracting the shapes from the resulting state.

### LINKED ISSUE #701: Individual shape defs don't have precise types
Currently we have an `IShape` type:

https://github.com/penrose/penrose/blob/5a194aac21349c5a8aafeb5e0d87144274d53b04/packages/core/src/types/shape.ts#L10-L13

We also have a `ShapeProps` type:

https://github.com/penrose/penrose/blob/5a194aac21349c5a8aafeb5e0d87144274d53b04/packages/core/src/renderer/Renderer.ts#L8-L13

These are not very precise because they don't encode anything about specific shape definitions. This results in having to unsafely typecast properties:

https://github.com/penrose/penrose/blob/5a194aac21349c5a8aafeb5e0d87144274d53b04/packages/core/src/renderer/Label.ts#L24

It would be nice if we had TypeScript type definitions for each of the shape defs, along with some way to keep those in sync with the actual shape defs.

This also came up in #698, where all the `bbox` functions type their parameter as `any` instead of something more specific.

### LINKED ISSUE #709: Polygon center does not get used
Currently, if a `Polygon` is defined with a `scale` and a `center`, the `scale` gets used as expected, but the `center` gets translated into `cx` and `cy` attributes on the SVG object, which do not change the visual shape at all.

In order to fix this, we'd first need to decide which happens first (scaling vs translation).

### LINKED ISSUE #712: Image rotation disagrees with other shapes
As discovered while working on #698, the `rotation` property on the `Image` shape acts very strangely, and in particular does not act the same way as it does for `PathString`, `Square`, and `Text`. For example, paste this Style program into https://panes.penrose.ink/:

```
canvas {
  width = 800
  height = 700
}

global {
  global.image = Image {
    path: "shading.svg"
    center: (0, 0)
    w: 100
    h: 100
    rotation: 30
  }

  global.square = Square {
    center: (0, 0)
    side: 100
    rotation: 30
  }
}
```

rendered diagram showing rotation discrepancy

### LINKED ISSUE #714: Organization of shapes and queries
Currently, much of the code for individual shapes can be found in two places:

- `renderer/ShapeDef.ts` which contains all the `shapedefs`, each of which has
  - a name
  - `properties`, each of which has
    - a `PropType`
    - a `Sampler`
- shape-specific files in `renderer/`, each of which has an `export default` for a render function

Once #698 is merged, `renderer/ShapeDef.ts` will also contain a `bbox` implementation for each shape type, and more shape queries will follow, so we need to figure out a place to put them. @joshsunshine and I have been discussing how to reorganize Penrose's shape-specific code in general.

One idea is to make a new `shapes/` directory alongside `renderer/` and move each shape from the latter into its own file in the former. We would still need a `shapes/ShapeDef.ts` file, but each individual `ShapeDef`s (and any queries defined for it, such as `bbox`) would be defined in that shape's individual file, along with that shape's render function. Josh mentioned that this could also help with documentation as requested in #570.

#558 also seems related.

samestep:
# Description

Fixes #700, fixes #701, fixes #709, fixes #712, fixes #714. Joint work with @joshsunshine and @maxkrieger.

This PR does the following:

- Remove the following shapes:
  - `Callout` (it's gross)
  - `FreeformPolygon` (it's redundant with `Polygon`)
  - `PathString` (it's redundant with `Path`)
  - `Square` (it's redundant with `Rectangle`, and it's not in SVG, unlike `Circle`)
  - `Arrow` (it's redundant with `Line`)
- Delete `renderer/ShapeDef`.
- Add a new type hierarchy for shapes:
  - Create `types/shapes` with several new `interface`s holding commonly shared properties
  - Create `shapes/` directory holding most of the other things:
    - Create `shapes/Samplers` holding samplers, which now just sample rather than returning sampler functions
    - Create a file for each individual shape type:
      - Export an interface for the properties of that shape
      - Export a sampler function to sample a random instance of that shape's properties
      - Export a type for a concrete instance of that shape
      - Export a constructor function to get a concrete instance of that shape given some subset of its properties
    - Create `shapes/Shapes`: 
      - Add a `Shape` type that is the union of all individual concrete shape types
      - Migrate a modified version of the previous `ShapeDef` type, which holds metadata about shape types
      - Export a `shapedefs` object which maps from shape type names to `ShapeDef`s
- Move [bbox computation functions](https://github.com/penrose/penrose/pull/698) into `engine/BBox`
  - Move bbox tests into the test file for `engine/BBox`
- Do a couple minor refactorings:
  - Move `propertiesOf` and `propertiesNotOf` from `compiler/Style` to `engine/EngineUtils`
  - Move `getShapeName` from `renderer/ShapeDef` to `engine/PropagateUpdate`
  - Type `State`'s `uninitializedPaths` field as `Path[]` instead of `any`
  - Move `arrowHead` and `makeRoomForArrows` from `renderer/Arrow` to `renderer/Line`
- Modify `compiler/Style` to use the new shapedefs and samplers
- Modify `renderer/Resample` to very inefficiently use the new samplers instead of the old ones (need to fix this)
- Modify all `examples/` and TypeScript code to get rid of the deleted shape types and migrate property names

<details>
<summary>expand to see subtask tracking used while developing this</summary>

- [x] `Rectangle`
- [x] the other shapes
- [x] name the properties :)
- [x] reconnect to `ShapeDef.test.ts`
- [x] samplers
- [x] connect it to the compiler
  - [x] Update Style.ts, in particular the function that initializes shapes: https://github.com/penrose/penrose/blob/d73815d07182556ac23f5a17e2dbb16a3184c5f3/packages/core/src/compiler/Style.ts#L2688-L2692
- [x] use these types in the optimizer
- [ ] use them in the renderer
  - [x] update the property names that were changed
  - [ ] ~Move content of render functions, e.g. the Arrow function below, to render methods in the Shape types:~ https://github.com/penrose/penrose/blob/d73815d07182556ac23f5a17e2dbb16a3184c5f3/packages/core/src/renderer/Arrow.ts#L84-L87
  - [ ] ~Update Render.ts, in particular the function that renders each shape to call the new render methods:~  https://github.com/penrose/penrose/blob/d73815d07182556ac23f5a17e2dbb16a3184c5f3/packages/core/src/renderer/Renderer.ts#L26
- [ ] ~(temporarily?) remove all the dragging stuff~
- [ ] ~get rid of `types/shape` probably~
- [x] fix all the bugs I introduced
  - [x] `src/compiler/Style.test.ts`
  - [x] `src/engine/BBox.test.ts`
  - [x] `src/__tests__/overall.test.ts`

</details>

# Checklist

(We don't currently check ESLint, but #755 will be a step on the path toward letting us do that. 🙂)

# Future work

- #759
- Use the new shape types throughout the codebase
- Move things from fields on the `ShapeDef` to methods on the `Shape`


packages/core/src/index.ts
```diff
@@ -31,7 +31,7 @@ import {
   prettyPrintExpr,
 } from "./utils/OtherUtils";
 import { bBoxDims, ops, toSvgPaintProperty } from "./utils/Util";
-import { Canvas } from "./renderer/ShapeDef";
+import { Canvas } from "./shapes/Samplers";
 import { showMutations } from "./synthesis/Mutation";
 import { getListOfStagedStates } from "./renderer/Staging";
 import { shapeAutodiffToNumber } from "engine/EngineUtils";
```
maxkrieger, line 37: could you also add 
```
export { shapedefs } from "./shapes/Shapes";
```
or whatever the current location is, so externals can access?

maxkrieger (approved):
LGTM

packages/core/src/index.ts
samestep: Done!

packages/core/src/compiler/Style.ts
```diff
@@ -2690,20 +2633,25 @@ const initShape = (
...
+    const instantiatedGPIProps: GPIProps<VarAD> = {
+      ...Object.fromEntries(
+        Object.entries(
+          shapedefs[stype].sampler(getCanvas(tr))
+        ).map(([propName, contents]) => [
+          propName,
+          { tag: isPending(stype, propName) ? "Pending" : "Done", contents },
+        ])
+      ),
+      ...Object.fromEntries(
```
wodeni, line 2646: My understanding is the entries above will contain all the properties in a shape, the then they will be overwritten by the entries below, if they are set by Style? Either way, it would be good to have a comment describing the intended behavior. A bit hard to decipher what this is doing.

wodeni (approved):
Very nice! I liked the design in general. I had one minor comment about `propTags` in `Shape` and `initShape` in the Style compiler. All the shape definitions look great to me!

packages/core/src/types/style.ts
```diff
@@ -372,6 +372,17 @@ export type Path =
 // Unused
 // | ITypePropertyPath;
 
+// TODO: move this somewhere outside of types/
```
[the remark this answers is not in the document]
samestep: Done!

packages/core/src/shapes/Shapes.ts
```diff
@@ -0,0 +1,169 @@
...
+  bbox: (properties: Properties) => BBox.BBox;
+  isLinelike: boolean; // TODO: use type predicate instead
+  isRectlike: boolean; // TODO: remove this
+}
+
+// hack to satisfy the typechecker
+export const ShapeDef = (shapedef: {
+  sampler: (canvas: Canvas) => unknown;
+  constr: (canvas: Canvas, properties: Properties) => Shape;
+  bbox: (properties: any) => BBox.BBox;
```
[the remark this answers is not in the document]
samestep: Yes, because `constr` receives `Partial<IWhatever>` rather than just `IWhatever`. If you replace this `any` with `Properties` then you get type errors. It doesn't matter too much though, because we're soon going to replace this with a method on the shape anyway.

keenancrane:
@samestep Confused about one thing: when we spoke about `strokeLineCaps` you mentioned that this is not yet supported by generic SVG pass through.  But it looks like @cmumatt already merged #749… right?

packages/core/src/shapes/Shapes.ts
```diff
@@ -0,0 +1,169 @@
...
+  isLinelike?: boolean;
+  isRectlike?: boolean;
+}): ShapeDef => {
+  const sampler = (canvas: Canvas) => <Properties>shapedef.sampler(canvas);
+  return {
+    sampler,
+    constr: shapedef.constr,
+
+    propTags: () => {
```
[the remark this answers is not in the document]
samestep: Good idea; done!

samestep:
> Confused about one thing: when we spoke about `strokeLineCaps` you mentioned that this is not yet supported by generic SVG pass through. But it looks like @cmumatt already merged #749… right?

@keenancrane In order to pass `strokeLineCap` through, you'd need to write it in the Style program as `stroke-linecap`, which you can't do until #759 is merged due to a limitation in the lexer.

packages/core/src/compiler/Style.ts
samestep: OK! I'll add a comment.

PR #749, opened 2022-01-04
feat: renderer: Passthrough unknown properties to Svg output

cmumatt:
# Description

This PR adapts the renderer to passthrough unknown properties to SVG output.  No validation of passthrough SVG properties or contents is currently performed in this initial support.  The general idea is to reduce the amount of code the team needs to write to accept the multitude of SVG properties we don't optimize.

If the property is unknown, has a value, and has not already been written to, this PR writes the value to SVG output.  This puts some responsibility onto the user to understand the passthrough property prior to using it, as it may cause unintended interactions Penrose does not check for.  Further, if a user mis-types a property name, we don't currently check for that, either.  The solution under discussion is to solve that in the IDE via a list of allowed properties partially imported from the SVG spec.

Where unused properties were identified (i.e., in the data structure but not doing anything), those were removed as they can be passthrough properties if needed.  

We met as a group and discussed various alternatives, including generating code to handle all properties in the SVG spec or maintaining the status quo, which involves thinking about each property before allowing its use within a style program.  This will make some things harder and may introduce hard-to-diagnose side effects in the SVG layer, but on balance this tradeoff is currently acceptable.

# Checklist

# Open questions

- We expect to provide support to the user in the future via the IDE when typing property names.
- Further work to rename internal property names is ongoing and will require updates in the renderer when complete.
- Further adaptations may be required in the compiler and optimizer to passthrough unknown properties to the renderer (work is currently active in that part of the system for more urgent items)
- A significant number of text properties and helpers can be removed after the compiler and optimizer begins providing unknown properties to the renderer.


samestep:
d3175c6145c7234971188a69cdd3de0053e2db94

keenancrane:
Sounds like a very reasonable solution.  Thank you @cmumatt!

Issue #752, opened 2022-01-04
`AutoLabel None` should define an empty label


keenancrane:
Currently, writing

```
AutoLabel None
```

in a Substance program results in the `.label` field not being created for any instance that does not receive an explicit `Label` invocation.  This behavior makes `AutoLabel None` fairly useless, because any Style block that tries to access `x.label` will fail, with an error message
```
Error looking up path 'x.label' in translation. Field 'label' does not exist.
```

A more reasonable behavior would be that `AutoLabel None` sets all label strings to the empty string, i.e., `x.label =""`.  This way, the label can still be accessed without error for instances that have been labeled.

wodeni:
Fixed by #754

PR #754, opened 2022-01-05
fix: use empty string as the default label and check autolabel statements

wodeni:
# Description

Related issue/PR: #752 

When `NoLabel` is used or no labeling statements exist for a Substance variable, the current system doesn't insert a `label` field for that variable in the `Translation`. This often causes expected errors. In this PR, we change the default to __empty strings as default labels__ even when there's no labeling statements. 

* #752 also uncovers a bug in the Substance checker: we didn't check the variables of statements like `AutoLabel A, B, C`. Added the checking in the Substance checker.
* Changed Substance postprocessing to insert default `""` labels for every variable and insert `""` for `NoLabel` statements

### DIFF
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -2008,27 +2008,24 @@ const insertNames = (trans: Translation): Translation => {
 const insertLabels = (trans: Translation, labels: LabelMap): void => {
   for (const labelData of labels) {
     const [name, label] = labelData;
-    if (label.isJust()) {
-      const labelString = label.value;
-      const labelValue: TagExpr<VarAD> = {
-        tag: "Done",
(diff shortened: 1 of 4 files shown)


keenancrane:
Awesome!  Looks good.  Thank you @wodeni!

keenancrane:
@wodeni Does `Autolabel None` no longer work as of this PR?  I get

```
Variable None (at line 12, column 11 of Substance program) does not exist.
```

wodeni:
> @wodeni Does `Autolabel None` no longer work as of this PR?  I get
> 
> 
> 
> ```
> 
> Variable None (at line 12, column 11 of Substance program) does not exist.
> 
> ```

`Autolabel None` is never supported iirc.

Issue #427, opened 2020-12-09
Style parser confuses kebab-case and subtraction


keenancrane:
If there are no spaces between a minus sign and two operands, the Style parser interprets the token as a kebab-case name.  E.g.,

`R-r`

yields an error in a context where `R` and `r` are scalars (and there is no variable named `R-r`), whereas

`R - r`

parses correctly.

**Proposal:** We have a couple of options here.  The easiest, perhaps, is to forbid variables from having kebab-case names.  Otherwise, upon failing to find a kebab-case token, the parser will have to split up the token and try parsing it as a subtraction.

wodeni:
Our new Style parser doesn't allow kebab-case anymore #428, as they lead to ambiguous grammar.

PR #759, opened 2022-01-05
feat: renderer/style: Passthrough unknown properties to Svg output

### LINKED ISSUE #767: Invalid characters permitted in style identifier names
**Describe the bug**
Characters such as [,],\,`, and ^ are ostensibly defined as allowed as the first character of a style identifier.  

This is due to an error in the identifier name regex: the first character of an identifier is specified as [A-z_], which includes characters resident between Z and a on the codepage (i.e., [,],`,^,\).

This is not as big a deal as it might initially appear: all but \ are defined as tokens elsewhere.  Trying to use them as the first character of a style identifier generates a different error.  The only one that is actually usable is \.

**To Reproduce**
Define and assign to a style identifier that begins with \.  

**Expected behavior**
[,],`,^, and \ are disallowed in the first character of a style identifier.

cmumatt:
- Disallow [,],\,^,` in first character of style identifiers (Fixes #767)
- Removed some renderer code made unnecessary by passthrough SVG
- Made Rectangle rotatable, similar to the old shape, square
- Removed 'unknown property' test

# Description

Fixes #767

Continued work on passthrough SVG properties (see PR# 749).  This PR eliminates Style's check for unknown properties and removes the test associated with the unknown property check.

Further, SVG passthrough properties make some renderer code unnecessary.  Where these properties are  rendered via the passthrough logic, specific logic for these fields has been eliminated.

This is a continuation of PR #749, which added support for passthrough properties to the renderer.  This PR makes it possible for users to utilize this functionality in a Style program, except in the situation where the property name includes dashes.  A future PR will address that issue since allowing dashes in an identifier name conflicts with the desire to allow computations in the form of x1-x2 (no spaces).

As before, there is presently no validation of passthrough SVG properties or their contents.  This is left as future work and reflects the current group consensus.  Further, there is no guarantee a passthrough property will not interfere with the optimized properties in the diagram.  Some care and knowledge of SVG may be necessary on the part of the user. 

Choose any example.  Add an unknown property such as "smileandwave" with a string literal value.  Run the example and locate the passthrough property and value on the shape within the SVG output.

# Checklist

# Open questions

See PR# 749


samestep (commented):
Looks good but let's wait until #743 merges.

packages/core/src/parser/Style.ne
```diff
@@ -36,7 +36,12 @@ const styleTypes: string[] =
 const lexer = moo.compile({
   ...basicSymbols,
   identifier: {
-    match: /[A-z_][A-Za-z_0-9]*/,
+    // Allow alpha, numbers, underscores, and dashes.  But an identified name:
+    //  - May not begin or end with a dash.
+    //  - May not begin with a number (ending with a number is ok)
+    //  - May be a single alpha character or underscore
+    //  - May begin or end with an underscore
+    match: /[A-Za-z_](?:[A-Za-z_0-9\-]*[A-Za-z_0-9])*/,
```
wodeni, line 44: In #427 we decided to disallow kebab case altogether. So I think we should remove dashes in the regex.

cmumatt: It is a good point.  Based on Nimo's comment, we re-discussed this using #427 as context.  The desire to allow passthrough SVG properties has shifted thinking somewhat, yet a desire to allow "x1-x2" type syntax for subtraction remains strong vs. requiring a clarifying space.  Given the tradeoffs, we decided to think about alternatives such as bracketing identifiers containing dashes and sleep on it.

maxkrieger:
Out of the loop: does the system warn the user when it's letting stuff pass through? Or, since the goal is to let everything pass through, it'd be too noisy?

cmumatt:
@maxkrieger, it's a good question.  Nearly all diagrams will have passthrough properties.  An example is r (radius), which is now a passthrough -- the typescript code previously mapping r is no longer necessary. 

One of the tradeoffs of the approach we discussed earlier in the week is when a user mistypes a Style property name.  There is no error or warning, which can lead to confusion.  The consensus was to (eventually) build support in the IDE to help users recognize non-Style properties since we no longer stop compilation upon encountering "unknown" properties -- instead, we pass them down the pipeline.

maxkrieger:
@cmumatt 

Also are we aware about `data-` attributes, which are the proper way to do custom data?

Also some attr names have colons like `xlink:href`

cmumatt:
The desire remains to broaden support for non-alphanumeric property names, but this PR does not implement a solution to this part of the problem.  Options discussed include:

- Expanding the definition of style identifier names to include dashes.  This breaks x1-x2 subtraction and does not accommodate all possibilities such as greek letters or namespace property qualifiers.
- Escaping with double quotes.  Since looks similar to a literal, its usage needs to be restricted to positions where literals do not occur: i.e., lvalues and the end of path statements.
- Escaping with brackets.  This was less popular than double quotes, but might have fewer restrictions than escaping with double-quotes.

packages/core/src/parser/Style.ne
samestep: Resolving this since we decided to postpone and not allow hyphens in this PR.

packages/core/src/renderer/AttrHelper.ts
```diff
@@ -127,7 +133,7 @@ export const attrXY = (
 };
 
 /**
- * Maps center, width, height, rotation --> transform !!!
```
samestep, line 130: nice

packages/core/src/renderer/AttrHelper.ts
```diff
@@ -347,7 +275,7 @@ export const attrStroke = (
 /**
  * Maps name --> new Title
  */
-export const attrTitle = (
+ export const attrTitle = (
```
samestep, line 278: interesting leading space

packages/core/src/renderer/AttrHelper.ts
```diff
@@ -414,7 +300,7 @@ export const attrStyle = (
  * like font-stretch, even though Adobe Illustrator does.
  *
  */
-export const attrFontFamily = (
+ export const attrFontFamily = (
```
samestep, line 303: another leading space

samestep (approved):
Looks great!

PR #760, opened 2022-01-06
feat: allow cycles in partial layer orderings

### LINKED ISSUE #730: Allow layering to fail gracefully
Currently if a Substance and Style program define an inconsistent ordering, the layout engine terminates with the error message
```
Error: no shape ordering possible from layering
```
This behavior is correct, but probably too strict: for programs written "in the wild," it can be (really) hard to ensure that a correct ordering will always exist.  Also, perfect layering often isn't "mission critical" for getting a diagram on the screen; more like extra polish that makes diagrams look really nice.

Would be better to fail gracefully by
- issuing a warning, rather than an error, and
- warning whatever partial ordering we can come up with (promising the user nothing about the behavior of layering, if the warning has been issued)
For instance, we can just use whatever order we've cobbled together up until the point where the warning is detected, and then give up on assembling an ordering after that.  Or just skip any offending layering command(s), and otherwise continue as normal.  The latter solution is probably better, because it enables us to stay reasonably well-ordered, even if contradictions are encountered early on.

Right now the only alternative is to either disable some layering commands globally (resulting in worse behavior than if we simply use the slightly broken ordering defined above), or go back through and meticulously arrange the Substance/Style code so that layering is consistent, which is a huge burden on the programmer.

**Additional context**
NA

wodeni:
# Description

Fixes #730 

When the partial layer orderings from Style (e.g. `A.text above A.icon`) contain cycles, our system throws an error and refuses to render a diagram since there's no topological ordering of shapes. In this PR, we introduce an approximation of topSort. Instead of failing directly, the system attempts a best-effort algorithm to approximate a global ordering. Here's a brief description from @keenancrane:

image

* We do cycle detection in `topSortLayering` and proceed as normal when the graph is a DAG
* Else, we use the new algorithm `pseudoTopsort`
* In `pseudoTopsort`, the heuristic is we always pick the node with the lowest in-degree to avoid breaking the cycles too early. The result tend to preserve most of the intended orderings.
* Added tests to `Style.test.ts`

See `Style.test.ts`.

# Checklist

# Open questions

NOTE: the current implementation can be improved by using a priority queue instead of sorting after each step.

### DIFF
--- a/examples/set-theory-domain/venn.sty
+++ b/examples/set-theory-domain/venn.sty
@@ -76,3 +76,7 @@ with Set A
 where Not(PointIn(A, p)) {
     ensure disjoint(A.icon, p.icon)
 }
+
+Set `G`, `C` {
+    `G`.icon above `C`.icon
+}
--- a/examples/set-theory-domain/venn.sty
+++ b/examples/set-theory-domain/venn.sty
@@ -76,7 +76,3 @@ with Set A
 where Not(PointIn(A, p)) {
     ensure disjoint(A.icon, p.icon)
 }
-
-Set `G`, `C` {
-    `G`.icon above `C`.icon
-}
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -33,10 +33,12 @@
     "prepack": "yarn build"
   },
   "dependencies": {
+    "@datastructures-js/heap": "^3.2.0",
     "consola": "^2.15.2",
     "eigen": "^0.2.0",
     "fast-memoize": "^2.5.2",
     "graphlib": "^2.1.8",
+    "heap-js": "^2.1.6",
     "immutable": "^4.0.0-rc.12",
     "lodash": "^4.17.15",
     "mathjax-full": "^3.0.1",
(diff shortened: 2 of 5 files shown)


keenancrane:
Awesome! A couple questions/comments:

1. The new algorithm provides a correct topological ordering when one exists. Is there a reason we still perform the cycle consistency check first, rather than always running `pseudotopsort` directly? (Would seem to simplify things.)
2. Sorting each time incurs an overall O(n^2 log n) cost, which doesn’t scale well for a large number of layered objects. Is it just the lack of a min heap w/ decrement? There seems to be a reasonable one here: < https://github.com/rombdn/js-binaryheap-decreasekey>

Thanks!

keenancrane:
>The cycle check was mainly for better error messages. I think it's still helpful for Style writers to know where the cycles are in the layer graph.

@wodeni: You can get similar information by issuing a warning whenever the minimum in degree is greater than zero.  But if the current solution works (and is asymptotically optimal) then I guess we're ok.  How are you detecting cycles in the initial phase?

samestep:
> # Open questions
> NOTE: the current implementation can be improved by using a priority queue instead of sorting after each step.

@wodeni This seems to conflict with your later comment:

> Indeed, I was debating if I wanted to write a min heap myself, but ended up using a library that's actually on npm and comes with TS support. See [f0152ed](https://github.com/penrose/penrose/commit/f0152ed3327887199c190960437e632c43f4adea).

Just to clarify: the current implementation _does_ use a priority queue instead of sorting after each step?

wodeni:
> Just to clarify: the current implementation _does_ use a priority queue instead of sorting after each step?

Yep, forgot to cross that off.

PR #777, opened 2022-01-07
feat: check existence and type of labels in Style selectors

### LINKED ISSUE #774: Make it possible to match on existence of a label in Style
Here's a strange-sounding but useful one:

**Problem.** A Style selector can match on almost anything in a Substance program—except for the existence of a label.  The idea that some instances are labeled and others are not is a pretty basic part of the design of Substance (e.g., we have the `AutoLabel All` and `AutoLabel None` mechanisms, and the ability to label only a subset of instances).  Style can match a _particular, extant_ label via the backtick syntax.  But it can't match on _the absence of a label_.

**Example.**

A basic (and very reasonable) use case is simply that you want to draw some widget if there's a label, and draw nothing if not.  For instance, a little box with some text on it.

Another example where this feature (really) matters for performance:

```
-- make sure no edge ij covers the label of
-- any other vertex k
forall Edge e; Vertex i; Vertex j; Vertex k
where e := MakeEdge(i,j) {
   
   ensure disjoint( k.bounds, e.segment, 1. )
}
```

This Style rule adds a constraint between every edge and every vertex of a mesh diagram, to ensure that the vertex label isn't covered by the edge's segment.  But if the vast majority of vertices are unlabeled, this rule is severe overkill: it adds O(VE) constraints (where V is the number of vertices, and E is the number of edges), even if only a small handful are needed.

**Proposed Solution.** I propose we add the ability to match on the absence of a label.  There are a few possibilities for syntax.  A few proposals:

```
where NoLabel(x) {
```

or

```
where NotLabeled(x) {
```

or

```
where x.label == "" {
```

Glad to hear other suggestions.  (And hopefully this one is not so hard to implement!)

wodeni:
# Description

Closes #774 

In this PR, we introduce a new Style selector pattern: `X has <math|text> label`. When included in the `where` clause of a selector, the compiler checks if `label` exists and whether it matches with the specified descriptor. Examples:

* `Label x $\vec{x}$` will match with `where x has math label`
* `Label x "xxxx"` will match with `where x has text label`
* Both of the above will match with `where x has label`
* `NoLabel x` will not match with `where x has label` nor `where x has math/text label`

Error handling:
* Only `math` and `text` are valid field descriptors. All others will result in parse errors
* `label` is the only field name allowed, but instead of throwing a parse error, the compiler reports `SelectorFieldNotSupported` error:
```
Cannot match on field x.blah (line 14, column 31 of Style program) because matching on fields is not fully supported. Currently, only "label" can be matched in selectors.
```

* Introduce `rel_field` rule in the Style parser
* Add `RelField` type in the Style compiler
* In Substance postprocessing, attach the type of the label to entries in `labelMap`
* In the Style compiler, check for the existence of `x` and the label existence and type during selector matching
* Add tests in Style parser, Substance compiler, and Style compiler

# Checklist

# Open questions

There's no infrastructure for testing selector matching results, so we are not including tests for checking selector matching results.

### DIFF
--- a/packages/core/src/compiler/Style.test.ts
+++ b/packages/core/src/compiler/Style.test.ts
@@ -7,7 +7,6 @@ import {
   parseSubstance,
 } from "compiler/Substance";
 import * as fs from "fs";
-import { Graph } from "graphlib";
 import _ from "lodash";
 import * as path from "path";
 import { Either } from "types/common";
@@ -408,7 +407,7 @@ describe("Compiler", () => {
     const errorStyProgs = {
       // ------ Selector errors (from Substance)
       SelectorVarMultipleDecl: [`forall Set x; Set x { }`],
-      SelectorFieldNotSupported: [`forall Set x where x has randomfield`],
+      SelectorFieldNotSupported: [`forall Set x where x has randomfield { }`],
 
       // COMBAK: Style doesn't throw parse error if the program is just "forall Point `A`"... instead it fails inside compileStyle with an undefined selector environment
       SelectorDeclTypeMismatch: [`forall Point \`A\` { }`],
--- a/packages/core/src/compiler/Style.test.ts
+++ b/packages/core/src/compiler/Style.test.ts
@@ -408,6 +408,7 @@ describe("Compiler", () => {
     const errorStyProgs = {
       // ------ Selector errors (from Substance)
       SelectorVarMultipleDecl: [`forall Set x; Set x { }`],
+      SelectorFieldNotSupported: [`forall Set x where x has randomfield`],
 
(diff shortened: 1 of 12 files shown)


samestep:
Ah so you changed `if/`else if`/`else` to `switch` to avoid having the `else` case which `throw`s? Nice 👌

wodeni:
> Ah so you changed `if`/`else if`/`else` to `switch` to avoid having the `else` case which `throw`s? Nice 👌

Yep! Still puzzling why there's a distinction in the first place, but yeah it's better to have static errors when types change, instead of these weird exceptions.

samestep (approved):
Looks good to me! :+1:

PR #799, opened 2022-01-10
ci: Cancel running CI on push to PR

samestep:
# Description

In the event that commit is pushed to a PR while CI is still running for a previous commit in that PR, this causes the CI for that previous commit [to be canceled](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency). The motivation for this is to help us avoid hitting our limit of [20 concurrent jobs](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration) now that we run 5 jobs at a time per commit as of #795.

Example: https://github.com/penrose/penrose/runs/4766192401?check_suite_focus=true

The downside of this is that [GitHub sends you an email](https://github.community/t/suppressing-run-cancelled-email/17940) when a workflow is canceled, just like it sends emails for workflows that actually fail. Some people (myself included) find this really annoying, so before we merge this, we should decide whether it's worth the tradeoff.

# Checklist


samestep:
Closing this for now since it doesn't seem to be a bottleneck.

PR #825, opened 2022-01-12
feat: resolve paths for included SVGs

### LINKED ISSUE #723: Fix SVG image referencing
Style supports an `Image` shape, but currently the only SVGs that can be imported come from a small, fixed directory of images in `penrose/packages/core/assets` (such as `shading.svg`, `sun.svg`, etc.).  Even to use images added to this directory, it appears that maybe (?) the Penrose package has to be re-built via `yarn`?  But even the procedure for doing that is murky.

It should be possible to load SVGs from any path, and to do so without re-building Penrose.  For instance, the Style code
```
m.background = Image {
      path : "atom.svg"
      center : (?,?)
      w : Global.atomRadius
      h : Global.atomRadius
   }
```
should use the image in the file `atom.svg`.  There is of course a question of what the default path should be.  Here we may want to consider how external paths are handled in SVG (and all the ensuing security issues...).  But a reasonable scheme might be:

- If an absolute path is given (i.e., a path starting with `/...`), use it.
- If a relative path is given (i.e., a pthat starting without `/...`), then the default path is the same as the path to the `.sty` file.
- If a relative path is given, and there is no actual `.sty` file (e.g., it was piped in from the command line), then the default path is just the root directory of the system — _this is a corner case, and we shouldn't design around it_.

Ultimately we may also want to add

- If a URL is given (i.e., if the path starts with `http`), try to access this resource and use it.

But see above the comment about security issues.

As described above, we've tried adding SVG files to the Penrose assets directory and rebuilding, to no avail.

maxkrieger:
Closes #723 

- [ ] Standardize HTTP support?
- [x] Path is in relation to `roger`'s running. Should be in relation to .sty, use a path parser?
- [x] Components support
- [x] Show checkerboard image for null results


maxkrieger:
If you `cd examples/structural-formula` and run `npx roger watch molecules/acetic-acid.sub structural-formula.sty structural-formula.dsl`, the images appear in front of the atoms, despite the layering directive:

`layer n.background below n.icon`

Behavior's normal when `n.background` is a circle shape background.

Is layering broken for images?

wodeni (approved):

Issue #741, opened 2022-01-03
Bounding box not computed for Text GPI


keenancrane:
**Describe the bug**
As of #740, there is a new `Text` GPI that renders SVG text, rather than MathJax formulas (which are now exposed via the renamed `Equation` GPI).  However, this GPI is missing a method to correctly compute its bounding box.  As a result, it cannot be properly optimized.

**Additional context**
There is code for correctly computing the bounding box already, commented out in `Text.ts` (and copied below), by constructing a temporary SVG.  (The reason we need to build and draw this temporary SVG is that the dimensions of an element are not available until it is actually rendered.)

However, there is one small snag: since `shape` is not passed by reference to the `Text` constructor, we cannot set the `.w` and `.h` fields of the Penrose `Text` GPI (which is used for optimization).  Need to figure out where and when to call this code.

A secondary issue is that the solution depends on the browser, which is needed for the `getBBox()` method.  This solution won't work in headless mode.  However, we should get the interactive version up and running first, and then worry about a headless solution.

```
/**
* This snippet correctly gets the bounding box, but
* doesn't work here since `shape` isn't passed by reference.
* Hence, we can't set the width/height of the GPI (for
* optimization).  Will have to put this somewhere else...
*
* Since the SVG hasn't been rendered yet, we temporarily
* create an invisible SVG containing just this text
* in order to determine its bounding box.
var tempDiv = document.createElement("div");
tempDiv.setAttribute("style", "position:absolute; visibility:hidden; width:0; height:0");
var tempSvg = document.createElementNS("http://www.w3.org/2000/svg", "svg");
tempSvg.appendChild(elem.cloneNode(true));
tempDiv.appendChild(tempSvg);
document.body.appendChild(tempDiv);
var bbox = tempSvg.getBBox();
document.body.removeChild(tempDiv);
shape.w = bbox.width;
shape.h = bbox.height;
console.log( "text bbox width  : " + String(bbox.width) );
console.log( "text bbox height : " + String(bbox.height) );
```

PR #829, opened 2022-01-12
feat: compute `Text` bounding box 

wodeni:
# Description

Related issue/PR: #740 
Fixes: #741

In #740, we separated out the MathJax generated labels to be `Equation` and  made `Text` to include more SVG attributes for regular text labels. For `Equation`, we compute the label dimensions using MathJax. This PR adds a step in `collectLabels` to compute the bounding box of a `Text` GPI.  

Since the SVG `<text>` has different defaults from `<rect>`, this PR also documents how we handle it differerntly from `<rect>`. 

* Thanks to @joshpoll's suggestion, we are using `canvas`'s `measureText` function to compute the dimensions of the text, notably the `width` and `height` of the bounding box.
* The support for `measureText` varies across platforms. We depend on the `actualBoundingBox*` properties of [`TextMetrics`](https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics), which has good support in Chrome, Firefox, Safari, and [`node-canvas`](https://www.npmjs.com/package/canvas).
* Added `width` and `height` of `Text` to `pendingProperties` so the Style compiler will insert them into the `Translation`. 
* In `collectLabels`, we call the text measurement function for `Text` shapes and collect the `width` and `height` for rendering. 
* We define the `center` of `Text` in Penrose to be the center of its bounding box. Note that this is different from the SVG definition of `x` and `y` of `<text>`, "the starting point of the baseline." (https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text).
* To get the desired behavior of `center`, the default properties of `Text` matter. Below are the important settings. Any changes to these settings in Style will impact where `center` is.
  * `textAnchor: middle`
  * `alignmentBaseline: middle`
  * `dominantBaseline: middle`
* When rendering `Text`, `x` and `y` are transformed to screen coordinates and rendered directly.
* `measureText` expects a single `font` string (as defined in https://developer.mozilla.org/en-US/docs/Web/CSS/font). The `toFontRule` function compose this string by aggregating these properties: `fontFamily, fontSize, fontStretch, fontStyle, fontVariant, fontWeight, lineHeight`. This function is also used in the renderer.

# Checklist

# Open questions

* We still need to add tests for the bbox function. What's the best way to test it?
* In `node-canvas`, monospace fonts like `Courier` will not have accurate bboxes. Not sure why.


maxkrieger (approved):
LGTM

Issue #818, opened 2022-01-11
fix: Fix typescript build warnings/errors in browser-ui package


cmumatt:
**Describe the bug**
Presently browser-ui generates a large number of build warnings/errors; consequently, CI is configured to not fail in this situation.  This is obviously not optimal.  Address the warnings in the code and remove the exception in CI.

**To Reproduce**
`yarn build`
Observe build warnings for browser-ui

**Expected behavior**
No build errors or warnings should be issued.  CI should reject builds with warnings / errors.

cmumatt:
Resolved by PR #839

PR #839, opened 2022-01-14
fix: Resolve browser-ui build warnings

cmumatt:
# Description

Closes issue #818

Resolves a number of typescript build errors in the browser-ui component: unsafe any types, unused variables, missing function return types, unsafe non-null assertions, etc.

This is a cleanup activity to help developers (and CI) detect newly-introduced build errors and warnings. 

# Open questions

Similar issues in synthesizer-ui will be addressed next.


samestep:
@cmumatt I see that you have a few commits re-running Prettier to get the `format` job to pass; I'm guessing that's due to VS Code producing different results from `yarn format`? @wodeni and I ran into similar issues in #836 (very annoying!), and we found that a solution is to be more explicit in our `.prettierrc`, to get VS Code to pick up on it; that's why that PR explicitly sets [`arrowParens`](https://prettier.io/docs/en/options.html#arrow-function-parentheses) and [`trailingComma`](https://prettier.io/docs/en/options.html#trailing-commas) to their default values.

On my machine I'm seeing similar issues in `CompGraph.tsx`:

https://github.com/penrose/penrose/blob/e48942461a4e21057b7ab4cf8252b7217a083cb0/packages/browser-ui/src/inspector/views/CompGraph.tsx#L114-L116

VS Code tries to reformat those lines like this:

```typescript
  const graphRef: React.RefObject<HTMLDivElement> = React.useRef<
    HTMLDivElement
  >(null);
```

I'll see if I can figure out what Prettier option to explicitly set to resolve this discrepancy.

samestep:
@cmumatt On a different note, I tried pushing a commit just now to see if getting rid of these build warnings would allow us to get rid this part of our CI config:

https://github.com/penrose/penrose/blob/c9cc55f95a58dc3d43a1527761252ab505a94e63/.github/workflows/build.yml#L10-L11

I tried pushing a [commit](https://github.com/penrose/penrose/pull/839/commits/e48942461a4e21057b7ab4cf8252b7217a083cb0) to this branch removing those lines, but [the build still seems to fail](https://github.com/penrose/penrose/runs/4817756600?check_suite_focus=true):

```
@penrose/browser-ui: Treating warnings as errors because process.env.CI = true.
@penrose/browser-ui: Most CI servers set it automatically.
@penrose/browser-ui: Failed to compile.
@penrose/browser-ui: ../core/build/dist/index.esm.js
@penrose/browser-ui: Critical dependency: the request of a dependency is an expression
@penrose/browser-ui: error Command failed with exit code 1.
```

Any idea if we can fix this? If not then we can just remove my commit (so, add that `env` setting back) and merge this anyways.

samestep:
OK, after getting rid of Prettier 1.19.1 from the browser-ui `devDependencies`, VS Code still disagrees with `yarn format` but now in a different way:

```typescript
  const graphRef: React.RefObject<HTMLDivElement> =
    React.useRef<HTMLDivElement>(null);
```

I'll continue investigating...

samestep:
OK, let's ignore the Prettier stuff for now; I'm going to try to fix it by upgrading Prettier to the latest version, but it looks like that'll introduce a bunch of formatting changes to our codebase so I'll do it in a separate PR.

cmumatt:
@samestep sounds good.  I will also break the post-ts->js warnings into a separate PR, similar to the build errors in synthesizer-ui.  Thanks for looking into the prettier item.  It took me a little bit to realize the IDE was formatting differently cli expects.

PR #842, opened 2022-01-14
fix: account for descent in `Text` center computation

wodeni:
# Description

Related issue/PR: #829 

In #829, we have `middle` as the default value for both `dominant-baseline` and `alignment-baseline` in Penrose. However, when computing the bbox with canvas, we use `alphabetic` as the baseline setting. As noted in [this article](http://bl.ocks.org/eweitnauer/7325338), `alphabetic` is a more consistent setting across platforms. 

In this setting, the relationship among the baseline, height, and descent are shown below.

image

As noted in #829, the `center` of `Text` is defined to mean the geometric center (i.e. the center of bounding box). Therefore, this PR adds an adjustment in the renderer that would compute the y coordinate of the center at render time.

- The baseline is set to `alphabetic` , in both the Canvas query and the SVG XML baseline attribute.
- The `y` location of the text is shifted down by `height/2 - descent`
- Tested with Firefox, Chrome, Safari, and Node. Results are consistent.

# Checklist


keenancrane:
LGTM!

Given that we've gone through all this trouble, might it make sense to expose an attribute `Text.descent` in Style?  I can imagine this being useful for people who want to do fancy things with text in their diagrams.

Issue #820, opened 2022-01-11
fix: Property consistency: strokeDasharray and strokeLinecap


cmumatt:
**Describe the bug**
Generally we intend to be consistent relative to SVG: if SVG already has a property name defined, we intend to  align to that given SVG is standardized and widely used.  Given this inclination, the following properties need to be adjusted slightly to be consistent with our current convention:

- strokeDashArray
   - SVG property name: stroke-dasharray
   - Rename to: strokeDasharray
- strokeLineCap
   - SVG property name: stroke-linecap
   - Rename to: strokeLinecap

Penrose does not allow dashes in style identifier names; hence, we cannot directly use SVG  names in this case.

PR #850, opened 2022-01-14
fix: strokeDasharray strokeLinecap consistency

cmumatt:
# Description

Closes issue #820 by addressing case consistency for style properties strokeDasharray and strokeLinecap.

This is a consistency improvement to benefit users as previously it was necessary to remember the odd case of these two properties relative to the corresponding SVG properties.

# Checklist


samestep (approved):
Awesome, thanks for fixing this!

PR #865, opened 2022-01-18
build: improve build time and add `clean` scripts 

wodeni:
# Description

Some our packages use create-react-app for the production build, and they slow the CI build workflow down significantly. The reason is CRA uses Webpack for bundling, which is very slow. To work around this, we have been using ESBuild as the bundler in live-reload mode, but not the build. This uses ESBuild for all packages in both live-reload and build modes.

Also, this PR adds `clean` scripts to all packages, so that when we run `yarn clean` at the top level, all build artifacts will be removed. 

* The `clean` scripts use `rimraf` to remove files
* In the top-level `package.json`, we also remove the global `node_modules` because [`lerna clean` does not remove it](https://github.com/lerna/lerna/tree/main/commands/clean#usage).
* Except for `core` (which uses both ESBuild for ESM and CJS builds and webpack for UMD build), all packages' build scripts are renamed to `build.js` for consistency.
* For easier deployment, `browser-ui` and `synthesizer-ui` build their `app.js/css` under `public`, instead of `build`, just like what `@penrose/panels` does.

# Checklist

# Open questions

N/A


samestep:
> * Except for `core` (which uses both ESBuild for ESM and CJS builds and webpack for UMD build), all packages' build scripts are renamed to `build.js` for consistency.

What do ESM and CJS and UMD stand for? 😅

wodeni:
> What do ESM and CJS and UMD stand for? 😅

[Here](https://dev.to/iggredible/what-the-heck-are-cjs-amd-umd-and-esm-ikm) is a quick intro. Basically we use ESM for browser, CJS for node, and UMD for vanilla JS to include as `<script>`.

samestep:
@wodeni You mentioned that I should play around with the scripts in this PR; just now I ran `yarn clean` at the repo root, then ran `git clean -xdn`, which printed this output:
```
Would remove packages/browser-ui/build/
Would remove packages/panels/public/app.css
Would remove packages/panels/public/app.js.map
Would remove packages/synthesizer-ui/build/
```
Is this expected, or should `git clean -xdn` be giving no output after `yarn clean`?

wodeni:
Good catch @samestep. Look what I learned today: `git clean -xdn`! Anyway, this revealed a few typos. Fixed now.

samestep:
I'm curious, what's the meaning of the 100755 → 100644 permissions change on `packages/synthesizer-ui/public/index.html`?

samestep:
> Good catch @samestep. Look what I learned today: `git clean -xdn`! Anyway, this revealed a few typos. Fixed now.

Ah cool! After pulling 8632b6bc9fa75afd4a144925a4c3ba3c3eded1f2, `git clean -xdn` after `yarn clean` still prints some lines, albeit fewer:
```
Would remove packages/browser-ui/build/
Would remove packages/browser-ui/public/app.css
Would remove packages/synthesizer-ui/build/
```

wodeni:
> I'm curious, what's the meaning of the 100755 → 100644 permissions change on `packages/synthesizer-ui/public/index.html`?

Huh no idea. How's git tracking the file permission?

samestep (approved):
Seems awesome!

wodeni:
> > Good catch @samestep. Look what I learned today: `git clean -xdn`! Anyway, this revealed a few typos. Fixed now.
> 
> Ah cool! After pulling [8632b6b](https://github.com/penrose/penrose/commit/8632b6bc9fa75afd4a144925a4c3ba3c3eded1f2), `git clean -xdn` after `yarn clean` still prints some lines, albeit fewer:
> 
> ```
> Would remove packages/browser-ui/build/
> Would remove packages/browser-ui/public/app.css
> Would remove packages/synthesizer-ui/build/
> ```

Awesome. Just fixed the css thing. The `build` folders you'll need to manually delete, because I changed the build directory for these two packages.

samestep:
> Awesome. Just fixed the css thing. The `build` folders you'll need to manually delete, because I changed the build directory for these two packages. They won't be generated in the future.

OK cool!

Actually, I just realized: probably the reason `lerna clean` doesn't remove the root `node_modules/` is that it removes the ability to run `lerna clean`, making our `yarn clean` command not idempotent (you can't run it twice in a row); is this OK?

wodeni:
@samestep I was wondering the same thing. Umm not so sure. Happy to remove the global delete if you think that's better.

samestep:
> @samestep I was wondering about the same thing. Umm not so sure. Happy to remove the global delete if you think that's better.

I have no preference whatsoever; I'm not going to use these `yarn clean` scripts at all because `git clean` is much more general and doesn't require maintenance on our part.

wodeni:
I'll just remove the global delete part of it. Agreed that `git clean` seems more complete. Still good to have `yarn clean` I guess.

maxkrieger (approved):
LGTM! Nice n clean

PR #810, opened 2022-01-11
feat: Line polygon constraints

jiriminarcik:
# Description

Related PRs: #648, #751 and #791

Added constraints for `Linelike`x`Polygon` and `Linelike`x`Rectlike` shape combinations. Added shape typing to `ConstraintsUtils.ts`. Unified Minkowski penalties interface for polygons.

# Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] My code follows the style guidelines of this project (e.g.: no ESLint warnings)

PR #868, opened 2022-01-22
feat: Polygon contains

jiriminarcik:
# Description

This PR tackles Open problems described in #810.

The constraint `contains` has been added for the following shape combinations:
* `Polygon` - `Polygon`
* `Polygon` - `Circle`
* `Circle` - `Polygon`

The containment inside of a polygon is handled using the half-plane signed distance function implemented for Minkowski penalties. Containment of a polygon is reduced to containment of all of its vertices and the containment of circle is handled as a containment of a point with padding equal to its radius.

This PR also includes a small fix for correct rendering of polygons. Namely, the property `scale` has been added to list `unoptimizedFloatProperties`  in `packages/core/src/compiler/Style.ts`. Without it, the polygon vertices may be rendered properly only on the first frame.

# Checklist

# TODO

* Unify the implementation of `padding` parameter across all constraints.


packages/core/src/contrib/Minkowski.ts
```diff
@@ -192,3 +192,40 @@ export const rectangleSignedDistance = (
...
+ * @param point Testing point.
+ * @param padding Padding applied to the polygon.
+ */
+const containsConvexPolygonPoints = (
+  p1: VarAD[][],
+  p2: VarAD[],
+  padding: VarAD
+): VarAD => {
+  const center = ops.vdiv(p1.reduce(ops.vadd), varOf(p1.length));
+  const sides = Array.from({ length: p1.length }, (_, i) => i).map((i) => [
```
wodeni, line 208: Had to look up how array-like objects work to understand this. For best readability (and potentially performance?), perhaps a simple loop would be better.

wodeni (approved):
Looks great to me! Just had one small comment on the `Array.from` magic. Hopefully we'll get rid of the hard coded list of uninitialized float props soon 😁.

packages/core/src/contrib/Minkowski.ts
jiriminarcik: Hopefully improved the readability a bit and added a clarifying comment. From some basic experiments, the performance should not be affected. It also does not matter here.

PR #873, opened 2022-02-01
refactor: Prefer undefined over null and MaybeVal

samestep:
# Description

Currently we have four different ways to represent "either something of type `T`, or nothing":

1. `T | undefined`
2. `T | null`
3. `Maybe<T>`, where `Maybe` is from [`true-myth`](https://www.npmjs.com/package/true-myth)
4. `MaybeVal<T>`, where `MaybeVal` is from `@penrose/core`

Options 1 and 2 are unavoidable due to signatures of functions built into JavaScript, the DOM, React, etc. There isn't really a reason to use 3 or 4. This PR replaces all of these with option 1 wherever possible, using option 2 where necessary.

# Checklist

### DIFF
--- a/packages/automator/index.tsx
+++ b/packages/automator/index.tsx
@@ -246,7 +246,7 @@ const singleProcess = async (
     }
 
     // HACK: return empty metadata??
-    return null;
+    return undefined;
   }
 };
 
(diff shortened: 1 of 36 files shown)


wodeni (approved):
LGTM! Honestly without nice monadic things it doesn't make a lot of sense to wrap things around with `Maybe`. I'm also open to refactor the `Either`s if you know a better way to handle them. The `true-myth` types  don't seem to work well with the client packages :(.

PR #876, opened 2022-02-03
test: Check word cloud example output in CI

samestep:
# Description

This is a followup to #875.

The determinism PR #864 started tracking the SVG outputs of all our diagrams in `examples/registry.json`, which ensures that future changes to Penrose don't inadvertently break any of those existing diagrams. Going forward, we would like to be able to similarly maintain new diagrams as we add them.

Thus, this PR adds `examples/word-cloud` to `examples/registry.json`, and replaces the custom screenshot with a link to the generated `diagrams/word-cloud-example-word-cloud.svg` which gets checked in CI.

# Open questions

- One downside of directly linking to our SVGs in GitHub Markdown files is that, while they look great in GitHub light mode, they don't look so good in dark mode. I tried setting `style="background-color:white;"` on the `img` tag, but that doesn't seem to show up on GitHub (even though it shows up for me locally in VS Code), so I removed that.
- I remember we recently removed history in order to greatly improve performance, which I think was the right decision. However, now that we have determinism, could we add back history, but have it be just for the variation textbox? This would be useful when clicking resample repeatedly, in case one sees a variation they like but accidentally click resample again without memorizing the variation string.
- Our schema for the registry is extremely cumbersome. Could we change it to just be a flat list of trios, where each substance/style/domain is what we currently have as the URI?


samestep:
Closing this for now because after some discussion with @wodeni and others, it's unclear whether there's a good way to resolve this issue with text determinism in CI.

PR #880, opened 2022-02-07
fix: Remove duplicate 'pow' conditional

### LINKED ISSUE #869: fix: Duplicate code for op 'pow' in AutoDiff.ts
**Describe the bug**
The conditional for op 'pow' is defined twice in the same if chain -- at line 1027 and 1041 -- in AutoDiff.ts.  Either code location may be removed to improve go-forward maintainability.

https://github.com/penrose/penrose/blob/9b7ad9932ff7c9f0f61cade4a36514a1e5d16aad/packages/core/src/engine/Autodiff.ts#L1027-L1028

https://github.com/penrose/penrose/blob/9b7ad9932ff7c9f0f61cade4a36514a1e5d16aad/packages/core/src/engine/Autodiff.ts#L1041-L1042

**To Reproduce**
The code in both locations is presently the same.  No error is externally visible at this time, but the duplicated code is redundant and needs to be removed to avoid future developer errors / confusion.

cmumatt:
# Description

Closes #869 

This is a non-functional change to remove a duplicate 'pow' conditional that could be confusing to future developers.  

# Checklist

### DIFF
--- a/packages/core/src/engine/Autodiff.ts
+++ b/packages/core/src/engine/Autodiff.ts
@@ -1033,8 +1033,7 @@ const traverseGraph = (i: number, z: IVarAD, setting: string): any => {
       stmt = `const ${parName} = ${childName0} / (${childName1} + ${EPS_DENOM});`;
     } else if (z.op === "atan2") {
       stmt = `const ${parName} = Math.atan2(${childName0}, ${childName1});`;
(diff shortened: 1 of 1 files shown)


wodeni (approved):
Good catch @cmumatt!

samestep (approved):
very cool :100:

PR #882, opened 2022-02-08
docs: Add demo to homepage

samestep:
# Description

This PR adds a new `Demo` component which cycles through a fixed collection of trios/variations (changing every 5 seconds), showing the Substance program and animating the diagram's convergence at 60 fps.

We expected that it would be straightforward to embed a React component in the Docusaurus site, but it was not. @wodeni and @maxkrieger put in a lot of effort to get this working, including replacing the build systems for a few of our packages. This also involved an upgrade to (among other dependencies) Docusaurus itself, which required us to update the filenames of some of our images from the tutorials.

Also, as part of this PR, I essentially completely rewrote the `Simple` Penrose diagram component. Now it has more features and also implements its React state management in a somewhat more idiomatic way.

# Checklist

# Open questions

The `Demo` component UI is still pretty barebones:

- its size on the Docusaurus site is defined in terms of pixels
- it doesn't have any buttons to navigate among the different examples
- it doesn't give any way to look at the underlying Domain or Style programs
- it doesn't give any visual indication of the 5-second countdown

@joshsunshine suggested that perhaps @hsharriman might be able to help improve this UI in the future.

Also, we need some spicier diagrams to actually show in the `Demo` component on the homepage. Perhaps @keenancrane has some up his sleeve?


maxkrieger:
btw I disabled netlify because they greedily tried to bill us just for using the free version - we'll want to switch to cloudflare pages (and nestle panes in the same build since it doesn't support monorepos)

packages/components/src/Listing.tsx
```diff
@@ -0,0 +1,48 @@
...
+      monaco.languages.setLanguageConfiguration("substance", SubstanceConfig);
+    }
+  }, [monaco]);
+  return (
+    <MonacoEditor
+      value={value}
+      width={width}
+      height={height}
+      defaultLanguage="substance"
+      // HACK
```
wodeni, line 42: This actually typechecks. You can remove the `HACK` and cast to `any` here.

packages/components/src/Listing.tsx
```diff
@@ -0,0 +1,48 @@
...
+};
+
+const Listing = ({
+  value,
+  width,
+  height,
+}: {
+  value: string;
+  width: string;
+  height: string;
```
wodeni, line 27: Weird that this component is not taking in an `Env`. Don't we need that for syntax highlighting? In fact, I'm not seeing any highlighting in all examples.

packages/components/src/Simple.tsx
```diff
@@ -3,92 +3,124 @@ import {
...
+    const compilerResult = compileTrio(this.props);
     if (compilerResult.isOk()) {
       // resample because initial sampling did not use the special sampling seed
-      const initState: PenroseState = resample(
-        await prepareState(compilerResult.value)
-      );
-      const stepped = stepUntilConvergence(initState);
+      this.penroseState = resample(await prepareState(compilerResult.value));
+    } else {
+      console.log(showError(compilerResult.error));
```
wodeni, line 38: Not in scope: should probably do better error handling like rendering errors on the canvas directly.

packages/components/src/Simple.tsx
```diff
@@ -3,92 +3,124 @@ import {
...
+    if (!this.props.animate) {
+      await this.converge();
+    }
+    this.renderCanvas();
+    this.timerID = window.setInterval(() => this.tick(), 1000 / 60);
   };
 
-  renderCanvas = async (state: PenroseState | undefined) => {
+  componentDidUpdate = async (prevProps: ISimpleProps) => {
+    if (
```
wodeni, line 74: A little surprised that prop updates need to be handled this manually. Wondering if we are leveraging the React lifecycle fully.

packages/components/src/Demo.tsx
```diff
@@ -0,0 +1,51 @@
...
+      style={{
+        display: "flex",
+        flexDirection: "row",
+        height: "100%",
+        flex: 1,
+      }}
+    >
+      <Listing value={example.sub} width={props.width} height={props.width} />
+      <div style={{ width: props.width, height: props.width }}>
+        <Simple
```
wodeni, line 38: Nice refactoring of `Simple`! Maybe this is not "simple" anymore... we should probably rename it to `Diagram` or something.

packages/components/src/stories/PenrosePrograms.tsx
```diff
@@ -1,12 +1,12 @@
 export const oneSet = {
```
wodeni, line 1: Seems like there's a way to reference to static files in storybook. Perhaps we can just load files from `examples` directly: https://storybook.js.org/docs/react/configure/images-and-assets#serving-static-files-via-storybook-configuration.

packages/components/build.js
```diff
@@ -9,6 +9,9 @@ const common = {
   define: {
```
wodeni, line 9: Now that everything is vite'd. Remove this config?

packages/docs-site/src/pages/index.js
```diff
@@ -2,7 +2,8 @@ import Link from "@docusaurus/Link";
 import useDocusaurusContext from "@docusaurus/useDocusaurusContext";
 import Layout from "@theme/Layout";
 import clsx from "clsx";
-import React from "react";
+import * as React from "react";
+import DemoWrapper from "../components/DemoWrapper";
```
wodeni, line 6: Fix import: `@penrose/components`

wodeni (changes requested):
Looks great to me! Everything very clean :P. The syntax highlighting issue (see below) is the only thing to look into for me.

packages/docs-site/src/pages/index.js
samestep: incorrect

packages/components/src/Listing.tsx
samestep: fixed

packages/components/src/Listing.tsx
```diff
@@ -0,0 +1,85 @@
...
+  substance,
+  width,
+  height,
+}: {
+  domain: string;
+  substance: string;
+  width: string;
+  height: string;
+}) => {
+  const env = compileDomain(domain).unsafelyUnwrap();
```
wodeni, line 36: Not in scope: same thing as the diagram component, some error handling would be good.

wodeni (approved):
Looks great to me! Let's merge it

PR #896, opened 2022-02-25
refactor: Replace eigen with math.js

samestep:
# Description

This PR replaces our dependency on [eigen](https://www.npmjs.com/package/eigen) with [mathjs](https://www.npmjs.com/package/mathjs), which

- is much more popular,
- has TypeScript declarations, and
- doesn't require an `await` `ready` step.

**Reviewer feedback is needed before merging this!** Specifically, I have _not_ done any performance benchmarking. I don't know the history behind why we selected Eigen.js in the first place; it's possible that it was consciously selected over math.js for performance (see, for instance, [this page](https://bertrandbev.github.io/eigen-js/#/benchmark)), in which case we should not merge this PR.

# Checklist

# Open questions

It's a bit annoying that we need a `checkMatrix` function. I opened https://github.com/josdejong/mathjs/pull/2448 to improve the math.js type declarations a bit. Note that the `identity` and `zeros` function type declarations can't easily be improved because they depend on [`config.matrix`](https://mathjs.org/docs/core/configuration.html).


samestep:
Closing this for now because #903 eliminates a big part of the motivation for this PR. Can reopen later if we want.

PR #893, opened 2022-02-21
feat: Add examples with nonconvex shapes

jiriminarcik:
# Description

Followup to #877.

Added new examples with non-convex shapes and fixed some issues with the convex partitioning algorithm by switching to library `poly-partition` (see comments below).

# Examples with steps to reproduce them

- Triangle maze example:

Run `npx roger watch maze.sub maze.sty maze.dsl` in folder `examples/minkowski-tests/maze`.

Screenshot 2022-02-21 at 1 12 53

- Yin-Yang example:

Run `npx roger watch yin-yang.sub yin-yang.sty yin-yang.dsl` in folder `examples/minkowski-tests/yin-yang`.

image

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new ESLint warnings
- [x] I have reviewed any generated changes to the `diagrams/` folder

# Open questions

The optimiser struggles with the Yin-Yang example and often get stuck at an undesirable local minimum.

Screenshot 2022-02-28 at 20 55 36

PR #901, opened 2022-03-02
feat: Allow convex partition on clockwise polygons

samestep:
Followup to https://github.com/penrose/penrose/pull/893#issuecomment-1055616275, made possible by https://github.com/x6ud/poly-partition-js/pull/1. Thanks @x6ud!


packages/core/src/contrib/__tests__/Minkowski.test.ts
```diff
@@ -125,39 +125,39 @@ describe("convexPartitions", () => {
...
   test("Figure 1 from [HM83]", () => {
     // https://link.springer.com/content/pdf/10.1007/3-540-12689-9_105.pdf
-    // point locations approximated by tracing in Inkscape, then y-coordinates
-    // are inverted because screen coordinates are upside down compared to math,
-    // and the library we use expects the polygon to be counterclockwise
+    // point locations approximated by tracing in Inkscape; also note, this
+    // polygon is actually clockwise because screen coordinates are upside down
+    // compared to math
     const p = [
-      [-316.39758, 117.40885],
```
jiriminarcik, line 132: Maybe we can leave this as well and test both orientations.

jiriminarcik (approved):
Wow, that was quick! :) Thanks @samestep and @x6ud!

packages/core/src/contrib/__tests__/Minkowski.test.ts
samestep: Hmm we could, but the other test is counterclockwise so we kind of already are testing both. Up to you though!

jiriminarcik: It is now, but both tests will be a bit more future proof I think

samestep: Fair enough! Done.

PR #904, opened 2022-03-03
refactor: Put examples into a package

samestep:
# Description

This PR creates a `@penrose/examples` package and moves the contents of the top-level `examples/` folder into it. The new `packages/examples/build.js` script generates a JSON file (which can be imported from TypeScript) for each Substance/Style/Domain/SVG file in our examples, and an `index.ts` file for each directory, so the entire directory structure is available as an exported `examples` object, which can then be used in any context, whether Node.js or otherwise.

# Checklist

# Open questions

- Several things now don't work unless one first runs `yarn build` from `packages/examples/`; is this OK?
- @maxkrieger Do you happen to know why Cloudflare Pages is failing to build on this PR / how to fix it?
- After/if we merge this PR, I'll need to update our [Registry](https://github.com/penrose/penrose/wiki/Registry) wiki page.


wodeni (approved):
LGTM! Shipping them as a part of the package is okay because the trios are pretty small in general. We just need to make sure that we don't ship super huge svgs from now on (?).

samestep:
@wodeni Makes sense 👍 just to clarify, by "the package" do you mean `@penrose/core`? Nothing from `@penrose/examples` should be appearing in the build artifact for `@penrose/core`, so if they are then I need to fix that; do you know how I can check?

wodeni:
I was just referring to the size of `@penrose/examples` itself. I checked the bundle itself, nothing in there, so I think we are okay.

Issue #885, opened 2022-02-08
Use complex arithmetic for all calculations done by the layout engine


GeometryCollective:
**Is your feature request related to a problem? Please describe.**

In Penrose, as in other systems for constraint-based drawing (Cinderella, Geogebra, etc.), one runs into trouble when the calculated quantity is not well-defined/does not exist.

A simple example is taking square roots.  For instance, consider two values `a` and `b` that should be identical, but due to numerical rounding we get

`b - a = -ε < 0`

for some small value ε, say, 0.000001.  Hence, if we take the square root of the difference, we get

`sqrt(b-a) = sqrt(-0.000001) = NaN` (not a number),

i.e., we get catastrophic failure, even though the argument to `sqrt` is only _very_ slightly negative, and the result should in principle be equal to zero.

**Describe the solution you'd like**

A far more robust alternative is to **represent every real floating-point number by a complex floating-point number** throughout the entire layout engine.  The reason this strategy is more robust is that operations on complex numbers are well-defined in many cases where operations on the corresponding real numbers are not.  Principally, _every complex polynomial equation has solutions_ (in fact, this statement is called the [fundamental theorem of algebra](https://en.wikipedia.org/wiki/Fundamental_theorem_of_algebra)!).  Since basic shapes (circles, lines, etc.) are often expressed as polynomials, and basic operations (like intersections) involve polynomial equations, this fact is very powerful when it comes to ensuring that drawings are always well-defined.

In fact, this "complexification" of real numbers is the strategy adopted by many algebraic drawing systems (such as those mentioned above), as well as many computer algebra systems (like _Mathematica_).  In the example above, for instance, the result would now be

`sqrt(0.000001) i`,

i.e., the square root of a small positive number, times the imaginary unit `i = √-1`.  To get a final, real-valued answer we can, e.g., just take the real part of the result (which in this case is equal to zero).  We never have to worry about NaNs.

**Describe alternatives you've considered**

A common alternative is to clamp arguments to a valid range.  For instance, in the case of square root one can always compute

`sqrt(max(0,x))`

to ensure a NaN does not occur.  Likewise, when dividing by a number, one can compute

`1/max(eps,x)`

for some small value `eps`, to ensure that `1/x` does not yield INF (infinity).  This approach is unsavory since:

- it requires additional computation to "fix" the arguments
- in our case, this additional computation blows up the size of the expression tree, and hence the derivatives, and
[...]

PR #906, opened 2022-03-03
feat: Add a polynomial roots node

samestep:
# Description

This PR adds a `polyRoots` function to `engine/AutodiffFunctions`, using [ml-matrix](https://github.com/penrose/penrose/pull/954) to compute the roots (the eigenvalues of the [companion matrix](https://en.wikipedia.org/wiki/Companion_matrix)) and [implicit differentiation](https://www.skewray.com/articles/how-do-the-roots-of-a-polynomial-depend-on-the-coefficients) to compute the gradient (we ignore the possibility of the roots being "structurally multiple"). Because we don't currently [support complex numbers](https://github.com/penrose/penrose/issues/885), `polyRoots` returns `NaN` for a root if its imaginary component is not _exactly_ zero, to avoid computing an incorrect derivative (because its implicit differentiation scheme only works if the returned value is genuinely a root of the polynomial). In this case, the gradient of that `NaN` root with respect to the polynomial coefficients is considered to be zero, because its value is essentially discrete.

Previously, every node in our symbolic-differentiation computation graphs represented either a number or a boolean. To deal with the fact that some nodes were booleans we allowed the sensitivity of an edge to be `undefined` if either of the nodes it connected was a boolean. Regardless of whether it represented a number or a boolean, every node was of type `VarAD`.

Because all the roots of a polynomial must be computed together, this PR needed to add a third type to our computation graph: vectors. To provide better typechecking, `VarAD` now specifically refers to numbers: types `ad.Bool` and `ad.Vec` have been introduced, as well as `ad.Expr` which includes all three, serving some of the purposes `VarAD` did before (mostly just in `engine/Autodiff` itself). The previous representation of sensitivity as `VarAD | undefined` no longer works, because now we need to account for vector nodes, which each require a vector of gradient nodes.

To deal with this, sensitivity is now of type `VarAD[][]`. The cell at row `i` and column `j` essentially represents the partial derivative of element `i` of the parent vector with respect to element `j` of the child vector. We currently don't have any nodes that map from vectors to vectors, so right now every sensitivity is either empty (if `ad.Bool`s are involved), 1x1 (if both nodes are `VarAD`s), or a row or column vector (if one node is a `VarAD` and the other is an `ad.Vec`).

# Checklist


samestep:
I need #954 to be merged before I can finish this.

packages/core/src/types/ad.ts
```diff
@@ -39,9 +39,22 @@ import { LbfgsParams } from "./state";
 
 //#region Types for implicit autodiff graph
 
-export type VarAD = Const | Input | Unary | Binary | Ternary | Nary | Debug;
+export type Expr = Bool | VarAD | Vec;
 
-export type Const = ConstNode;
+export type Bool = Comp | Logic;
+
+// this type name is a relic, doesn't follow our current convention
```
wodeni, line 46: Perhaps rename to `ad.Var`?

packages/core/src/engine/Autodiff.test.ts
```diff
@@ -270,3 +296,89 @@ const gradGraph0 = (): ad.Graph => {
...
+    const [z] = polyRoots([input({ key: 0, val: 0 })]);
+    const g = primaryGraph(z);
+    const f = genCode(g);
+    const x = 42;
+    expect(f([x])).toEqual({ gradient: [-1], primary: -x, secondary: [] });
+  });
+
+  type F = (v: VarAD, w: VarAD) => VarAD;
+
+  const testQuadratic = (f1: F, f2: F) => {
```
wodeni, line 313: Super cool test case. Took me a while to figure out this is testing if the polyroot solution is the same as the closed form formula for both roots of quadratic equations. Would be great to add a short comment that explains what this function is doing.

My understanding is that the polyroots comp graph (`implicit`) finds the roots of `ax^2 + bx + c = 0`, where `a = PI^2` and `b = - PI - E`?

wodeni: Also, is that a monic polynomial? Or the monic part is just what the system does internally?

packages/core/src/engine/Autodiff.ts
```diff
@@ -920,14 +1009,31 @@ const compileNode = (
...
+  const n = coeffs.length;
+  // https://en.wikipedia.org/wiki/Companion_matrix
+  const companion = Matrix.zeros(n, n);
+  for (let i = 0; i + 1 < n; i++) {
+    companion.set(i + 1, i, 1);
+    companion.set(i, n - 1, -coeffs[i]);
+  }
+  companion.set(n - 1, n - 1, -coeffs[n - 1]);
+
+  const decomp = new EigenvalueDecomposition(companion);
```
wodeni, line 1022: So basically the trick here is to construct a square matrix ("companion matrix") for a monic polynomial, whose eigenvalues are roots of this polynomial... right?

A comment would be super helpful for folks who are a little rusty on linear alg (like me :P).

packages/core/src/engine/Autodiff.ts
```diff
@@ -920,14 +1009,31 @@ const compileNode = (
...
+  return zip2(
+    decomp.realEigenvalues,
+    decomp.imaginaryEigenvalues
+  ).map(([r, i]) => (i === 0 ? r : NaN));
```
wodeni, line 1026: And this is discarding imaginary roots and treating them as NaNs?

packages/core/src/engine/Autodiff.ts
```diff
@@ -939,16 +1045,16 @@ export const genCode = ({
...
+    // valid JavaScript syntax for array literals with holes, so everything
+    // works out perfectly
     `gradient: [${gradient.join(", ")}]`,
     `primary: ${primary}`,
     `secondary: [${secondary.join(", ")}]`,
   ];
   stmts.push(`return { ${fields.join(", ")} };`);
-  const f = new Function("inputs", stmts.join("\n"));
-  return (inputs) => f(inputs);
+  const f = new Function("polyRoots", "inputs", stmts.join("\n"));
```
wodeni, line 1056: Interesting that this is how it works: passing function pointers to `Function`!

wodeni (approved):
Reviewing this PR was pretty cool! The changes make sense to me, modulo some math that I wanted to understand (see below). The bit about sensitivity sounds good to me conceptually. 

One thing I'm not sure about at the high-level: are the coefficients in `Polyroot` nodes (i.e. `coeffs`) assumed to be less than 1? I saw a bunch of references to "monic` polynomials. The test cases seem to have greater-than-1 coefficients. Just trying to understand if that's a requirement in the engine.

samestep:
> One thing I'm not sure about at the high-level: are the coefficients in `Polyroot` nodes (i.e. `coeffs`) assumed to be less than 1? I saw a bunch of references to "monic` polynomials. The test cases seem to have greater-than-1 coefficients. Just trying to understand if that's a requirement in the engine.

Good question! "Monic" just means that the leading coefficient is 1; it says nothing about the other coefficients. So for instance, this code:
```typescript
polyRoots([1, 2, 3, 4])
```
corresponds to the polynomial 1 + 2x + 3x² + 4x³ + x⁴; that is, the coefficient on the x⁴ term is implicit.

packages/core/src/types/ad.ts
samestep: I'm fine with doing that in a followup PR that just does the rename, but I want to keep this diff small and focused.

samestep: Although if we do rename, I'd want to call it `ad.Num` instead of `ad.Var`.

packages/core/src/engine/Autodiff.test.ts
samestep: > Took me a while to figure out this is testing if the polyroot solution is the same as the closed form formula for both roots of quadratic equations. Would be great to add a short comment that explains what this function is doing.

Good point! I'll add a comment about that.

> My understanding is that the polyroots comp graph (`implicit`) finds the roots of `ax^2 + bx + c = 0`, where `a = PI^2` and `b = - PI - E`?
>
> Also, is that a monic polynomial? Or the monic part is just what the system does internally?

Not quite; the `polyRoots` comp graph is finding the roots of c + bx + x², where c = πe and b = -(π + e).

packages/core/src/engine/Autodiff.ts
samestep: Exactly (and these semantics are documented on the `polyRoots` function in `engine/AutodiffFunctions`). To be precise, these aren't necessarily imaginary roots, since that implies that the real component is zero. We discard any non-real root.

packages/core/src/engine/Autodiff.ts
samestep: Haha yes 😛 this seemed better than writing the body of the `polyRoots` function into our generated code every single time, because this way we get more typechecking.

packages/core/src/types/ad.ts
wodeni: Yeah that sounds great!

packages/core/src/engine/Autodiff.test.ts
wodeni: aha interesting! I was reading it backwards apparently!

packages/core/src/engine/Autodiff.ts
wodeni: Yep, would be good to note that explicitly in case it matters in the future

packages/core/src/engine/Autodiff.ts
wodeni: Yeah def better!

wodeni:
> > One thing I'm not sure about at the high-level: are the coefficients in `Polyroot` nodes (i.e. `coeffs`) assumed to be less than 1? I saw a bunch of references to "monic` polynomials. The test cases seem to have greater-than-1 coefficients. Just trying to understand if that's a requirement in the engine.
> 
> Good question! "Monic" just means that the leading coefficient is 1; it says nothing about the other coefficients. So for instance, this code:
> 
> ```ts
> polyRoots([1, 2, 3, 4])
> ```
> 
> corresponds to the polynomial 1 + 2x + 3x² + 4x³ + x⁴; that is, the coefficient on the x⁴ term is implicit.

Aha that explains it! Somehow I kept thinking other coeffs have to be less than 1. Makes perfect sense now. I guess the usage will be quite limited so that assumption is well understood? If not, simply throwing an error might be enough. Otherwise, we could introduce a new kind of compilation error just for handling errors coming from autodiff.

packages/core/src/engine/Autodiff.test.ts
wodeni: I was lazy enough to think the `4ac` term in the closed form solution explained the ordering. Well multiplication commutes!

samestep:
> I guess the usage will be quite limited so that assumption is well understood? If not, simply throwing an error might be enough. Otherwise, we could introduce a new kind of compilation error just for handling errors coming from autodiff.

I don't quite understand what you mean by this; could you clarify? Because the leading coefficient is implicit, there is no way to pass an "incorrect" `coeffs` array to `polyRoots`.

packages/core/src/engine/Autodiff.ts
samestep: Done :+1:

samestep:
Perhaps a better example would be this:
```typescript
polyRoots([2, 3, 4]) // 2 + 3x + 4x² + x³
```

packages/core/src/types/ad.ts
```diff
@@ -39,9 +39,22 @@ import { LbfgsParams } from "./state";
 
 //#region Types for implicit autodiff graph
 
-export type VarAD = Const | Input | Unary | Binary | Ternary | Nary | Debug;
+export type Expr = Bool | VarAD | Vec;
```
jiriminarcik, line 42: Would it make sense to drop the `AD` in `VarAD` or use `BoolAD`, `VecAD` etc.  for consistency?

packages/core/src/engine/Autodiff.ts
```diff
@@ -258,41 +265,94 @@ const children = (x: VarAD): Child[] => {
...
+
+      const sensitivities: VarAD[][] = x.coeffs.map((_, index) => {
+        const t: VarAD = { tag: "Index", index, vec: x }; // a root
+
+        let power: VarAD = 1;
+        const powers = [];
+        for (let i = 0; i < n; i++) {
+          powers.push(power);
+          power = mul(power, t);
+        }
```
jiriminarcik, line 330:
```suggestion
        const powers = [power];
        for (let i = 1; i < n; i++) {
          power = mul(power, t);
          powers.push(power);
        }
```

jiriminarcik (approved):
Looks great! I don't fully understand some of the code related to symbolic differentiation (still need to go through some of your previous PRs), so I checked mainly the math - looks good to me. :)

packages/core/src/types/ad.ts
samestep: See @wodeni's earlier comment: https://github.com/penrose/penrose/pull/906#discussion_r869626946 The plan is to merge this PR first and then do a refactor from `VarAD` to `ad.Num` later.

packages/core/src/engine/Autodiff.ts
```diff
@@ -258,41 +265,94 @@ const children = (x: VarAD): Child[] => {
...
+
+        const minusDerivative = neg(
+          addN(zip2(derivCoeffs, powers).map(([c, p]) => mul(c, p)))
```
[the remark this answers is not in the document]
samestep: Ah good point, we actually already do have a dot product!

https://github.com/penrose/penrose/blob/32623914ee70b3aa267189a6bcee42a18c044574/packages/core/src/engine/Autodiff.ts#L645-L656

But the current implementation uses `reduce` and `add` rather than `addN` so it generates a larger graph.

packages/core/src/types/ad.ts
jiriminarcik: sorry, didn't catch that

packages/core/src/engine/Autodiff.ts
```diff
@@ -258,41 +265,94 @@ const children = (x: VarAD): Child[] => {
...
+        );
+
+        // if the root is `NaN` then it doesn't contribute to the gradient
+        const real = eq(t, t);
+        return powers.map((p) => ifCond(real, div(p, minusDerivative), 0));
```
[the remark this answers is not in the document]
samestep: Yeah this happens in the case of a multiple root, as mentioned in the blog post I linked. You can get around this by taking the second derivative of the polynomial and using that, but that results in a larger computation graph. I don't want to pursue that until we resolve #642, because otherwise we'd be evaluating all those further derivatives unnecessarily even for non-multiple roots.

packages/core/src/engine/Autodiff.ts
samestep: TypeScript doesn't like your suggestion. Could you clarify the reasoning behind it?

jiriminarcik: Oh, ok..  I just didn't like that the last `power = mul(power, t)` was happening for no reason. :D

samestep: Ohh good point! Fixed :) the only thing missing from your suggestion was a type annotation on the `powers` variable.

Issue #888, opened 2022-02-14
Style compiler re-write


wodeni:
# Motivation

We plan to re-write the Style compiler such that it produces a more performant runtime date structure. The current representation of the `Translation`: a giant dictionary. Lookups and updates are way too costly and error prone with it, especially when evaluating at runtime. We agreed that an explicit computational graph should be the compilation target for Style because:
* The current dictionary structure does not allow efficient lookups and incremental updates. Any update to values in Style requires the entire dictionary to be traversed.
* There are known evaluation algorithms for computational graphs with an explicit graph structure

Since we are rewriting the compiler, other issues and feature requests might come up. This is the umbrella issue to organize  them, and keep track of our progress.

# Tracking issues

* Selector matching for special predicates: #744 
* Find varying values automatically: #890
* Shape property sets: #924 
* Function modules: #894 
* Refactor the evaluator so it doesn't construct a new comp graph every single time: #928
* Anonymous shape: #720

PR #907, opened 2022-03-03
refactor: Replace all autodiff with symbolic diff

samestep:
# Description

This PR rewrites our "autodiff" engine, which previously had the following issues:

- The semantics for every possible operation needed to be specified twice: once in its `VarAD` constructor function in `engine/AutodiffFunctions` (which would compute an "initial" `val` at computation graph construction time), and again in the `traverseGraph` function in `engine/Autodiff` (generated JavaScript code which would be run in every step of the optimizer).
  - This made it difficult and error-prone to add new primitives, or to modify the semantics of existing primitives.
  - Because every `VarAD` had an easily-accessible `val` field, a lot of code abused this to thoughtlessly convert from `VarAD`s to `number`s, which is almost always incorrect because in general the `val` of a `VarAD` at its creation does not reflect all the values it will take on as the optimizer runs.
- The code for computing symbolic derivatives and compiling a computation graph into JavaScript code was messy and made heavy use of both recursion and mutation, imposing the stack's depth limit on our own computation graphs, and requiring careful usage of `clear...` functions to ensure that the graph was in the right state before doing anything with it.
- Any computation graph for computing derivatives reuses most or all of the original graph for computing the actual function, but previously we would compile separate functions for computing both the function and the gradient, instead of computing both at the same time to save work.
- The constructor functions in `engine/AutodiffFunctions` were each incredibly verbose and repetitive, mutating both the children and the new parent, and unnecessarily case-splitting on whether the node is only involved in computing gradients.
  - Because each child stored an array of its parents in the order they were created, and because the code for computing derivatives used the order of this array, and because floating-point addition is not associative, the exact result of our engine depended not just on the structure of a computation graph, but also on when different parts of it were constructed.
- The `VarAD` type itself was simultaneously overcomplicated and underpowered: it was a single `interface` with many fields, only a subset of which mattered for any given node, and did not reflect the structure of, for instance, our set of primitives well enough to provide good type safety.
  - Because every `VarAD` needed to be an object with all of these fields, every literal number in a computation graph needed to be wrapped in a `constOf` call, adding unnecessary verbosity throughout our codebase.

The rewrite resolves all these issues, with the exception of the sub-bullet about the order of partial derivatives; see my comment in `engine/AutodiffFunctions`. I will immediately open a PR to fix that once this one is merged, but I am not doing that in this PR because it is important that this large rewrite not introduce any changes to the actual `diagrams/` produced by the system.

The representations for the structures used by our new differentiation engine are heavily documented in `types/ad`. A few other lower-level notes:

- Because `number` is now a subtype of `VarAD`, many functions such as `constOf`/`constOfIf`/`varOf`/`valueNumberToAutodiff` are completely unnecessary and have been removed. In a similar vein, every `number | VarAD` type is now just `VarAD`.
- Because `VarAD` is now a union type with no single interface, the `IVarAD` type has been removed and every usage of it has been replaced with `VarAD`.
- The `graphlib` library proved to be quite valuable for writing this PR, but it is not as type-safe as we'd like. To address this, I wrote a `Multidigraph` class in `utils/Graph` which acts as a thin wrapper around `graphlib`'s `Graph` class and provides a much nicer interface for the specific kind of graph we're using here.

I also added a `.github/bench.sh` script which uses [`gh`](https://cli.github.com/) to download the benchmarking results for both `main` and the current branch, and then uses [`icdiff`](https://www.jefftk.com/icdiff) to compare them. The numbers are a bit noisy, but it looks like this PR does not hurt performance.

- in a few places this PR replaces "magic numbers" with constants from `Math` (e.g., `E`, `LOG10E`, `LOG2E`, `PI`); in each of these cases, this is purely a cosmetic/readability change, i.e., if you compare the two using `===` in a `node` terminal, you'll get `true`

- minor note: previously there was code for `VarAD`s whose `op` field was `"div"`, which would add `EPS_DENOM` to the denominator:

  https://github.com/penrose/penrose/blob/ffd63d022a825a96795b5ef5e982200801b40159/packages/core/src/engine/Autodiff.ts#L869-L870

  however, the actual `div` function set the `op` field to `"/"`, not `"div"`, so this was effectively dead code; the new implementation instead makes use of the `else` case in that section of code, which just produces a simple division without changing the denominator

- previously there were some minor discrepancies between the logic to populate the `val` fields of `VarAD`s and the generated JavaScript code from computation graphs:

  - `div` would throw an `Error` if the initial value of its denominator was close enough to zero, but the compiled code would just perform division without checking the denominator first
  - `gt`, `lt`, `eq`, `and`, `or`, and `ifCond` would use numbers just like all the other nodes when constructing the graph, but the compiled code for these would (and still does) actually use booleans, not numbers

- I modified `compDict` from `contrib/Functions`:

  - removed `getVar` because it was previously unused, and accessed the `val` field from `VarAD`, and I couldn't tell what exactly its purpose was
  - modified `selectColor` to not use the `val` field, but that means it needs to choose a color type at AD graph construction time, so I chose RGBA because that's the only color type it's used with in the examples
  - restricted `ticksOnLine` to require that `numTicks` be a constant, since it doesn't work otherwise anyway

- there was also an unused `furthestFrom` function in `contrib/Functions` which was not exported and marked with "Do not use this!" so I deleted it because it called `numOf`

- I removed the "invalid range" check from `inRange` in `contrib/Utils`; it already had a TODO saying "do this range check better" because, since it uses the `val` field, it wouldn't actually work if the `l` and `r` endpoints of the range change during optimization

- now instead of just calling `numOf`, the `convexPartitions` function in `contrib/Minkowski` compiles a computation graph composed of the polynomial coordinates it's passed, calls the compiled function of that graph with the initial values embedded in any relevant input nodes, and uses that to compute a convex partition

- pending #927, this PR does not modify `packages/docs-site/docs/ref/style/functions.mdx` (e.g. to replace `IVarAD` with `VarAD`), because it would clutter this already-too-long diff, and any effort spent doing that now will be made moot once automation is put in place

- while modifying `engine/EngineUtils`, I also refactored some `if`/`else if`/`else` chains into `switch` statements in that module

- while debugging the usage of `engine/Autodiff` in `engine/Optimizer`, I found it useful to know the exact lengths of mismatched vectors when `addv`/`subv`/`dot` from `utils/Util` would throw, so I modified their error messages to provide that information

- previously `engine/Autodiff` had a few unused "Useful (grad) constants" with the comment "TODO -- But it seems to be bad to use them..."; I removed them because the removal of `constOf` makes them unnecessary

# Checklist

# Open questions

- As mentioned above, my immediate followup to this PR will remove the global `count` variable from `engine/AutodiffFunctions`, slightly changing some of our optimized `diagrams/` but providing a greater level of determinism for the future. When we do this, though, we should probably make an explicit choice for the ordering of partial derivatives (as opposed to, e.g., just using whatever order `graphlib`'s `outEdges` method gives us) so that we are more resistant to future refactors. Perhaps we should use the order given by the breadth-first search at the beginning of the `makeGraph` function, since that is intuitive, easy to define, and only depends on the structure of the graph?
- Nearly every usage of the compiled `f` and `gradf` functions needs both, so our new strategy to compile the function and its gradient together and compute both at the same time almost always saves work, but the one exception is the `evalEnergy` function in `src/index.ts`. There may be an easy way to eliminate this unnecessary computation of the gradient, since the optimizer already computes the energy at every step.
- There are several places where `EPS_DENOM = 10e-6` is added to denominators to try to avoid divide-by-zero issues. This is obviously inelegant, and also doesn't resolve the issue in general because it can even cause divide-by-zero where it wouldn't have happened otherwise, if the original denominator was a very small negative number. Is there a better way to approach this?
- Some places call autodiff constructor functions on constants, such as the implementation of `average` in `contrib/Functions`. These should be replaced by calls to the more basic JavaScript functions instead.


wodeni (approved):
LGTM! The PR is super well documented and clean, so I don't really have any substantive suggestions. My high-level takeaway is that this approach requires re-codegen whenever the structure of the comp graph changes, which doesn't happen in our current design.

One thing to discuss is value updates: the system does update values in the comp graph that are not considered "input" nodes now. For example, the frontend inserts dimensions of labels post-codegen. My current understanding is we can include all pending variables as input nodes in the graph. Then we are left with a trickier case of other types of frontend updates, e.g. dragging shapes. If the frontend interaction updates DOFs only, I believe this system can handle this scenario. f it updates constants, however, it might require us to add all shape properties as input nodes. Anyway, this is obviously a much broader discussion to have in #888. The PR is great!

samestep:
> One thing to discuss is value updates: the system does update values in the comp graph that are not considered "input" nodes now. For example, the frontend inserts dimensions of labels post-codegen. My current understanding is we can include all pending variables as input nodes in the graph. Then we are left with a trickier case of other types of frontend updates, e.g. dragging shapes. If the frontend interaction updates DOFs only, I believe this system can handle this scenario. f it updates constants, however, it might require us to add all shape properties as input nodes. Anyway, this is obviously a much broader discussion to have in https://github.com/penrose/penrose/issues/888.

@wodeni Thanks for bringing this up, definitely something we should discuss more. Currently though I don't quite follow all of this since you have a broader knowledge of the system architecture than I do 😅

PR #908, opened 2022-03-04
chore: Prefix synthesizer-ui name with @penrose

samestep:
I'm guessing this is just an oversight?


wodeni (approved):
Thanks for catching it! The package is never published so it didn't matter, but eventually we should publish it tho.

PR #916, opened 2022-03-18
fix: Allow Docusaurus and IDE co-existence

cmumatt:
# Description

This PR allows Docusaurus and IDE co-existence at CloudFlare by injecting the IDE build output into the Docusaurus build, which eliminates the need for two deployments and two URLs.

# Checklist

# Open questions

- I temporarily commented out the live demo since it is just showing a circle; I figured that's still a work in progress.


packages/docs-site/package.json
```diff
@@ -6,7 +6,7 @@
     "docusaurus": "docusaurus",
     "start": "NODE_OPTIONS='--max-old-space-size=8192' docusaurus start",
     "watch": "NODE_OPTIONS='--max-old-space-size=8192' docusaurus start",
-    "build": "NODE_OPTIONS='--max-old-space-size=8192' docusaurus build",
+    "build": "rm -f ./static/try/*;cp ../panels/public/* ./static/try/;NODE_OPTIONS='--max-old-space-size=8192' docusaurus build",
```
wodeni, line 9: Would be better to have a separate script for this. You could use a pre-hook (`prebuild`) for this purpose: https://docs.npmjs.com/cli/v8/using-npm/scripts#pre--post-scripts

wodeni: Also, `rimraf` is supposedly faster and platform independent for `rm -rf`

wodeni (commented):
Thanks for doing this one! Still a little unclear to me how the deployment process look like. Are we using Cloudflare now? Looking at it now, it's still deploying `panels`. In my head I was thinking we were going to:

* Add a top-level script in `package.json`: `yarn build:sites` builds everything, right now just panels and docs-site
* This script should build `@penrose/docs-site` and `@penrose/panels`, and then collect build artifacts into a singular folder for deployment.

Lmk if I was just thinking about the wrong thing, and how deployment is done at the moment.

packages/docs-site/package.json
cmumatt: Awesome!  Thanks for reviewing and for that suggestion; I also like the idea of a pre-hook better and will create a to-do issue to implement that.  At the same time, I will check if there is a similar analog for cp if we aren't assuming a *ix-like script environment.

Cloudflare is temporarily still deploying penrose/main to the old penrose-panels project -- but all traffic to that project is redirected to the penrose-site project (penrose.cs.cmu.edu), which currently hosts this PR's build output.  So, CD is currently broken on purpose.  Once this PR is merged, I plan to re-point penrose/main's CD to penrose-site, which involves re-creating both projects due to repo associations being 1:1 and immutable at CF, as you pointed out previously, which is kind of a bummer -- but our current reality.

Issue #922, opened 2022-03-23
Add sugared syntax for string concatenation in Style


keenancrane:
We currently support string concatenation via a function `concat`, e.g.,
```
forall Point p
where p has label {
   shape t = Text {
      string: concat( p.label, "!" )
   }
}
```
will display a more-excited version of the label given for `p`.

It would be nice to have some more lightweight sugared syntax, like
```
string: p.label + "!"
```

(I don't have a strong preference for `+` versus anything else, but it seems to be pretty common across JavaScript, Python, C++, etc.)

PR #923, opened 2022-03-23
chore: evaluate binary addition on strings as concatenation

wodeni:
# Description

Fix: #922 

In `evalBinOp`, pattern match on string-type arguments to `+`, and perform plain string concatenation of them.

# Checklist

# Open questions

This is only lightly tested, and we don't really have a suite for `Evaluator`. We should have one fairly soon though!


samestep (approved):
Seems pretty straightforward! Would be nice if this PR added some tests though (that would fail previously).

PR #930, opened 2022-04-19
 refactor: produce semantically checked AST from the Substance checker

wodeni:
# Description

Related issue/PR: #638, #446  

(@samestep commented on the `disambiguateFunctions` function in #638, which reminded me of this issue)

There are two ambiguous parts of the Substance (and also Style selector) spec:

1. The right hand side of `Bind` could either be a constructor or function: `C := SubSet(A, B)`. It's syntactically impossible to see if `SubSet` is a function or constructor.
2. Nested predicate vs. anonymous function/constructor: In `Not(Intersecting(A, B))`, `Interesecting(A, B)`'s type cannot be determined by the parser.

To resolve 1 and 2, the current system parse both cases into a generic `Func` type and implements `disambiguateFunctions`, which mutates an AST, to resolve these issues. However, the Substance semantic checker is already doing similar work in functions like `checkFunc`. In this PR, I update the semantic checker so it returns a new, disambiguated AST along with the Substance environment.

* Parametrize `CheckerResult` by the type of the AST node returned
* Thread the return values in the checker and eventually return the new AST in `SubstanceEnv.ast`

* In `Substance.test.ts`, I added a test case for disambiguating cases 1 and 2 above.

# Checklist

# Open questions

* There's no test for the same thing in Style selectors. Add them?

### DIFF
--- a/packages/core/src/compiler/Style.test.ts
+++ b/packages/core/src/compiler/Style.test.ts
@@ -2,11 +2,7 @@
 
 import { examples } from "@penrose/examples";
 import * as S from "compiler/Style";
-import {
-  compileSubstance,
-  disambiguateFunctions,
-  parseSubstance,
-} from "compiler/Substance";
+import { compileSubstance, parseSubstance } from "compiler/Substance";
 import _ from "lodash";
 import { A, C } from "types/ast";
 import { Either } from "types/common";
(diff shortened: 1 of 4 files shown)


samestep (approved):
LGTM! I mostly just skimmed; it looks most of it is just threading stuff around right? And then the important parts are the deletion of `disambiguateFunctions` and the modification of the `checkPredArg` function?

PR #931, opened 2022-04-19
ci: add semantic PR to GH Action

wodeni:
# Description

[Semantic PR](https://github.com/zeke/semantic-pull-requests)'s service is unstable. This PR adds a GitHub Action that does the same thing, which should be much more reliable.

### DIFF
--- a/.github/workflows/lint-pr.yml
+++ b/.github/workflows/lint-pr.yml
@@ -1,0 +1,17 @@
+name: "Lint PR"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+jobs:
+  main:
+    name: Validate PR title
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-pr.yml
```diff
@@ -0,0 +1,22 @@
+name: "Lint PR"
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - synchronize
```
samestep, line 8: I think it's incorrect to have this triggered on both `pull_request` and `pull_request_target`; judging from [the action's README](https://github.com/amannn/action-semantic-pull-request/tree/v4#event-triggers), it seems like we should be using only `pull_request_target`?
```suggestion
```

PR #939, opened 2022-04-25
ci: Use automator for benchmarking

samestep:
# Description

This replaces #921. After some discussion, @wodeni and I decided that [jest-bench](https://github.com/pckhoi/jest-bench) does not fit our needs, because what we want is to get a breakdown of how much time is taken by each stage of the diagram generation process, for each diagram in our registry. That functionality is already present in `@penrose/automator`, so this PR replaces our existing end-to-end tests and benchmark suites with a CI job that just uses `@penrose/automator` directly.

This PR also makes a few minor improvements to the code in `@penrose/automator`, such as `import` instead of `require`, and now pretty-prints the JSON and SVG files it generates. One important thing to note is that it removes the `-instance-...` part of the generated folder name, because that information is already present in the generated `meta.json` file.

# Checklist

### DIFF
--- a/packages/automator/index.tsx
+++ b/packages/automator/index.tsx
@@ -338,7 +338,7 @@ const batchProcess = async (
   if (folders) {
     fs.writeFileSync(
       join(out, "aggregateData.json"),
-      JSON.stringify(finalMetadata)
+      JSON.stringify(finalMetadata, null, 2)
     );
     console.log(`The Aggregate metadata has been saved to ${out}.`);
   }


joshsunshine:
What's an example of output of benchmarking?

wodeni:
> What's an example of output of benchmarking?

It's uploaded as artifacts: https://github.com/penrose/penrose/actions/runs/2242107757.

samestep:
> What's an example of output of benchmarking?

@joshsunshine For a run of the Build workflow in CI: https://github.com/penrose/penrose/actions/runs/2242107757

If you scroll down to the "Artifacts" section and download the "diagrams" artifact, you should get a file called `diagrams.zip`. If you unzip this then, for instance, `tree-venn/meta.json` looks like this:
```json
{
  "substanceName": "Tree",
  "styleName": "Venn",
  "domainName": "Set Theory",
  "id": "instance-uiq1d3l2jjgb73",
  "renderedOn": 1651182545285,
  "timeTaken": {
    "overall": 3541.570209,
    "compilation": 150.48668,
    "labelling": 837.717461,
    "optimization": 2537.525157,
    "rendering": 15.712809
  },
  "selectorMatches": [],
  "optProblem": {
    "constraintCount": 42,
    "objectiveCount": 7
  },
  "reference": {
    "substance": "tree",
    "style": "venn",
    "domain": "set-theory",
    "variation": ""
  }
}
```
And `aggregateData.json` contains all of these JSON files put together, keyed by the `id`.

PR #945, opened 2022-05-03
fix: Make autodiff deterministic in graph shape

samestep:
# Description

This PR fixes the global `count` hack leftover by #907, replacing it with an ordering on nodes according to breadth-first search and the ordering on edge labels already given by the `children` function.

# Checklist

### DIFF
--- a/packages/core/src/engine/Autodiff.ts
+++ b/packages/core/src/engine/Autodiff.ts
@@ -4,7 +4,7 @@ import * as _ from "lodash";
 import * as ad from "types/ad";
 import { GradGraphs, VarAD } from "types/ad";
 import { WeightInfo } from "types/state";
-import { Edge, Multidigraph } from "utils/Graph";
+import { Multidigraph } from "utils/Graph";
 import { safe } from "utils/Util";
 import {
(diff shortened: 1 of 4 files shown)


packages/core/src/engine/Autodiff.ts
```diff
@@ -277,6 +298,7 @@ const children = (x: VarAD): Child[] => {
 };
 
 const indexToID = (index: number): ad.Id => `_${index}`;
+const idToIndex = (id: ad.Id): number => parseInt(id.slice(1), 10);
```
wodeni, line 301: A little weird that you have to parse this. Can't the edge store the indices too?

wodeni (approved):
Looks great! Glad that the index hack is gone now. Had a small question about `idToIndex` below. I obviously don't have the context for this. If it's needed, please feel free to just merge this.

packages/core/src/engine/Autodiff.ts
wodeni: Yeah that makes sense! Just wanted to flag this. Given that I don't really see a better way.

PR #946, opened 2022-05-04
fix: Automator NPE in staged mode. Resolves #887

### LINKED ISSUE #887: fix: null de-reference in Automator
**Describe the bug**
Variable listOfCanvasData is defined and used used but never initialized in Automator's `singleProcess()` method.  Consequently, this method will always fail with a NPE when staged = true.  

https://github.com/penrose/penrose/blob/4b749ae2359b3a3e257103264bcb9b12039772bf/packages/automator/index.tsx#L118-L128

Further null de-references follow in the code.  It appears this regression was introduced by PR# 825, about a month ago when the line to initialize this data structure was removed.

**To Reproduce**
This was identified via static analysis.  I need to come back to this issue and type up a repro.

cmumatt:
# Description

Resolves #887

This PR corrects an NPE in Automator's staged mode.  Starting with PR #825, `listOfCanvasData` was never initialized and executing in staged mode would always fail.  This bug was discovered via static analysis.

Initialized the variable `listOfCanvasData` prior to its use.

```
cd ./packages/automator
yarn start draw tree.sub venn.sty setTheory.dsl ./out/ --src-prefix=../examples/src/set-theory-domain --folders --staged
```

# Checklist

# Open questions

No open questions.

### DIFF
--- a/packages/automator/index.tsx
+++ b/packages/automator/index.tsx
@@ -119,7 +119,8 @@ const singleProcess = async (
   const reactRenderStart = process.hrtime();
 
   // make a list of canvas data if staged (prepare to generate multiple SVGs)
-  let listOfCanvasData, canvas;
+  const listOfCanvasData: string[] = [];
+  let canvas;
   const resolvePath = async (filePath: string) => {
     const parentDir = parse(join(prefix, sty)).dir;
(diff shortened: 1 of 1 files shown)


wodeni (approved):
LGTM. Thanks for fixing it!

Issue #535, opened 2021-03-30
Labels disappear on drag or inspector frame selection


maxkrieger:
Possibly due to ID conflicts throughout the DOM among MathJax refs.

joshpoll:
This also seems to happen when downloading the SVG.

wodeni:
Update: bizarre case of disappearing labels when multiple diagrams are shown on the same page:
![image](https://user-images.githubusercontent.com/11740102/141319951-77bc5831-451b-414c-91e1-65be2f2ffbce.png)
Deleting `memoize` in `collectLabels` fixed it, although I wasn't sure about the effect on performance. Also, even with `memoize` gone, the other phenomena in this issue still persisted.

wodeni:
Another example likely for the same reasons: in `synthesizer-ui`, exporting the diagrams cause the labels to disappear on the page.

PR #953, opened 2022-05-09
feat: display errors in the `Simple` component

wodeni:
# Description

Related issue/PR: #535, #671

`Simple` is a raw canvas for Penrose-generated diagrams. In the past iterations, the component didn't handle errors and rendered a blank canvas. In this PR, whenever `Simple` encounters an error, it renders the pretty-printed error message directly on canvas.

* Remove `memoize` from `CollectLabels`
* Keep `error` in the state of `Simple`. This is mostly to let React handle re-rendering, which is a different case from `penroseState` (?)
* Used error rendering code from `browser-ui`
* Added errors to the storybook example
* Change Edgeworth to use `Simple` (see below)

In `synthesizer-ui`, I changed `Gridbox` so it's just a wrapper around `Simple`, a few minor notes:
* `Gridbox` has an overlay for mutation info and used to conditionally render the Penrose diagram whenever the user toggles the overlay. When using `<Simple>`, this can be very slow because re-rendering will also run the Penrose optimizer. Instead, I always keep `<Simple>` in the parent `div`, and only stack the info `div` on top of it once toggled.
*  In terms of the behavior of Edgeworth, the system seems to render all diagrams at once after loading on a blank screen for a while, not sure if this is new.
* See below: "mutated program 3" show an error

image

# Checklist

# Open questions

### DIFF
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -117,7 +117,7 @@ import {
 import { checkTypeConstructor, isDeclaredSubtype } from "./Domain";
 
 const log = consola
-  .create({ level: LogLevel.Info })
+  .create({ level: LogLevel.Warn })
   .withScope("Style Compiler");
 const clone = rfdc({ proto: false, circles: false });
 
--- a/packages/core/src/utils/CollectLabels.ts
+++ b/packages/core/src/utils/CollectLabels.ts
@@ -1,4 +1,3 @@
-import memoize from "fast-memoize";
 import { browserAdaptor } from "mathjax-full/js/adaptors/browserAdaptor.js";
 import { RegisterHTMLHandler } from "mathjax-full/js/handlers/html.js";
 import { TeX } from "mathjax-full/js/input/tex.js";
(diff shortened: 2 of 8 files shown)


packages/core/src/utils/CollectLabels.ts
```diff
@@ -64,37 +63,32 @@ type Output = {
 
 /**
  * Call MathJax to render __non-empty__ labels.
- * NOTE: this function is memoized.
  */
-const tex2svg = memoize(
```
samestep, line 69: Could you clarify why you removed this `memoize`?

packages/core/src/compiler/Style.ts
```diff
@@ -117,7 +117,7 @@ import {
 import { checkTypeConstructor, isDeclaredSubtype } from "./Domain";
 
 const log = consola
-  .create({ level: LogLevel.Info })
+  .create({ level: LogLevel.Warn })
```
samestep, line 120: Is this change relevant to this PR, or just for general quality-of-life?

samestep (approved):
Awesome, thanks!

I tried running `synthesizer-ui` locally (to compare this branch with `main`) and had a couple questions:

1. When I ran it on `main`, I got an error in Mutated Program <span>#</span>5, but when I ran it on this branch, the error was instead in Mutated Program <span>#</span>4. Is the synthesizer nondeterministic?
2. For the diagrams that don't error, I can click anywhere on the box to see the mutation overlay, but for the diagrams that do error, this only works if i click above the bottom of the error message. Could you change it so that errored diagrams let you click anywhere on the box to see the mutation overlay?

packages/core/src/utils/CollectLabels.ts
wodeni: Yes, #535 noted this issue, but there's no PR that fixes it. This fix doesn't close the issue though, because when downloading the SVGs (which is currently broken in Edgeworth, to be addressed in another PR) the labels still disappear.

packages/core/src/compiler/Style.ts
wodeni: The latter, mostly annoyed by the console haha

wodeni:
>     1. When I ran it on `main`, I got an error in Mutated Program #5, but when I ran it on this branch, the error was instead in Mutated Program #4. Is the synthesizer nondeterministic?

Yes, I pass in a randomized seed to the synthesizer, which is the expected behavior of this GUI, as users expect different diagrams every time they hit `Generate diagrams`.

>     2. For the diagrams that don't error, I can click anywhere on the box to see the mutation overlay, but for the diagrams that do error, this only works if i click above the bottom of the error message. Could you change it so that errored diagrams let you click anywhere on the box to see the mutation overlay?

Correct. I finally learned that flexbox and `height: 100%` does the trick for me. The most recent commit includes that fix.

samestep:
Makes sense, thanks!

PR #956, opened 2022-05-11
refactor: Remove children field from ASTNode

### LINKED ISSUE #841: Make AST children a function instead of a field
Currently our `ASTNode` type includes `children` as a field.

https://github.com/penrose/penrose/blob/6a89b032628e6b36df267decfc94b38024b80cc6/packages/core/src/types/ast.ts#L12-L16

This is annoying because as @wodeni points out, there are some cases in which we want to remove the `children` from all the fields (one example of this could be when using the `printAST` function in `packages/core/src/parser/*.test.ts`).

It would be better to remove the `children` field and instead just have a function that computes the list of children of an AST node, since this is fairly easy to do.

samestep:
# Description

Resolves #841.

# Checklist

- [ ] I have reviewed any generated changes to the `diagrams/` folder

# Open questions


wodeni (approved):
Looks great! This is much cleaner and easier to maintain.

PR #955, opened 2022-05-10
docs: Update the README

samestep:
# Description

Resolves #569.

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

# Implementation strategy and design decisions

Include a high-level summary of the implementation strategy and list important design decisions made, if any.

# Examples with steps to reproduce them

# Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new ESLint warnings
- [ ] I have reviewed any generated changes to the `diagrams/` folder

# Open questions

Questions that require more discussion or to be addressed in future development:

PR #958, opened 2022-05-11
docs: Bring back the demo

samestep:
# Description

This PR reenables the homepage demo which was added in #882 and then removed in #916. I don't remember the exact reasoning for removing the demo before; @cmumatt could you comment on this? If the problem is simply that the text should be tweaked, I can do that in this PR, I just wasn't sure exactly what needed to change.

In any case, now that #904 has made a `@penrose/examples` package available to `@penrose/docs-site`, we can simply use the more sophisticated trio we show in the README as of #955. The animation is pretty fun to watch, in my opinion.

Finally, thanks to @wodeni, this PR also adds dark mode support to the demo.

Webpack (which we use for Docusaurus) doesn't seem to correctly handle the spread operator for ES6 iterators. Specifically, given this:
```typescript
const [...foo] = arr.keys();
```
it assigns this to `foo`:
```javascript
arr.keys().slice(0)
```
And it translates this:
```typescript
[...arr.keys()]
```
into this:
```javascript
[].concat(arr.keys()
```
These are very wrong, and they cause things to break when we use `@penrose/core` or `@penrose/components` in `@penrose/docs-site`. To fix this, I replaced this pattern with [`Array.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) throughout our entire codebase.

# Checklist

### DIFF
--- a/packages/components/src/Demo.tsx
+++ b/packages/components/src/Demo.tsx
@@ -11,6 +11,7 @@ const Demo = (props: {
   }[];
   width: string; // the width of each half; total width is twice this
   // height must be equal to width (including in the passed Style canvas!)
+  darkMode: boolean;
 }) => {
   const [index, setIndex] = useState(0);
 
@@ -38,6 +39,7 @@ const Demo = (props: {
         substance={example.sub}
         width={props.width}
         height={props.width}
+        monacoOptions={{ theme: props.darkMode ? "vs-dark" : "vs" }}
       />
       <div style={{ width: props.width, height: props.width }}>
         <Simple
--- a/packages/components/src/Listing.tsx
+++ b/packages/components/src/Listing.tsx
@@ -8,7 +8,7 @@ import {
   SubstanceLanguageTokens,
 } from "./editing/languages/SubstanceConfig";
 
-export const monacoOptions: editor.IStandaloneEditorConstructionOptions = {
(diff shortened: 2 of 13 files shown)


maxkrieger (approved):

samestep:
Re-drafting this until we figure out why it's crashing:

> **This page crashed.**
>
> n.keys(...).slice is not a function

cmumatt:
@samestep IIRC, the demo was removed only because it wasn't working at that time and we had to get the docs-site published.  So, probably just unfortunate timing.  I think, generally, everyone was in favor of an animated demo as long as it worked, loaded fast, and looked nice.

samestep:
@cmumatt OK, makes sense! Thanks for the clarification.

samestep:
Now the error is:

> substance: the @ match target 'types' must be an array of strings, in rule: tokenizer.root: [a-z_A-Z$][\w$]*

If I comment out the call to `monaco.languages.setMonarchTokensProvider` in `Listing.tsx`, the page no longer crashes, but this error appears 49 times in the console:

> `Set not found in the subtype graph.`

wodeni (approved):
Looks great. Thanks again for going through all the pain to find out what went wrong!!

samestep:
> Looks great. Thanks again for going through all the pain to find out what went wrong!!

@wodeni You helped quite a bit too!

PR #960, opened 2022-05-12
build: Disable Babel's iterableIsArray assumption

samestep:
# Description

This PR reverts all the iterator spread removal from #958, instead fixing the Docusaurus issue by disabling Babel's [`iterableIsArray`](https://babeljs.io/docs/en/assumptions#iterableisarray) assumption. This by itself doesn't quite work, as it introduces another error, but that can be resolved by setting `sourceType` to `"unambiguous"`: https://github.com/babel/babel/issues/9187#issuecomment-447704595

The relevant plugin for the original issue seen in #958 is `@babel/plugin-transform-spread`, which [uses `options.loose` as a fallback for the `iterableIsArray` assumption](https://github.com/babel/babel/blob/6dc633e71ef65943554e4087c39a4b939b0eadd0/packages/babel-plugin-transform-spread/src/index.ts#L16). Thus, my guess is that the problem was caused by Docusaurus's Babel preset [enabling the `loose` option](https://github.com/facebook/docusaurus/blob/32ec84ef3c0a238436e913b2026ab809e5750fa8/packages/docusaurus/src/babel/preset.ts#L32).

# Checklist

### DIFF
--- a/packages/components/src/editing/languages/SubstanceConfig.ts
+++ b/packages/components/src/editing/languages/SubstanceConfig.ts
@@ -31,11 +31,11 @@ export const SubstanceLanguageTokens = (
   domainCache: Env
 ): languages.IMonarchLanguage => {
   const refs = {
-    types: Array.from(domainCache.types.keys()),
+    types: [...domainCache.types.keys()],
     functionLikes: [
-      ...Array.from(domainCache.constructors.keys()),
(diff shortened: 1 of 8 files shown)


samestep:
@wodeni Yeah, that would probably be wise; where exactly in the wiki should this go?

PR #966, opened 2022-05-13
test: Warn about type assertions in @penrose/core

samestep:
# Description

This PR enables [`@typescript-eslint/consistent-type-assertions`](https://github.com/typescript-eslint/typescript-eslint/blob/v4.5.0/packages/eslint-plugin/docs/rules/consistent-type-assertions.md) with `{ assertionStyle: "never" }` in `@penrose/core`, at the `"warning"` level because we currently have a bunch of them. I got rid of as many as I could without making nontrivial code changes:

- many were completely unnecessary and could be removed without any further changes
- many could be directly replaced with type annotations, which are safe and should always be preferred
- some could be replaced with non-null assertions, which are still unsafe but are less general and thus preferable
  - we already have [`@typescript-eslint/no-non-null-assertion`](https://github.com/typescript-eslint/typescript-eslint/blob/v4.5.0/packages/eslint-plugin/docs/rules/no-non-null-assertion.md) enabled at the `"warning"` level, so previously our ESLint config was encouraging people to prefer type assertions over non-null assertions, which is the opposite of what we want
  - to remedy that issue, this PR also enables [`@typescript-eslint/non-nullable-type-assertion-style`](https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/non-nullable-type-assertion-style.md) at the `"error"` level
- some could be eliminated via very minor code changes

I left the rest as-is. We now have more ESLint warnings, but the goal is that this discourages us from adding more unsafe type assertions in the future.

# Checklist

### DIFF
--- a/packages/core/.eslintrc.js
+++ b/packages/core/.eslintrc.js
@@ -34,6 +34,10 @@ module.exports = {
     "eslint-plugin-import",
   ],
   rules: {
+    "@typescript-eslint/consistent-type-assertions": [
+      1,
+      { assertionStyle: "never" },
+    ],
     "import/no-cycle": 2,
     "no-fallthrough": 2,
   },
--- a/packages/core/.eslintrc.js
+++ b/packages/core/.eslintrc.js
@@ -38,6 +38,7 @@ module.exports = {
       1,
       { assertionStyle: "never" },
     ],
+    "@typescript-eslint/non-nullable-type-assertion-style": 2,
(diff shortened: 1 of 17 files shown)


cmumatt (approved):
Looks great and a big improvement!  Amazing to see so many now-unnecessary type assertions go away!

PR #968, opened 2022-05-13
test: Use eqeqeq and no-template-curly-in-string

samestep:
# Description

This PR enables the ESLint rules [`eqeqeq`](https://eslint.org/docs/rules/eqeqeq) and [`no-template-curly-in-string`](https://eslint.org/docs/rules/no-template-curly-in-string) at the `"error"` level in `@penrose/core`. I chose these specific ones because they both revealed some code that actually should be changed, rather than just being stylistic or not surfacing any issues in our codebase at all.

# Checklist

### DIFF
--- a/packages/core/.eslintrc.js
+++ b/packages/core/.eslintrc.js
@@ -34,8 +34,10 @@ module.exports = {
     "eslint-plugin-import",
   ],
   rules: {
+    eqeqeq: 2,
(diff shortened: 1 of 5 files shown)


cmumatt (approved):
Glad we are getting errors for these now.  Thanks for doing this!

PR #969, opened 2022-05-13
test: Enable a few good typescript-eslint rules

samestep:
# Description

This PR enables a few good typescript-eslint rules:

- [`@typescript-eslint/no-base-to-string`](https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/no-base-to-string.md) catches a bunch of places where we try to stringify things that would just show up as `"[object Object]"`; I didn't have the bandwidth in this PR to fix these so I just set this rule to `"warning"`, but they are all genuine bugs as far as I can tell
- [`@typescript-eslint/no-confusing-void-expression`](https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/no-confusing-void-expression.md) might be a bit more of a stylistic thing, but I think it's pretty good for clarity
- [`@typescript-eslint/no-unnecessary-boolean-literal-compare`](https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/no-unnecessary-boolean-literal-compare.md) didn't surface much, but it seems like a good thing to have enabled
- [`@typescript-eslint/no-unnecessary-condition`](https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/no-unnecessary-condition.md) surfaced a bunch of bugs and also a bunch of things that are due to us using the type system poorly; again I didn't have the bandwidth here to fix all these so I just made it a `"warning"`, but this is very good to have enabled
- [`@typescript-eslint/prefer-ts-expect-error`](https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/prefer-ts-expect-error.md) doesn't actually affect anything currently, but it is quite good so I'm enabling it proactively

# Checklist

### DIFF
--- a/packages/core/src/synthesis/Synthesizer.ts
+++ b/packages/core/src/synthesis/Synthesizer.ts
@@ -695,7 +695,9 @@ export class Synthesizer {
         possibleOps = del ? [del] : [];
       }
     }
-    log.debug(`Found mutations for delete:\n${showMutations(possibleOps)}`);
+    if (possibleOps.length > 0) {
+      log.debug(`Found mutations for delete:\n${showMutations(possibleOps)}`);
+    }
     return possibleOps;
   };
 
--- a/packages/core/src/synthesis/Synthesizer.ts
+++ b/packages/core/src/synthesis/Synthesizer.ts
@@ -695,10 +695,8 @@ export class Synthesizer {
         possibleOps = del ? [del] : [];
       }
     }
-    if (possibleOps) {
-      log.debug(`Found mutations for delete:\n${showMutations(possibleOps)}`);
-      return possibleOps;
-    } else return [];
+    log.debug(`Found mutations for delete:\n${showMutations(possibleOps)}`);
+    return possibleOps;
   };
 
   // TODO: add an option to distinguish between edited vs original statements?
--- a/packages/core/.eslintrc.js
+++ b/packages/core/.eslintrc.js
@@ -34,6 +34,11 @@ module.exports = {
     "eslint-plugin-import",
   ],
   rules: {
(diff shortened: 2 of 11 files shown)


packages/core/src/synthesis/Synthesizer.ts
```diff
@@ -695,10 +695,8 @@ export class Synthesizer {
         possibleOps = del ? [del] : [];
       }
     }
-    if (possibleOps) {
-      log.debug(`Found mutations for delete:\n${showMutations(possibleOps)}`);
-      return possibleOps;
-    } else return [];
+    log.debug(`Found mutations for delete:\n${showMutations(possibleOps)}`);
```
cmumatt, line 698: This change makes this part of the routine much simpler -- thanks for doing that!  Question: should we continue to wrap the debug message in a conditional to only log "mutations found" when possibleOps is not [] -- I guess it doesn't hurt anything to log it in either case?

samestep: Sure, I can make that change! I'm guessing that was @wodeni's original intent?

samestep:
> Do we need to log an issue so we go back and clean up any remaining warnings and change the mode from warning to error?

Hmm we can; the thing is, a lot of the things we have set to `"warning"` would take a significant amount of refactoring effort to change, because they result from systematic typing deficiencies in our system architecture. If we do end up resolving them to the degree that we can actually change the mode to `"error"`, that would just be a side effect of a much larger effort.

PR #911, opened 2022-03-06
feat: Minkowski penalties for Ellipse-Polygon

jiriminarcik:
# Description

This PR implements the simplest version of implicit-based Minkowski constraints for `Ellipse` and `PolygonLike` shapes. This case requires only quadratic roots, which can be computed analytically without the use of #906.

# Implementation strategy and design decisions

Simple interface for implicit shapes was added `ImplicitEllipse` and `ImplicitHalfPlane` to make the code more readable.

# Examples with steps to reproduce them

To test triangle-ellipse overlap, run `npx roger watch maze.sub ellipse-maze.sty maze.dsl` in `packages/examples/src/minkowski-tests/maze`.

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new ESLint warnings
- [x] I have reviewed any generated changes to the `diagrams/` folder

# Open questions

- The next step is to implement ellipse-ellipse constraints using #906.

PR #977, opened 2022-05-17
feat: Minkowski penalties for Ellipse-Ellipse

jiriminarcik:
# Description

Related PR: #911

This PR implements Minkowski constraints for Ellipses using root solver from #906. It also covers penalties for Ellipse-Circle combination as a special case.

The helper functions for computations involving implicit shapes are now in a new file `packages/core/src/contrib/ImplicitShapes.ts`.

# Checklist

# Open questions

There is only one remaining family of shapes to be covered by Minkowski penalties: Bezier curves.


samestep (changes requested):
Looks incredible, thanks so much for doing this! I'll review more carefully later when I have time, but one thing I want to note right away is that your usage of `numsOf` in non-testing code is a bit of a non-starter. Let's discuss this more when we're both in the office later this morning.

samestep (changes requested):
a few typos

packages/core/src/contrib/Minkowski.ts
```diff
@@ -59,31 +71,7 @@ export const rectangleDifference = (
...
-  const normal = ops.vnormalize(
-    ops.rot90(ops.vsub(lineSegment[1], lineSegment[0]))
-  );
-  const insideValue = ops.vdot(ops.vsub(insidePoint, lineSegment[0]), normal);
-  return ops.vmul(neg(signOf(insideValue)), normal);
-};
-
-/**
- * Return value of the Signed Distance Function (SDF) of a half-plane evaluated at the origin.
+ * Return value of the Signed Distance Function (SFD) of a half-plane evaluated at the origin.
```
samestep, line 74:
```suggestion
 * Return value of the Signed Distance Function (SDF) of a half-plane evaluated at the origin.
```

packages/core/src/contrib/Queries.ts
```diff
@@ -58,3 +59,27 @@ export const polygonLikePoints = ([t, s]: [string, any]): ad.Pt2[] => {
...
   }
 };
+
+/**
+ * Return -1.0 for negative number, +1.0 otherwise.
+ */
+const signOf = (x: ad.Num): ad.Num => {
+  const negative = lt(x, 0);
+  return ifCond(negative, -1, 1);
+};
```
samestep, line 69: Can't/shouldn't you instead just use `sign` from `AutodiffFunctions`?

samestep (commented):
Looks amazing, thank you so much for doing this! I had a few comments.

packages/core/src/contrib/ImplicitShapes.ts
```diff
@@ -0,0 +1,274 @@
...
+const ellipsePolynomialParams = (
+  a: ImplicitEllipse,
+  b: ImplicitEllipse
+): ad.Num[] => {
+  return [
+    ellipsePolynomialAlpha0(a, b),
+    ellipsePolynomialAlpha1(a, b),
+    ellipsePolynomialAlpha2(a, b),
+    ellipsePolynomialAlpha3(a, b),
+    ellipsePolynomialAlpha4(a, b),
```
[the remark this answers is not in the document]
minarj: Unfortunately, β is sometimes undefined due to zero-division.

packages/core/src/contrib/ImplicitShapes.ts
```diff
@@ -0,0 +1,274 @@
...
+  ];
+};
+
+// Return monic polynomial coefficients
+// (the highest order coefficient is ommited and assumed to be 1)
```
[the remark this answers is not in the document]
minarj: fixed

packages/core/src/contrib/ImplicitShapes.ts
```diff
@@ -0,0 +1,274 @@
...
+      mul(mul(sub(b.a, a.a), sub(b.b, a.b)), sub(b.c, a.c)),
+      add(
+        mul(mul(a.a, b.a), mul(squared(sub(b.x, a.x)), sub(b.b, a.b))),
+        mul(mul(a.b, b.b), mul(squared(sub(b.y, a.y)), sub(b.a, a.a)))
+      )
+    )
+  );
+};
+
+// Quintic coefficient from the ellipse-ellipse polynomial
```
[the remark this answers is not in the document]
minarj: fixed

packages/core/src/contrib/Minkowski.ts
minarj: fixed

packages/core/src/contrib/Queries.ts
minarj: Nice catch, didn't notice that. Also removed `msign` from `Functions.ts`

samestep: Cool, thanks! @joshsunshine just to check: is this OK? I vaguely remember having a conversation previously about how `msign` is specifically meant to be distinct from `sign` in that it differs at zero. Should we add it back with a documentation comment clarifying this distinction, or is it OK to remove it?

packages/core/src/contrib/ImplicitShapes.ts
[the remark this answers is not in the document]
samestep: Ah OK, makes sense. Thanks for the clarification!

packages/core/src/contrib/Queries.ts
joshsunshine: @jiriminarcik I don't think we should remove `msign` for the reason @samestep describes above.

minarj: I see.. Don't worry, it's back! Thanks for clarification. :)

samestep (approved):
Looks good to me! Awesome work, thank you :)

samestep:
Oh actually it looks like the tests are failing; I'm guessing once you fix those we should be good to merge this?

packages/core/src/engine/Autodiff.test.ts
```diff
@@ -370,7 +370,7 @@ describe("polyRoots tests", () => {
     expect(gradient[2]).toBeCloseTo(-1 / 3);
   });
 
-  test("quintic", () => {
+  test("quartic", () => {
```
samestep, line 373: Hey @jiriminarcik, I just noticed that you erroneously changed this test description in this PR. The polynomial is indeed quintic, since the fifth-degree coefficient is implicitly 1. I'm opening a PR to change this line back.

PR #979, opened 2022-05-17
feat: Signed distance functions for Penrose shapes

joshsunshine:
# Description

I created a new function in the `Functions.ts` `compDict` called `signedDistance` that generates the Autodiff nodes that compute the distance between a Penrose shape and a point. The math in the function is mostly borrowed from [Inigo Quilez](https://iquilezles.org/articles/distfunctions2d/).

The function works properly for Rectangle, Circle, Text, Image, Polyline, and Polygon. However, it does not yet work for Ellipse and Path. There is a stub function for Ellipse in the codebase but there is a bug in the symbolic differentiator that is preventing it from working properly. 

- Unit tests for signed distance functions are in `contrib/SDF.test.ts`
- End-to-end tests are in `examples/shape-distance'

# Checklist


packages/core/src/contrib/Functions.ts
```diff
@@ -1431,6 +1554,175 @@ export const compDict = {
...
+  // if( p.x>p.y ){ p=p.yx; ab=ab.yx; }
+  const pOffset = ops.vsub(pInput, center);
+  const p = ops.vabs(pOffset);
+  const ab = [radiusx, radiusy];
+  p[0] = ifCond(gt(p[0], p[1]), p[1], p[0]);
+  p[1] = ifCond(gt(p[0], p[1]), p[0], p[1]);
+  ab[0] = ifCond(gt(p[0], p[1]), ab[1], ab[0]);
+  ab[1] = ifCond(gt(p[0], p[1]), ab[0], ab[1]);
+  // float l = ab.y*ab.y - ab.x*ab.x;
+  const l = sub(mul(ab[1], ab[1]), mul(ab[0], ab[0]));
```
jiriminarcik, line 1661: you can use `squared(x)` instead of `mul(x, x)`

packages/core/src/contrib/Functions.ts
```diff
@@ -1431,6 +1554,175 @@ export const compDict = {
...
+ 
+    vec2 r = ab * vec2(co,si);
+	
+    return length(r-p) * msign(p.y-r.y);
+}
+*/
+export const sdEllipseAsNums = (
+  radiusx: ad.Num,
+  radiusy: ad.Num,
+  center: ad.Num[],
```
jiriminarcik, line 1648: we have more accurate typing `ad.Pt2` (it's just `[ad.Num, ad.Num]`)

packages/core/src/contrib/Functions.ts
```diff
@@ -1431,6 +1554,175 @@ export const compDict = {
...
+  // float q = c3 + m2*n2*2.0;
+  const q = add(c3, mul(m2, mul(n2, 2)));
+  // float d = c3 + m2*n2;
+  const d = add(c3, mul(m2, n2));
+  // float g = m + m*n2;
+  const g = add(m, mul(m, n2));
+
+  //if branch
+  // float h = acos(q/c3)/3.0;
+  const hif = div(acos(div(q, c3)), 3);
```
jiriminarcik, line 1682: are we using the same `acos` as iquilezles?

packages/core/src/contrib/Functions.ts
```diff
@@ -1431,6 +1554,175 @@ export const compDict = {
...
+  const ry = mul(sub(s, t), sqrt(3));
+  // float rm = sqrt( rx*rx + ry*ry );
+  const rm = sqrt(add(mul(rx, rx), mul(ry, ry)));
+  // co = ry/sqrt(rm-rx) + 2.0*g/rm;
+  const coelse = add(div(ry, sqrt(sub(rm, rx))), mul(2, div(g, rm)));
+
+  // co = (co-m)/2.0;
+  // if (d<0.0)
+  const co_pred = ifCond(lt(d, 0), coif, coelse);
+  const co = div(min(co_pred, m), 2);
```
jiriminarcik, line 1716:
```suggestion
  const co = div(sub(co_pred, m), 2);
```

packages/examples/src/shape-distance/shape-distance.sty
```diff
@@ -0,0 +1,69 @@
...
+forall Polyline l {
+    l.icon = Polyline {
+        strokeWidth : 8
+        points : [(-200,-200), (0,200), (200,-300), (-200,150), (250,150)]
+    }
+}
+
+forall Text t {
+    shape t.icon = Text {
+      string: "Hello Nimo"
```
samestep, line 48: 😄 @wodeni

samestep:
> There is a stub function for Ellipse in the codebase but there is a bug in the symbolic differentiator that is preventing it from working properly.

@joshsunshine Can you please create a GitHub issue about this bug?

samestep:
@joshsunshine To better match the `for` loop for the polygon example, would it work to do this instead?
```diff
diff --git a/packages/core/src/contrib/Functions.ts b/packages/core/src/contrib/Functions.ts
index 778a2a302..6cbb4b106 100644
--- a/packages/core/src/contrib/Functions.ts
+++ b/packages/core/src/contrib/Functions.ts
@@ -1499,7 +1499,7 @@ export const compDict = {
       let d = ops.vdot(ops.vsub(p, v[0]), ops.vsub(p, v[0]));
       let ess: ad.Num = 1.0;
       let j = v.length - 1;
-      for (let i = 0; i < v.length; i++) {
+      for (let i = 0; i < v.length; j = i, i++) {
         const e = ops.vsub(v[j], v[i]);
         const w = ops.vsub(p, v[i]);
         const clampedVal = clamp([0, 1], div(ops.vdot(w, e), ops.vdot(e, e)));
@@ -1515,8 +1515,6 @@ export const compDict = {
         const c8 = and(and(c5, c6), c7);
         const negEss = mul(-1, ess);
         ess = ifCond(or(c4, c8), negEss, ess);
-        // last line to match for loop in code we are borrowing from
-        j = i;
       }
       const result = mul(ess, sqrt(d));
       return {
```

packages/core/src/contrib/Functions.ts
```diff
@@ -1431,6 +1555,195 @@ export const compDict = {
...
+  const ba = ops.vsub(b, a);
+  const h = clamp([0, 1], div(ops.vdot(pa, ba), ops.vdot(ba, ba)));
+  return ops.vnorm(ops.vsub(pa, ops.vmul(h, ba)));
+};
+
+const sdPolyline = (s: Polyline, p: ad.Num[]): ad.Num => {
+  const dists: ad.Num[] = [];
+  for (let i = 0; i < s.points.contents.length - 1; i++) {
+    const start = s.points.contents[i];
+    const end = s.points.contents[i + 1];
```
samestep, line 1582: Just to make sure I understand: this works because polylines cannot be closed?

packages/core/src/contrib/Functions.ts
```diff
@@ -1431,6 +1555,195 @@ export const compDict = {
...
+  return minN(dists);
+};
+
+export const sdEllipse = (s: Ellipse, p: ad.Num[]): ad.Num => {
+  return sdEllipseAsNums(s.rx.contents, s.ry.contents, s.center.contents, p);
+};
+
+/*
+  float msign(in float x) { return (x<0.0)?-1.0:1.0; }
+  TODO: merge with Jiri's signOf function
```
samestep, line 1594: From #977?

packages/core/src/contrib/Functions.ts
```diff
@@ -1431,6 +1555,195 @@ export const compDict = {
...
+    }
+    co = (co-m)/2.0;
+
+    float si = sqrt( max(1.0-co*co,0.0) );
+ 
+    vec2 r = ab * vec2(co,si);
+	
+    return length(r-p) * msign(p.y-r.y);
+}
+*/
```
samestep, line 1647: Where exactly does this reference code come from? It does not match the `sdEllipse` code on this page: https://iquilezles.org/articles/distfunctions2d/

samestep (approved):
Looks great, thanks so much for doing this! I read everything except the implementation of `sdEllipseAsNums`, and it seems like @jiriminarcik already looked at that more closely.

packages/core/src/contrib/Functions.ts
joshsunshine: Yes.

packages/core/src/contrib/Functions.ts
joshsunshine: It is from the actual shadertoy code that is linked from the iquielezes blog post (https://www.shadertoy.com/view/4sS3zz). 
-Josh

packages/core/src/contrib/Functions.ts
joshsunshine: Yes. Is there a way to mark this todo as connected to a particular issue?

samestep: @joshsunshine Usually I just do that by putting the URL of the GitHub issue in the comment, because then in VS Code you can just cmd+click on it to open the issue in the browser.

PR #984, opened 2022-05-18
refactor: Rewrite half of the Style compiler

### LINKED ISSUE #720: Make anonymous shapes legal
It doesn't make much sense to give every shape a name, since most shapes will never be referenced again.  For instance, right now you have to write things like
```
forall Point x; Point y
where Connected(x,y) {
   shape myLine = Line {
      ...
   }
}
```
even if `myLine` will never be referenced.  (Seems you can also drop the type name `shape`, but you still have to pick a name for the shape.)

Forcing shapes to be names effectively pollutes the namespace with a bunch of throwaway names that will not be used, which makes code harder to read/understand.  ("Will I need to remember this name for later?")

Would be great to allow anonymous shapes, e.g.,
```
forall Point x; Point y
where Connected(x,y) {
   shape Line {
      ...
   }
}
```
or perhaps just
```
forall Point x; Point y
where Connected(x,y) {
   Line {
      ...
   }
}
```
In this scenario, **Style programmers would still have the _option_ to give a shape a name**; they simply wouldn't be _required_ to.

Not clear there are currently any good alternatives, other than cooking up names for each shape.

**Additional context**
It seems like it shouldn't be too hard to implement at the parser level, without doing any sort of deep analysis/lookahead/etc.:
Suppose the parser sees the pattern
`shape <token> {`
then it can simply pretend that this line said
`shape <generated_name> = <token> {`
where `generated_name` is some automatically-generated name.  For instance, `anonymous_shape_NNN` where `NNN` is an integer that gets incremented each time an anonymous shape is encountered.  This way, there is still a (somewhat) meaningful name available for the inspector/debugger to display.

### LINKED ISSUE #888: Style compiler re-write

### LINKED ISSUE #890: Find DOFs in a more principled way
The Style compiler seems to treat all float properties as "varying" (with some hardcoded exceptions). So the list of DOFs the optimizer receives is usually a superset of the actual DOFs in the optimization graph. 

In principle, the Style compiler should analyze the Style program to decide which paths are variables in the optimization graph, not by generally assuming float properties are variables by default. Concretely, we should start by removing this list: https://github.com/penrose/penrose/blob/280fed5e8377ef0e474651d8fae648acac6b19f8/packages/core/src/compiler/Style.ts#L2230

### LINKED ISSUE #928: Reduce redundant comp graph construction in the evaluator
* __The main goal of the evaluator__: take the `Translation`, which contains expressions defined in Style `radius: 3 + b.shape.radius`, and turn it into a list of shapes with actual values `radius: 5`.
* To accomplish this, the evaluator walks through the translation, and turns Style expressions like `3 + b.shape.radius` into `add(3, 2)`. 
* Effectively, the evaluator constructs an autodiff computational graph from Style expressions. The leaf nodes of that graph are properties of shapes.
* In this process, the autodiff engine also computes gradients in the graph.
* The evaluator then extracts the numerical values for all leaf nodes and __discards the entire graph (including the gradients)__.
* Therefore, the evaluator is doing a lot of redundant computation per step of the optimizer, because we re-evaluate after each step to show the “animation.”
* The sensible solution: because the computational graph never really changes, we only need to construct the graph once and just plug in new values to get new lists of shapes.
* This may speed up evaluation by a lot, because we are basically turning a complicated JS module into one compiled JS function.
* Some future work
  * Error handling: the evaluator does try to catch some errors (e.g. circular dependencies), so we need to catch them at compile time now. If we actually encode a computation graph, this should be easy to do.
  * Changes to the computational graph: we don’t have a use case for that now, but if someone decides to change the Style program, the whole graph is re-compiled, and that’s costly. If there’s ever a need to speed that up, we can look into incremental compilation of the graph.

samestep:
# Description

Resolves #720, resolves #888. resolves #890, resolves #928.

This PR primarily replaces `engine/Evaluator` with a three-stage compiler in `compiler/Style` that constructs the computation graphs and other necessary `State` data in a much more straightforward way:

1. process `delete`s and `override`s
2. make a dependency graph of all field/property definitions
3. compile all expressions in topological sorted order

We no longer have a concept of `varyingPaths`, which eliminates a ton of complexity. Instead, `shapes/Samplers` now defines a `Context` type (replacing the previous `Context` type from `contrib/Functions`) which takes in a `malloc`-style `InputFactory` function that takes either a sampler (for sampled values that the optimizer can change) or an initial value (for pending label properties), and returns an `ad.Input`.

The Style compiler defines this callback and keeps track of the samplers/metadata in an array the same length as `varyingValues`. This makes resampling incredibly simple: seed an RNG with the variation, then iterate through the samplers array and call each one in turn. Thus, this PR simply deletes the overly complex `engine/Resample`.

This reworking of how randomness works in Penrose fixes a lot of things (e.g. `sampleColor` as mentioned in a now-deleted comment in `types/style`, and also the annoying pattern of having to resample after `prepareState`), but because it changes the way sampling works, it means all the preexisting `variation`s for diagrams in our registry are outdated. @wodeni and I came up with new `variation`s for the affected diagrams.

Because the comp graph tab in the UI made heavy use of `varyingPaths`, and there wasn't a straightforward way to rework it to not use those, this PR removes that tab. From a Slack conversation with @wodeni:

> I think nobody found graphutils to be helpful atm. Perhaps the right way is not to support it as it is, and figure out a better viz for graphs

- Previously `engine/Optimizer` mutated the state's `params`. After the compiler rewrite, this caused some tests to break, so I changed it to instead first shallow-clone the `params` field before mutating it.

- We no longer use [`rfdc`](https://www.npmjs.com/package/rfdc), so this PR removes it from our `dependencies`.

- I changed all our direct imports of members of [Immutable.js](https://immutable-js.com/) to instead just import the whole thing as `im`, to avoid confusing its [`Map`](https://immutable-js.com/docs/v4.1.0/Map/) with the builtin JavaScript [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

- The `IntV` subtype of `Value` was previously unused (couldn't be produced by the Style parser), so this PR removes it.

- This PR removes the following Style AST types:

  - `AccessPath`
  - `FieldPath`
  - `InternalLocalVar`
  - `ListAccess`
  - `LocalVar`
  - `Matrix`
  - `MatrixAccess`
  - `PluginAccess`
  - `PropertyPath`
  - `VaryAD`
  - `VaryInit`
  - `VectorAccess`

  Some were previously subtypes of `Path` which have now been merged into a single type (along with the modifications that brings to the parser), and the others were artifacts of the previous Evaluator which are now unnecessary in the new compiler.

- For some reason, the `LabelCache` type was previously just a list of pairs instead of an actual `Map`. This PR fixes that, and deletes the now-unnecessary `retrieveLabel` function from `utils/CollectLabels`.

- Previously `shapes/Samplers` had a bunch of generic utility functions for constructing `Value`s. This PR moves those into `utils/Util`.

# Checklist

# Open questions

- This PR adds a `warnings` field to `State`, which is currently only used in `Style.test.ts`. In the future we should expose these warnings in the UI.
- Every `ad.Input` still has a `val` field, which is currently only used for convex partitioning. As we have discussed before, this convex partitioning only happens once at autodiff graph construction time, and remains fixed during optimization and even through resampling. This constitutes a way in which resampling can act different from the initial sampling with a given seed. Our goal is to eventually get rid of the `val` field entirely by coming up with a way to do convex partitioning directly inside the autodiff engine.
- We don't currently have any tests for `renderer/Staging` so I'm not sure whether I reimplemented it correctly. In particular, judging from the comments in the previous implementation, it seems like it was supposed to use a specific order for the Substance objects, but in the new implementation I made no effort to ensure that same order. Is this important?


wodeni:
Suggested edit:
```diff
diff --git a/packages/core/src/utils/Util.ts b/packages/core/src/utils/Util.ts
index cd333520f..897e828a0 100644
--- a/packages/core/src/utils/Util.ts
+++ b/packages/core/src/utils/Util.ts
@@ -1,5 +1,4 @@
 import * as _ from "lodash";
-import { times } from "lodash";
 import seedrandom from "seedrandom";
 import { LineProps } from "shapes/Line";
 import * as ad from "types/ad";
@@ -109,17 +108,6 @@ export const zip3 = <T1, T2, T3>(
 
 const RAND_RANGE = 100;
 
-/**
- * Generate a random float. The maximum is exclusive and the minimum is inclusive
- * @param min minimum (inclusive)
- * @param max maximum (exclusive)
- */
-export const randFloats = (
-  rng: seedrandom.prng,
-  count: number,
-  [min, max]: [number, number]
-): number[] => times(count, () => randFloat(rng, min, max));
-
 /**
  * Generate a random float. The maximum is exclusive and the minimum is inclusive
  * @param min minimum (inclusive)

```

wodeni:
Suggested edit:
```diff
diff --git a/packages/core/src/utils/Util.ts b/packages/core/src/utils/Util.ts
index cd333520f..897e828a0 100644
--- a/packages/core/src/utils/Util.ts
+++ b/packages/core/src/utils/Util.ts
@@ -1,5 +1,4 @@
 import * as _ from "lodash";
-import { times } from "lodash";
 import seedrandom from "seedrandom";
 import { LineProps } from "shapes/Line";
 import * as ad from "types/ad";
@@ -109,17 +108,6 @@ export const zip3 = <T1, T2, T3>(
 
 const RAND_RANGE = 100;
 
-/**
- * Generate a random float. The maximum is exclusive and the minimum is inclusive
- * @param min minimum (inclusive)
- * @param max maximum (exclusive)
- */
-export const randFloats = (
-  rng: seedrandom.prng,
-  count: number,
-  [min, max]: [number, number]
-): number[] => times(count, () => randFloat(rng, min, max));
-
 /**
  * Generate a random float. The maximum is exclusive and the minimum is inclusive
  * @param min minimum (inclusive)

```

wodeni:
Suggested edit:
```diff
diff --git a/packages/core/src/utils/Util.ts b/packages/core/src/utils/Util.ts
index cd333520f..897e828a0 100644
--- a/packages/core/src/utils/Util.ts
+++ b/packages/core/src/utils/Util.ts
@@ -1,5 +1,4 @@
 import * as _ from "lodash";
-import { times } from "lodash";
 import seedrandom from "seedrandom";
 import { LineProps } from "shapes/Line";
 import * as ad from "types/ad";
@@ -109,17 +108,6 @@ export const zip3 = <T1, T2, T3>(
 
 const RAND_RANGE = 100;
 
-/**
- * Generate a random float. The maximum is exclusive and the minimum is inclusive
- * @param min minimum (inclusive)
- * @param max maximum (exclusive)
- */
-export const randFloats = (
-  rng: seedrandom.prng,
-  count: number,
-  [min, max]: [number, number]
-): number[] => times(count, () => randFloat(rng, min, max));
-
 /**
  * Generate a random float. The maximum is exclusive and the minimum is inclusive
  * @param min minimum (inclusive)

```

wodeni:
Suggested edit:
```diff
diff --git a/packages/core/src/utils/Util.ts b/packages/core/src/utils/Util.ts
index cd333520f..897e828a0 100644
--- a/packages/core/src/utils/Util.ts
+++ b/packages/core/src/utils/Util.ts
@@ -1,5 +1,4 @@
 import * as _ from "lodash";
-import { times } from "lodash";
 import seedrandom from "seedrandom";
 import { LineProps } from "shapes/Line";
 import * as ad from "types/ad";
@@ -109,17 +108,6 @@ export const zip3 = <T1, T2, T3>(
 
 const RAND_RANGE = 100;
 
-/**
- * Generate a random float. The maximum is exclusive and the minimum is inclusive
- * @param min minimum (inclusive)
- * @param max maximum (exclusive)
- */
-export const randFloats = (
-  rng: seedrandom.prng,
-  count: number,
-  [min, max]: [number, number]
-): number[] => times(count, () => randFloat(rng, min, max));
-
 /**
  * Generate a random float. The maximum is exclusive and the minimum is inclusive
  * @param min minimum (inclusive)

```

wodeni:
Suggested edit:
```diff
diff --git a/packages/core/src/utils/Util.ts b/packages/core/src/utils/Util.ts
index cd333520f..897e828a0 100644
--- a/packages/core/src/utils/Util.ts
+++ b/packages/core/src/utils/Util.ts
@@ -1,5 +1,4 @@
 import * as _ from "lodash";
-import { times } from "lodash";
 import seedrandom from "seedrandom";
 import { LineProps } from "shapes/Line";
 import * as ad from "types/ad";
@@ -109,17 +108,6 @@ export const zip3 = <T1, T2, T3>(
 
 const RAND_RANGE = 100;
 
-/**
- * Generate a random float. The maximum is exclusive and the minimum is inclusive
- * @param min minimum (inclusive)
- * @param max maximum (exclusive)
- */
-export const randFloats = (
-  rng: seedrandom.prng,
-  count: number,
-  [min, max]: [number, number]
-): number[] => times(count, () => randFloat(rng, min, max));
-
 /**
  * Generate a random float. The maximum is exclusive and the minimum is inclusive
  * @param min minimum (inclusive)

```

wodeni (approved):

packages/core/src/__tests__/overall.test.ts
```diff
@@ -67,17 +67,18 @@ describe("Determinism", () => {
     const stateSample3Opt = resSample3Opt.value;
     const svgSample3Opt = await render(stateSample3Opt);
 
-    expect(svgSample1NotOpt).not.toBe(svgSample1Opt); // optimization does something
-    expect(svgSample1NotOpt).not.toBe(svgSample2NotOpt); // resampling is different from initial sampling
-    expect(svgSample1NotOpt).not.toBe(svgSample2Opt); // optimizing resample doesn't get us back
```
joshsunshine, line 72: Why did you remove these comments?

joshsunshine (approved):

joshsunshine:
@samestep @wodeni What is the purpose of `renderer/Staging`?

packages/core/src/__tests__/overall.test.ts
samestep: @joshsunshine I didn't remove them; I just reorganized the assertions and prefaced each group with a comment that describes the whole group, rather than putting a comment at the end of every assertion line.

samestep:
> What is the purpose of `renderer/Staging`?

@joshsunshine It was created by @heleaf to turn a diagram into a sequence of diagrams that gradually bring in the shapes attached to each Substance object. I think it is currently supported by `@penrose/automator`, but I haven't tried it.

PR #988, opened 2022-05-19
ci: Check platform-specific diagrams

samestep:
# Description

This PR removes our `diagrams/` folder in favor of storing CI-generated diagrams in separate branches (allowing us to use `Text` in the registry), finishes #876, undoes the `Equation` substitution from #1153, and adds back the `Text` from #1167. This also has the benefit that people no longer need to manually run `automator` on their PR before pushing if it results in changes to `diagrams/`; those changes are instead discovered automatically by CI.

I also updated our CI script to autogenerate the README, so now it builds just that one diagram (which does not use `Text`) and puts it in `docs/assets/output.svg`. Incidentally, this caused me to notice a few files under `.github/` that I created before (e.g. for #972) which nobody uses now (mostly because of #1083), so I deleted those.

See [the comment](https://github.com/penrose/penrose/pull/988#issuecomment-1357051436) posted by **github-actions** below.

# Checklist

# Open questions

- This strategy, of having CI push to a different branch in the repo and/or comment on the PR with a summary, only works if the contributor already has push access to the Penrose repo. In other words, the only parts of this machinery that work for external contributors are the artifact upload and the summary text on the CI job itself. I think this is still worth it because it allows us to include `Text` examples in the registry, but might want to think about how we can get some more signal on registry diagram changes for external contributors.
- Now that we can use `Text` in the registry, we should add back all the examples that we were previously leaving out because of that restriction! @keenancrane
- This decoupling in this PR opens up an opportunity for us to show PR reviewers a more focused generated registry image diff, instead of showing a diff item for every single image that has a slightly different SVG. For instance, we could run the SVGs through a rasterizer, then compare those rasterized images to the ones from `main`, and only include in the diff the ones that change noticeably.


wodeni:
Looking good! I haven't read through the CI script carefully enough, but had one question: is the textual diff in the comment the only signal? Are contributors expected to diff the `ci/*` branches manually to find out how diagrams visually differ?

> This decoupling in this PR opens up an opportunity for us to show PR reviewers a more focused generated registry image diff, instead of showing a diff item for every single image that has a slightly different SVG. For instance, we could run the SVGs through a rasterizer, then compare those rasterized images to the ones from main, and only include in the diff the ones that change noticeably.

My understanding is that this is a solution to the visual diff problem above?

samestep:
@wodeni

> had one question: is the textual diff in the comment the only signal? Are contributors expected to diff the `ci/*` branches manually to find out how diagrams visually differ?

No; take a look at [the bot comment above](https://github.com/penrose/penrose/pull/988#issuecomment-1357051436): the "Registry diff" heading is a link to this page, which shows a visual diff with the same information as what we had before: https://github.com/penrose/penrose/compare/ci/refs/heads/main..ci/refs/pull/988/merge

wodeni (approved):

PR #999, opened 2022-05-20
feat: Lagrange bases

keenancrane:
# Description

This PR adds a new example `lagrange-bases`, which is described in the README.  Mostly this is just a dogfooding example; in the process of dogfooding we (re)-discovered some things like exponential cost of long selectors (which put some limitations on the sophistication of this example).  It also changes the `contains1D` function to use a simple `max` on the difference of interval endpoint values, rather than `ifCond` on the square of the difference, which was causing the optimizer to hang on this and several other examples.  Finally, this PR adds a new function `interpolateQuadraticFromPoints` which draws a quadratic Bézier curve that interpolates the three given points, rather than using the 2nd point as a control point.

# Checklist


diagrams/allShapes-dashedShapes.svg
```diff
@@ -130,9 +130,9 @@
     <title>L.icon</title>
   </g>
   <rect
-    x="726.4839694022653"
+    x="10.5452332147903"
     y="408.9659586607471"
-    width="65.1769202023942"
+    width="-406.09798464016694"
```
samestep, line 135: Strange that the `width` became negative...

packages/examples/src/lagrange-bases/README.md
```diff
@@ -0,0 +1,9 @@
+# lagrange-bases
+
+`lagrange-bases` is a stub for a package that draws meshes made of higher-order Lagrange finite elements. Currently it supports only linear triangular and quadratic elements, because quadratic quadrilateral elements, and other higher-order elements, require so many constructor arguments that we run into the [exponential scaling bug in Style selectors](https://github.com/penrose/penrose/issues/566).
+
+The basic idea is that `Nodes` can be connected by `Elements` of any order. The rule for elements of type `QuadraticTriangle` in `lagrange-bases.sty` serves as a nice simple example of drawing quadratic Bézier curves; note that because we use the function `interpolateQuadraticFromPoints` rather than `quadraticCurveFromPoints`, these curves interpolate the three given points (rather than using the middle point as an approximated control point). There are some other cool things we could in principle do here (like automatically restricting edges that belong to multiple elements to polynomials of the lower degree), but see again the note about the bug above, which makes it hard to use long selectors.
```
samestep, line 5: Awesome, thanks for documenting this so well!

PR #1000, opened 2022-05-20
feat: unify browser-ui and editor

maxkrieger:
exciting times indeed

See [here](https://github.com/penrose/penrose/pull/1000#issuecomment-1135071568) for more docs.

image


wodeni:
# Description

This PR introduces a new version of the Penrose IDE (`@penrose/editor`) for both end-users and Penrose developers. `editor` unifies `browser-ui` and `panels` by allowing both (1) local development mode with `roger` and (2) online editing mode. 

For (1), `roger` is re-structured to work with folders instead of individual trios. `roger watch` now takes in a single argument of a path to a folder, and recursively searches for files of `.sub`, `.sty`, and `.dsl` extensions. If `process.env` is `development` (which is true when running `yarn start`. the value is `production` for deployed IDE). A `roger` panel will show up in the IDE with dropdowns for selecting files in the folder that `roger` serves. 

For (2), `editor` allows file saving through both local-storage and GitHub gists. File sharing is also implemented similar to what `panels` had, i.e. through GitHub gist. 

# Implementation

* `EditorPane` is now a part of `@penrose/components`, reused for all three languages
* Other components like `Opt` and `StateiInspector` are included in the `components` folder
* Because `editor` uses [recoil](https://recoiljs.org/) for state management, no props need to be passed to these components. Instead, a call to `useRecoilState` will fetch the right data from the global state of the application

Issue #775, opened 2022-01-07
Allow Boolean logic in `where` clause of Style selectors


keenancrane:
Issue #290 already discusses the need for more expressive universal quantifiers in Style selectors.

A separate but related issue is more expressive statements in the `where` clause.  At present, we can only provide a list of predicates, functions, and constructors that _must_ appear, e.g.,

```
forall A a; B b; C c
where Predicate(a); Function(a,b); Constructor(a,b,c)
```

It would be very helpful (and very natural) to extend this clause with basic Boolean logic, e.g.,

```
forall A a; B b; C c
where (Predicate(a) and Function(a,b)) or not Constructor(a,b,c)
```

The semantics seem pretty unambiguous; there are perhaps some minor quibbles to be had about syntax.  For instance, should we stick only with natural language keywords `and`, `or`, `not`, or should we also allow syntax like

```
forall A a; B b; C c
where (Predicate(a) && Function(a,b)) || !Constructor(a,b,c)
```

The latter feels decidedly "less Penrose," but is maybe more readable for some.  Similarly, should we provide additional operations like `xor`?  Mathematical set builder notation typically just uses a few [common logical connectives](https://en.wikipedia.org/wiki/Logical_connective#Common_logical_connectives), of which XOR is not one; XOR is also far less common than NOT, AND, and OR in many general-purpose programming languages.  Perhaps best to just start with the "big three" and see if we find ourselves/users seeking others.

keenancrane:
A baby version of this change would be to replace the syntax

```
where PredicateA(x); PredicateB(x); PredicateC(x)
```

with

```
where PredicateA(x) and PredicateB(x) and PredicateC(x)
```

In other words, at least make it clearer from the syntax that the list of conditions following a `where` statement is a logical **and** (rather than, say, a logical **or**).  This meaning may be clear to us as expert users of Penrose, but is not clear to a novice reader.  I suspect it also affects cognition: when the brain sees the word "and", it has an easier time thinking about what's going on than if it just sees a semicolon.  Finally, this change naturally paves the way to the more general Boolean syntax described above.

liangyiliang:
Hmmm ... I think this might break the general algorithm behind the _faster matching semantics_ which requires the predicates to all hold ... we'll have to think more about this

PR #1007, opened 2022-05-23
feat: Persistent homology example

jiriminarcik:
# Description

This PR adds new example illustrating Persistent homology. 

The example was added to the registry.

Run `npx roger watch persistent-homology.sub persistent-homology.sty persistent-homology.dsl` from`/packages/examples/src/persistent-homology`. Or `npx roger watch persistent-homology.sub persistent-homology.sty persistent-homology-basic.dsl` for a simplified version without labels.

# Checklist

# Open questions

- The `.sub` file can be simplified after #775 is resolved - we are unable to use negator in selector within the `.sty` file now.
- The complexity of this example was constrained by  #1008.


samestep (approved):
This is super cool!

joshsunshine:
@jiriminarcik can you explain exactly how you would you would use `not` in this example?

minarj:
> @jiriminarcik can you explain exactly how you would use `not` in this example?

In this example I need to make sure that two circles overlap _iff_ the respective points satisfy e.g. the `ConnectedOnFrame1` predicate, otherwise they must be disjoint. It is straightforward to enforce the positive condition, but not the negative one.

The current work-around involves defining the `NotConnectedOnFrame1` predicate and explicitly writing down all combinations of points for which `ConnectedOnFrame1` does not hold in `.sub` file.

With `not`, we could just use `ConnectedOnFrame1` and write (in `.sty`):
```
forall Point p1; Point p2 
where not ConnectedOnFrame1(p1, p2) {
  ensure disjoint(p1.circle1, p2.circle1, 10)
}
```
instead of the current work-around:
```
forall Point p1; Point p2 
where NotConnectedOnFrame1(p1, p2) {
  ensure disjoint(p1.circle1, p2.circle1, 10)
}
```
 

This would eliminate most of the lines in  `packages/examples/src/persistent-homology/persistent-homology.sub` (see comment on line 43).

Issue #566, opened 2021-05-25
Style selector matching doesn't scale; freezes app


k-qy:
**Describe the bug**

Style selectors with many variables to match, in a large-ish Substance program, seem to take a long time.

Domain expert 1 reported: 

> Even without Zoom, Penrose hangs when I uncomment all the disjointness selectors. But I think this only adds one additional disjoint constraint to the problem relative to what we were looking at, so my uninformed guess is that it's freezing up while doing the pattern matching.

I believe this is the case. It is a known problem from past versions of the system. Indeed, it may be because I didn't port this optimization: 

https://github.com/penrose/penrose/blob/main/packages/core/src/compiler/Style.ts#L601

**To Reproduce**

https://github.com/srossd/penrose/commits/monoidal-category (waiting for latest changes to be pushed) — or any selector with a lot of distinct variables

**Expected behavior**

Style pattern matches don't take exponential time

keenancrane:
I'm also bumping up against this issue.  Here's a minimal version of what I'm trying to do:

```
-- minimal.dsl
type S
type T
constructor f( T t1, T t2, T t3, T t4, T t5, T t6, T t7, T t8 ) -> S
```

```
-- minimal.sty
canvas {
   width = 240
   height = 180
}

forall S s; T t1; T t2; T t3; T t4; T t5; T t6; T t7; T t8
where s := f( t1, t2, t3, t4, t5, t6, t7, t8 ) {
   s.shape = Circle {
      center: (0,0)
      r: 10.0
   }
}
```

```
-- minimal.sub
T t1, t2, t3, t4, t5, t6, t7, t8
S s := f( t1, t2, t3, t4, t5, t6, t7, t8 )
```

The browser just hangs (and I'd guess it craps out even before we get to eight arguments).

liangyiliang:
I'm closing this since _faster matching_ has been implemented.

PR #1016, opened 2022-05-24
fix: selector matching `n^m` complexity

wodeni:
# Description

Related issue/PR: #566 

This PR slightly improves the performance of the selector matching algorithm. The current implementation enumerates all possible substitutions for each type declaration in the selector head position, and return a Cartesian product with all previous substitutions found. For example, take the following sub/sty pair:
```
-- .sub
T t1, t2, t3, t4

-- .sty
T s1, s2 {
}
```

Current impl will produce `4^2 = 16` substitutions. However, it would include cases such as `{s1: t1, s2: t1}` because the Cartesian product doesn't produce a bijection. 

Instead, this PR adds a duplicate check in `cartesianProduct`, and now this step of selector matching will be `n choose m` time instead of `n^m`, which is [asymptotically faster](https://proofwiki.org/wiki/N_Choose_k_is_not_greater_than_n%5Ek).

A bigger rewrite of the selector matching dynamics will happen later. This is just a small step to get bigger examples in the registry compile a bit faster.

`lagrange-bases-lagrange-bases` in https://82210426.penrose-72l.pages.dev/try

# Checklist

### DIFF
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -120,7 +120,7 @@ import {
 import { checkTypeConstructor, isDeclaredSubtype } from "./Domain";
 
 const log = consola
-  .create({ level: LogLevel.Debug })
+  .create({ level: LogLevel.Warn })
   .withScope("Style Compiler");
 
 //#region consts
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -21,7 +21,7 @@ import {
   propertiesOf,
 } from "engine/EngineUtils";
 import { alg, Edge, Graph } from "graphlib";
-import _, { range } from "lodash";
+import _, { range, some } from "lodash";
 import nearley from "nearley";
 import { lastLocation } from "parser/ParserUtil";
 import styleGrammar from "parser/StyleParser";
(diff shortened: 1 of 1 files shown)


samestep:
Correction: @joshsunshine pointed out that this actually changes it to `n permute m`, not `n choose m`.

wodeni:
> Correction: @joshsunshine pointed out that this actually changes it to `n permute m`, not `n choose m`. Also, we should note that this explicitly changes the semantics. For instance, this Substance program:
> 
> ```
> Node a, b
> Edge(a, a)
> Edge(a, b)
> ```
> 
> would not cause `Edge(a, a)` to match in this Style program:
> 
> ```
> forall Node x, y where Edge(x, y) { }
> ```

Yep, you're right. I later did this manually and realized `4 choose 2 = 6`, and this fix only brings it down to `4 permutes 2 = 12`. Anyway, the semantics aren't changed, because there's a check later for the same thing: https://github.com/penrose/penrose/pull/1016/files#diff-733e806e9efda85c1ac245b12dead521c03112aee625a21928ed6eb03ef2875dL1476. Note that doing this check earlier still improves performance because it cuts down substitutions much earlier.

samestep:
@wodeni Ah ok, thanks for the clarification! Although I am a bit confused, because Josh explicitly told me on Wednesday that we chose to change the semantics in this case; @joshsunshine I'm guessing I simply misunderstood what you meant?

PR #1022, opened 2022-05-26
feat: Walk on spheres

keenancrane:
This PR adds a new example `walk-on-spheres`.  See the README for details.


packages/examples/src/walk-on-spheres/walk-on-spheres.dsl
```diff
@@ -0,0 +1,22 @@
+type Domain -- a region in ℝⁿ
```
samestep, line 1: Unicode, nice 😎

packages/examples/src/walk-on-spheres/Laplace.sub
```diff
@@ -0,0 +1,20 @@
...
+-- a basic Laplace equation Δu = 0, which involves both
+-- boundary samples x and source samples y.
+
+Domain U
+Step x0, x1, x2, x3, x4
+
+x1 := sampleBoundary( x0 )
+x2 := sampleBoundary( x1 )
+x3 := sampleBoundary( x2 )
+x4 := sampleBoundary( x3 )
```
samestep, line 13: This is just a matter of aesthetics, but could we do this instead? (same for the other `.sub` files in this directory)
```suggestion
Step x0

Step x1 := sampleBoundary( x0 )
Step x2 := sampleBoundary( x1 )
Step x3 := sampleBoundary( x2 )
Step x4 := sampleBoundary( x3 )
```

samestep (approved):
This is incredible, thanks so much for making this @keenancrane! Possibly our most compelling Penrose example yet.

joshsunshine:
@samestep I think we should merge this now. I don't think your comment above is significant enough to delay.

joshsunshine (approved):

keenancrane:
@samestep Yeah, I can see an argument for that coding style also.  I think ultimately how this _should_ look is some kind of array syntax:

```
Step x[5]
for i = 1, .., length(x)
   x[i] = sampleBoundary( x[i-1] )
```

Of course, that doesn't exist yet!  It may also be useful for people to see that declaring all your variables up front is _possible_ as one way to write Substance code.  So, I'm gonna leave it for now… (and squash/merge).  Thanks!

Issue #1024, opened 2022-05-26
Don't emit Penrose-specific attributes in SVG


keenancrane:
Earlier this year @cmumatt did an [awesome PR](https://github.com/penrose/penrose/pull/749) to pass through arbitrary attributes from Style to SVG.

One final little thing to resolve is that Penrose-specific attributes _also_ get passed through to the SVG.  For instance, here's the output for a circle:

```
<circle fill="none" cx="98.82251439462043" cy="116.59791200601026" stroke="#000000" stroke-opacity="1" stroke-width="1.5" stroke-linecap="butt" r="62.11043606853752" ensureOnCanvas="false">
```

Notice in particular the `ensureOnCanvas` attribute, which is part of Penrose but not SVG.  Fortunately the browser ignores this attribute, but it seems like a problem just waiting to happen.  It also adds unnecessary size/bulk to the output files.

### Proposed solution

To get the ball rolling, let's just have a list of attributes that are Penrose-specific, and filter these out when constructing the SVG.  This is way (way) easier than keeping a list of attributes that are SVG-specific, and keeping this list in synch with various SVG versions.  In fact, the only two that come to mind are:

- `ensureOnCanvas`
- `name`

The latter (`name`) is especially insidious, since it sounds like something that _could_ be an SVG attribute—but isn't.  Instead, we already write this name within an [SVG `<title>` tag](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title), which is used for accessibility.  But the SVG attribute `name="..."` is meaningless.

cmumatt:
@keenancrane, thanks for observing this in the SVG output and opening this issue!

Agreed: the renderer should _not_ pass through the known (and non-SVG) internal properties `ensureOnCanvas` and `name`.  The original development included a mechanism to prevent this.  I'll take a look at what's going on, fix it in a PR, and add a test case to prevent future regression.

PR #1025, opened 2022-05-26
fix: issue #1024 exclude name, ensureOnCavas in SVG

cmumatt:
# Description

Fixes issue #1024 

This year we added svg-passthrough functionality, which allows unknown properties to be passed through to the SVG output.  Around this time, we also renamed most internal Style properties to correspond with the SVG names so most properties could flow right through the renderer into the SVG.  This approach eliminated code and make life easier for the Style developer who happens to already be familiar with SVG.

The tradeoff is that we sometimes have properties that are purely internal that we do not want to "leak" into the SVG.  This is implemented as an exclusion list in AttrHelper.ts.  Issue #1024 observes leakage of two internal-only properties, `name` and `ensureOnCanvas`.  This PR adds these two properties to the exclusion list to elide them from the SVG output.

Updated exclusion list in AttrHelper.ts.  Updated the rendered diagram set, which now lacks the `name` and `ensureOnCanvas` properties in the SVG output.  This diagram set is the current means by which we catch unintended changes to rendered properties.  

# Checklist

### DIFF
--- a/diagrams/allShapes-dashedShapes.svg
+++ b/diagrams/allShapes-dashedShapes.svg
@@ -1,5 +1,5 @@
 <svg version="1.2" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 800 700">
-  <g ensureOnCanvas="true">
+  <g>
     <marker
       id="E.icon-leftArrowhead"
       markerUnits="strokeWidth"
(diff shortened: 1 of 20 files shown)


samestep (approved):
Thanks for fixing this!

Issue #1023, opened 2022-05-26
Support typical SVG attributes (stroke-width) in Equations


maxkrieger:
as per @keenancrane's remark: https://github.com/penrose/penrose/blob/walk-on-spheres/packages/examples/src/walk-on-spheres/walk-on-spheres.sty#L336

MathJax labels are just ordinary `<path>`s so we can do some tree walking and apply attributes to all of them.

Just have to do `renderedLabel.getElementsByTagName("path").setAttribute("stroke-width", ...`.

But why are we setting stroke-width to 0 right now? https://github.com/penrose/penrose/blob/228746ee7544e4cf69c84f7bf871f0c9d95edcc5/packages/core/src/renderer/Equation.ts#L40

cmumatt:
Thanks for submitting this issue!  IIRC, the code snippet above pre-dates the renderer refactor and is probably due to Equation's lack of stroke properties.  I'll take a look at this one, too.

PR #1026, opened 2022-05-26
fix: Issue #1023 Allow stroke on Equations

cmumatt:
# Description

Resolves issue #1023  This PR:
- Adds stroke properties to the `Equation` shape and renders those properties onto any path elements within the label.  The pre-rendered label is re-scaled to be 1:1 with the canvas for proper rendering of properties such as `strokeWidth`.  
- Adds a white stroke and subscripts to `Equation` labels in the Walk-on-Spheres examples.  Updated the variation for nested WOS.
- Resolves several lint warnings and comment typos.  

This PR does not change the fact that labels are rendered prior to optimization, which limits optimizing some shape properties such as an `Equation`'s `fontSize` and `string`.

# Checklist


maxkrieger:
thanks Matt! Appreciate it

diagrams/circle-example-euclidean.svg
```diff
@@ -402,7 +402,7 @@
       viewBox="0 -750 429 1000"
     >
       <g
-        stroke="none"
+        stroke="currentColor"
```
samestep, line 405: @cmumatt Could you explain what this means?

cmumatt: Yes, I also had to look into this last night, also -- thanks for asking about it.

Since we are now setting stroke at the path level (where it should be), we are no longer overwriting this DOM default, which is set when we instantiate the <g> node.

samestep: OK! Thanks for clarifying.

keenancrane:
@cmumatt Thanks for the quick action on this!

Trying it out I notice a couple things.  First, something seems off with the units: at first I thought it wasn't working, because setting a stroke width of either `2.0` or `"2px"` gives an _extremely_ thin line.  For instance, in this example the black line has thickness 1.5:

image

Things don't show up unless I use a much bigger stroke width (here: 50):

image

In contrast, the stroke width for `Text` seems to correctly match the units used in the rest of the SVG. For instance, here's a `Text` with stroke width 2.0 on the same diagram:

image

I would guess that (perhaps) there is an SVG group `<g>` containing the MathJax path, which involves some kind of scaling transformation.

A separate issue is that for very thick strokes, things seem to be clipped to the text bounding box:

image

I suspect that, again, MathJax spits out the text paths in a group, and this group is clipped/masked using the bounding box.

Perhaps the answer to both of these issues is to:
- remove enclosing `<g>` tags (if this is indeed how things are being rendered to SVG…), and
- apply the scaling transformation directly to the control points in the SVG path

Alternatively, we could perhaps:
- keep the `<g>` tags/transformations as-is
- eliminate any directive(s) that cause clipping to occur
- inversely scale the stroke width, so that it appears the correct size after scaling

The former seems cleaner, but also more work (since we have to manipulate path data).

keenancrane:
Looking at the SVG output of Penrose, it looks like each expression rendered by MathJax is its own entire SVG, embedded within the larger SVG.  For instance, here's just one small piece of a SVG output by Penrose (below), which renders the TeX string `$m_{ki}`.

This SVG has a width/height associated with it, which in turn scales up/down the apparent size of the strokes around the paths.

Seems the "right" thing to do would be to strip away the `<svg>` tags, and absorb the transformations into the path itself.  Maybe there's a way to do this so that we don't have to change the path coordinates, but still get the right stroke width.  For instance, if we set the stroke attributes _**outside**_ the group that does the transformation, things may just work as expected.

TL;DR: Seems a good first stab at this issue is to
- get rid of the outer `<svg>` tags
- move the transformation from the SVG level to the inner level where the path is actually drawn

```
<svg xmlns="http://www.w3.org/2000/svg" width="18.345844000000003" height="11.66" role="img" focusable="false" viewBox="0 -750 1573.4 1000" style="vertical-align: -0.566ex;"><g stroke="currentColor" fill="#000000" stroke-width="0" transform="scale(1,-1)" fill-opacity="1"><g data-mml-node="math"><g data-mml-node="mtable"><g data-mml-node="mtr"><g data-mml-node="mtd"><g data-mml-node="msub"><g data-mml-node="mi"><path data-c="1D45A" d="M21 287Q22 293 24 303T36 341T56 388T88 425T132 442T175 435T205 417T221 395T229 376L231 369Q231 367 232 367L243 378Q303 442 384 442Q401 442 415 440T441 433T460 423T475 411T485 398T493 385T497 373T500 364T502 357L510 367Q573 442 659 442Q713 442 746 415T780 336Q780 285 742 178T704 50Q705 36 709 31T724 26Q752 26 776 56T815 138Q818 149 821 151T837 153Q857 153 857 145Q857 144 853 130Q845 101 831 73T785 17T716 -10Q669 -10 648 17T627 73Q627 92 663 193T700 345Q700 404 656 404H651Q565 404 506 303L499 291L466 157Q433 26 428 16Q415 -11 385 -11Q372 -11 364 -4T353 8T350 18Q350 29 384 161L420 307Q423 322 423 345Q423 404 379 404H374Q288 404 229 303L222 291L189 157Q156 26 151 16Q138 -11 108 -11Q95 -11 87 -5T76 7T74 17Q74 30 112 181Q151 335 151 342Q154 357 154 369Q154 405 129 405Q107 405 92 377T69 316T57 280Q55 278 41 278H27Q21 284 21 287Z" stroke="#ff6666" stroke-opacity="1" stroke-width="200px" stroke-linecap="butt"></path></g><g data-mml-node="TeXAtom" transform="translate(911,-150) scale(0.707)" data-mjx-texclass="ORD"><g data-mml-node="mi"><path data-c="1D458" d="M121 647Q121 657 125 670T137 683Q138 683 209 688T282 694Q294 694 294 686Q294 679 244 477Q194 279 194 272Q213 282 223 291Q247 309 292 354T362 415Q402 442 438 442Q468 442 485 423T503 369Q503 344 496 327T477 302T456 291T438 288Q418 288 406 299T394 328Q394 353 410 369T442 390L458 393Q446 405 434 405H430Q398 402 367 380T294 316T228 255Q230 254 243 252T267 246T293 238T320 224T342 206T359 180T365 147Q365 130 360 106T354 66Q354 26 381 26Q429 26 459 145Q461 153 479 153H483Q499 153 499 144Q499 139 496 130Q455 -11 378 -11Q333 -11 305 15T277 90Q277 108 280 121T283 145Q283 167 269 183T234 206T200 217T182 220H180Q168 178 159 139T145 81T136 44T129 20T122 7T111 -2Q98 -11 83 -11Q66 -11 57 -1T48 16Q48 26 85 176T158 471L195 616Q196 629 188 632T149 637H144Q134 637 131 637T124 640T121 647Z" stroke="#ff6666" stroke-opacity="1" stroke-width="200px" stroke-linecap="butt"></path></g><g data-mml-node="mi" transform="translate(521,0)"><path data-c="1D456" d="M184 600Q184 624 203 642T247 661Q265 661 277 649T290 619Q290 596 270 577T226 557Q211 557 198 567T184 600ZM21 287Q21 295 30 318T54 369T98 420T158 442Q197 442 223 419T250 357Q250 340 236 301T196 196T154 83Q149 61 149 51Q149 26 166 26Q175 26 185 29T208 43T235 78T260 137Q263 149 265 151T282 153Q302 153 302 143Q302 135 293 112T268 61T223 11T161 -11Q129 -11 102 10T74 74Q74 91 79 106T122 220Q160 321 166 341T173 380Q173 404 156 404H154Q124 404 99 371T61 287Q60 286 59 284T58 281T56 279T53 278T49 278T41 278H27Q21 284 21 287Z" stroke="#ff6666" stroke-opacity="1" stroke-width="200px" stroke-linecap="butt"></path></g></g></g></g></g></g></g></g></svg>
```

cmumatt:
So far no cigar getting strokeWidth 1:1 on the pre-rendered `Equation` label output using the suggested and other approaches.  The fundamental problem with the present solution in this PR is, as @keenancrane alluded, that we pre-render labels at one XXL size and subsequently scale them down for rendering on the canvas.  

This works great when stroke is not present, but it creates the visible disconnect of `strokeWidth` among `Equation` and the other shapes.  This would also impact some other features if we added those in the future.

I have a few ideas that I plan to explore.  In the meantime, moved the PR back to draft status.  Thanks for looking at it!

joshsunshine (changes requested):
The code looks good to me. Please change the subscripts in the examples, but otherwise I think we should merge it.

packages/core/src/utils/CollectLabels.ts
```diff
@@ -66,31 +69,98 @@ type Output = {
...
+    for (const path in paths) {
+      const thisPath = paths[path];
+      if (typeof thisPath === "object") {
+        const thisD = thisPath.getAttribute("d") || "";
+        thisPath.setAttribute(
+          "d",
+          svgpath(thisD)
+            .scale(scaledRatio) // re-scale the label
+            .abs() // use absolute coordinates (for BBox)
+            .unarc() // replace arcs with bezier curves (for BBox)
```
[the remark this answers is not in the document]
cmumatt: .scale() --> Scale the label to be 1:1 with the main canvas
.abs() --> Not strictly necessary.  Just avoids writing unnecessary code to deal with relative coordinates
.unarc() --> Not strictly necessary.  Just avoids writing unnecessary code to deal with arcs.

PR #1037, opened 2022-05-31
fix: Local image resolution not working in IDE

### LINKED ISSUE #1035: Local image resolution is not working in IDE
**Describe the bug**
Currently resolving `local` images is not working in the IDE. Local resolution in Automator works fine.  This primarily impacts users on https://penrose.cs.cmu.edu/try/ 

**To Reproduce**
1. Start the IDE.  
2. Click examples-->wos-laplace-estimator-walk-on-spheres
3. Add a blank line to any of the source files
4. Click compile.
5. Observe no images load

**Expected behavior**
Images should load

**Additional context**
The logic branch to load these images is currently marked "TODO" in `DiagramPanel` `pathResolver`.

cmumatt:
# Description

Resolves #1035.

This PR implements local serving of images and examples in the IDE.  This has two benefits:
1. It allows trios that are not examples to use the "stock" images within the set of examples, and
2. Local serving of examples and images eliminates the confusing situation where the local instance of Penrose (i.e., a local build or preview site) has different examples than penrose main.

During build, examples and images are deployed to vite's `public` directory ([documentation](https://vitejs.dev/guide/assets.html#the-public-directory)) and are served locally from there.  

# Open questions

- This PR does not implement the gist or upload functionality.  That will need to be implemented in a subsequent PR.
- Images are written to local storage, but we do not yet have a way to invalidate these resources at this time.


packages/examples/src/persistent-homology/persistent-homology.sty
```diff
@@ -142,21 +142,21 @@ forall Point p {
     center: p.circle1.center 
     width: p.circle1.r * 2.0
     height: p.circle1.r * 2.0
-    href: "shading.svg"
+    href: "persistent-homology-shading.svg"
```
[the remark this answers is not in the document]
cmumatt: Thanks for reviewing this.  As you noted below, once an example is modified, we lose the example `root` information and no longer know which folder contains the image resource.  If there are different, for example, `shading.svg`s in different folders, it is ambiguous which one to choose.  The opportunity for name collisions in this case seems equivalent no matter if we access images via HTTP or local storage.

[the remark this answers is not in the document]
wodeni: Actually there's no need to cache image from previous sessions. Within a Style program, images should be unique, so we can cache them when we are in the example phase, and just retrieve the images in local mode.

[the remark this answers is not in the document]
cmumatt: Thanks and a few clarifying questions since I am new to the new IDE.  How are we currently defining `session` and `workspace id`?  How is the `workspace id` accessed and when do we generate a new id?  In the described workflow, are we requiring a user to load an example with the stock images he or she wants prior to using them in their own diagram?  How do we want to invalidate the entries we create in local storage, particularly if they are workspace id specific -- whenever the `workspace id` changes?  Thanks for helping with the additional context!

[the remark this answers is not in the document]
wodeni: > Thanks and a few clarifying questions since I am new to the new IDE. How are we currently defining session and workspace id? How is the workspace id accessed and when do we generate a new id?

`WorkspaceMetadata` has the ID of the current workspace, which is roughly identified by a trio of Penrose programs. So whenever you load a new trio, `uuid()` is called to generate a new ID. You can simply search for all invocations for `uuid()` to see the call sites. The state management is done by recoil, so maybe look at https://recoiljs.org to see how it works.

> In the described workflow, are we requiring a user to load an example with the stock images he or she wants prior to using them in their own diagram?

Not sure if I understand the question. Image loading is done automatically if loading from gist, examples, or roger. Are you talking about the user uploading their own image? In that case, yeah I hope we can build that fairly easily (ie a simple file dialog can be a good starter). 

> How do we want to invalidate the entries we create in local storage, particularly if they are workspace id specific -- whenever the workspace id changes? 

Yeah, except when the workspaces are explicitly saved. You can get the list of these ids from `localFilesState`.

packages/editor/src/components/DiagramPanel.tsx
```diff
@@ -177,9 +210,16 @@ export default function DiagramPanel() {
...
       case "gist":
-      // TODO: cache images in local storage?
-      case "local":
         return undefined;
+      case "local": {
+        return fetchUrl(
+          relativePath,
+          new URL(
+            relativePath,
+            window.location.origin + window.location.pathname + "svg/"
```
wodeni, line 219: Instead of hard-coding the `svg/`  path, `local` workspace should only get the image from local storage because the image must came from gist, example, roger, or file upload (this is not implemented but doesn't change the logic)

packages/editor/.gitignore
```diff
@@ -8,6 +8,8 @@ pnpm-debug.log*
 lerna-debug.log*
 
 node_modules
+public/svg
```
wodeni, line 11: Okay to deploy the example source files, but svgs shouldn't be in the same folder anymore, so this shouldn't be needed.

packages/editor/package.json
```diff
@@ -6,9 +6,9 @@
     "start": "vite",
     "dev": "vite",
     "watch": "vite",
-    "clean": "rimraf dist",
+    "clean": "rimraf dist;rimraf public/svg;rimraf public/examples",
```
wodeni, line 9: No need to have `svg` folder.

wodeni: Also do `rimraf dist public/examples`

packages/editor/package.json
```diff
@@ -6,9 +6,9 @@
...
     "typecheck": "tsc --noEmit",
-    "build": "NODE_OPTIONS='--max-old-space-size=8192' vite build",
+    "build": "shx mkdir -p ./public/svg/;shx cp ../examples/src/*/*.svg ./public/svg;shx mkdir -p ./public/examples/;shx cp -r ../examples/src/* ./public/examples/;NODE_OPTIONS='--max-old-space-size=8192' vite build",
```
wodeni, line 11: Should remove the `svg` related steps

wodeni (changes requested):
Thanks for making the changes! Now that we are use local storage, I think the `svg` public folder is no longer necessary. My comments are mostly about removing that. Once that's done, we can undo the file renamings too.

packages/editor/package.json
cmumatt: Removed

packages/editor/package.json
cmumatt: Removed

packages/editor/.gitignore
cmumatt: Removed.

packages/editor/src/components/DiagramPanel.tsx
cmumatt: Removed.  This breaks the ability for example images to work for trios that are pasted in, but that doesn't sound like a user case we are supporting right now, which would make this unnecessary.

packages/examples/src/persistent-homology/persistent-homology.sty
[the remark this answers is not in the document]
cmumatt: Thanks, in the latest commit, local storage images are `workspace id` specific.  The editor was creating new `uuid`s when a workspace was saved, which caused saved workspaces to lose their images.  I removed the line of code creating the new uuid on workspace save.

It seems it would be better if the images were stored within the Recoil workspace state rather than outside of it.  For instance, when a saved workspace is deleted by the user, Recoil would also delete the images from local storage.  Right now these images accumulate each time a workspace with images is created and never go away. 

My suggestion is to move ahead with this PR since it fixes a bug in production and create an issue to move image resource management into Recoil state.

Issue #1044, opened 2022-06-05
Shape documentation pages are crashing


cmumatt:
**Describe the bug**
Currently the shape documentation pages are crashing with error: `This page has crashed`

**To Reproduce**
Example page: https://penrose.cs.cmu.edu/docs/ref/style/shapes/line/

**Expected behavior**
Shape details should be displayed.

PR #1045, opened 2022-06-05
fix: Docs-site Shape Property page errors and crashes

cmumatt:
# Description

Resolves issue #1044 plus others (see below).  

The `ShapeProps` React component in `docs-site` previously pulled shape data from `ShapeDef` in `@penrose/core` by creating shapes on a canvas and inspecting their properties at runtime.  With the Style compiler re-write, instantiating shapes requires passing a `Context` object.  As a JS React component, `ShapeProps` lacked type checking and this component was never updated to work with the new compiler.  This caused all shape documentation pages to crash in production.

This PR replaces the former JS React `ShapeProps` component with a TS component of the same name that imports a JSON file of shape properties calculated at build time rather than calling into `@penrose/core` to create and evaluate shapes at runtime.

This PR also fixes some additional issues that weren't previously logged as issues:
- `ColorV` values were inaccurately rendered on the page.
- `PointV` values were inaccurately rendered on the page.
- The `ShapeProps` component would hang the UI when first accessed if the `Demo` component were accessed prior.
- `Docs-site` build was extremely slow when both the `ShapeProps` and `Demo` components were active.
- The JS `ShapeProps` component lacked any type checking, which lead to issue #1044, and others.
- The `ShapeProps` component would not compile on the server side; previously a workaround was in place to force it to compile in the browser at runtime. This workaround is no longer necessary.

- This PR creates a new `automator` command, `shapedefs`, which outputs a JSON file of all shapes and their properties, including which properties are sampled and which properties are defaulted.  
- The `ShapeProps` component imports this data during build to render the shape properties pages server side. 
- This eliminates the need to load `@penrose/core` and instantiate shapes dynamically at runtime within `ShapeProps`. 
- The JS `ShapeProps` component is converted to TypeScript to allow type-checking.  Type checking would have prevented the original issue #1044 and others.

# Checklist

# Open questions

- None


packages/automator/index.tsx
```diff
@@ -357,6 +362,57 @@ const batchProcess = async (
...
+        outThisShapeDef.defaulted[propName] = shapeSample1[propName];
+      } else {
+        outThisShapeDef.sampled[propName] = shapeSample1[propName];
+      }
+    }
+  }
+
+  // Write the shape definition output
+  if (outFile === undefined) {
+    console.log(JSON.stringify(outShapes, null, 2));
```
joshsunshine, line 406: I assume this is debug code. Consider removing.

joshsunshine (approved):

joshsunshine:
Looks good to me. I think it's ready to merge.

PR #1060, opened 2022-07-06
feat: Numeric fuzzer

cmumatt:
# Description

This PR is a work in progress that includes a prototype numeric fuzzer intended to help find bugs in Penrose computation, constraint, and other functions. The fuzzer is an in-progress research project with @rhit-chois3.  In its current state, the fuzzer uses an implicit oracle to check for the presence of errors such as NaN, Infinity, etc. 

Areas of further and ongoing work:
- Auto-configuring the function to fuzz.  Currently functions are manually configured, e.g., in `NumFuzzConfig.ts`.  We  included in this PR a lightweight TS analyzer that we may use to extract the properties we need to auto-configure the function with the exception of min/max and any non-varying values we want to pass.  
- Calling the function to get its result.  Due to the variety of function signatures, we currently require this call to be coded by a human, but we may escape hatch to eval() or something similar to automate this step if we can't think of a better idea.
- Reading and presenting the output, which is not currently in human-friendly form.  We plan to pipe the results to JSON and then to an HTML/JS/CSS report viewer (TBD) we can run locally and perhaps integrate into a VS Code plugin to present the fuzzer results to the programmer.
- The current version simply randomizes inputs within a range that is specified by argument.  A next step here is to add an option to provide a custom input sampler in the fuzzer options and to provide a few "canned" samplers.

# Checklist

- [ ] I have reviewed any generated changes to the `diagrams/` folder


packages/core/src/utils/NumFuzz.ts
```diff
@@ -0,0 +1,417 @@
...
+/**
+ * Using a set of provided FuzzOptions, sets up a fuzzing environment
+ * and returns it to the client.  Before calling fuzz(), call this function
+ * to set up the fuzzing environment.
+ *
+ * @param fo Describes the fuzzing environment to setup
+ * @returns Fuzzing environment
+ */
+export const fuzzSetup = (fo: FuzzOptions): FuzzEnv => {
+  const context = simpleContext(fo.variation || "");
```
samestep, line 35:
```suggestion
  const context = simpleContext(fo.variation ?? "");
```

packages/core/src/utils/NumFuzz.ts
```diff
@@ -0,0 +1,417 @@
...
+  passed: boolean; // true if output matches oracle; false, otherwise
+};
+
+/**
+ * Fuzzing options for a given computation function
+ */
+export type FuzzOptions = {
+  fnCfg: FnDef; // Function configuration
+  variation?: string; // Variation / seed (optional)
+  graphFn: typeof primaryGraph; // Function for generating the computation graph. TODO: Should also accept secondaryGraph
```
samestep, line 392: I haven't read all the code in this file, but I'd guess it's best to just use `makeGraph` directly here, rather than case-splitting on `primaryGraph` vs `secondaryGraph`.

Issue #744, opened 2022-01-03
Add a `symmetric` and other keywords to Domain predicates


keenancrane:
**Is your feature request related to a problem? Please describe.**

Mathematical relations, such as [predicates](https://en.wikipedia.org/wiki/Relation_(mathematics)), often have important properties such as reflexivity, transitivity, symmetry, etc.  Yet Domain schemas currently support only a generic `predicate` type.  The lack of such properties can make it difficult or impossible to implement some common pattern matches, especially if one wishes to avoid redundant code.

As a real motivating example, consider the `structural-diagrams` example, which includes types
```
type Atom
type Hydrogen <: Atom
type Oxygen <: Atom
predicate Bond( Atom a1, Atom a2 )
```

In this domain, we can construct a water molecule in a variety of ways, e.g.,
```
Hydrogen H1, H2
Oxygen O
Bond( O, H1 )
Bond( O, H2 )
```
or
```
Hydrogen H1, H2
Oxygen O
Bond( H1, O )
Bond( O, H2 )
```
etc.

Suppose that in Style we now want to enforce the constraint that the bond angle for a water molecule is 104.5 degrees.  Our first attempt is
```
forall Oxygen o; Hydrogen h1; Hydrogen h2
where Bond(o,h1); Bond(o,h2) {
   -- enforce the bond angle
}
```
However, this pattern will match only the first version of our code, not the second.  Currently the only solution is to duplicate the code:
```
forall Oxygen o; Hydrogen h1; Hydrogen h2
where Bond(o,h1); Bond(o,h2) {
   -- enforce the bond angle
}
where Bond(h1,o); Bond(o,h2) {
   -- enforce the bond angle
}
```
Of course, even this modified code doesn't cover all cases.

**Describe the solution you'd like**
One possibility is to allow a Domain `predicate` to be tagged with additional properties.  For instance, we might write
```
symmetric predicate Bond(Atom a1, Atom a2)
```
The semantics here are that `Bond` holds for either ordering of the arguments.  More generally, `symmetric` would most naturally indicate that the order of arguments is irrelevant, for any number of arguments.

Another natural property (though perhaps more difficult to handle) is `transitive`.  For instance, consider the predicate
```
transitive predicate P(T x, T y)
```
and the associated Substance program
```
T x, y, z
P( x, y )
P( y, z )
```
In this case, the following Style pattern
```
forall T u; T v
where P(u,v)
```
[...]

PR #1061, opened 2022-07-07
feat: basic symmetric predicates

liangyiliang:
# Description

Related issue/PR: #744 

The Penrose languages now support basic "symmetric" predicates. If a predicate `MyPred` is declared `symmetric`, then within the Style program, `MyPred(a, b)` also matches `MyPred(b, a)` automatically. We currently only support binary predicates.

* We modified the Domain parser to parse `symmetric predicate` correctly.
* We modified the Domain compiler to enforce some conditions for symmetric predicates: all symmetric predicates can only have two arguments, and they must have the same type.
* We modified the Style compiler to take symmetric predicates into consideration when matching predicates.
* We added an example that corresponds to the Water-bond example in issue #744.

# Checklist


packages/core/src/parser/Domain.ne
```diff
@@ -82,11 +83,15 @@ type_decl -> "type" __ identifier (_ "(" _ type_params _ ")"):? (_ "<:" _ sepBy1
   }
 %}
 
-predicate -> "predicate" __ identifier type_params_list args_list {%
-  ([kw, , name, params, args]): PredicateDecl<C> => ({
+# This now works with symmetric predicate declarations.
+predicate -> ("symmetric" __):? "predicate" __ identifier type_params_list args_list {%
+  ([sym, kw, , name, params, args]): PredicateDecl<C> => ({
     ...nodeData,
     ...rangeFrom([rangeOf(kw), ...args, ...params]),
```
wodeni, line 90: Minor point: if qualifiers like `symmetric` come before `predicate`, the starting source position will need to change. Now that I think about it, I kind of like `predicate symmetric` a bit more? Not sure if the concrete syntax was discussed at all

wodeni: Yep that seems right!

liangyiliang: I just pushed some new code if you want to take a look!

joshsunshine:
It looks great overall. Can you edit the style program to make the Hydrogens closer to the Oxygen molecules in the example?

liangyiliang:
Sure I can do that - hold on

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

liangyiliang:
Just fixed the diagram ...

packages/examples/src/symmetric-bond/symmetric-bond.sty
```diff
@@ -0,0 +1,55 @@
...
+    v1 = o.bond1.end - o.bond1.start
+    v2 = o.bond2.end - o.bond2.start
+
+    o.layering1 = h1.icon above o.bond1
+    o.layering2 = h2.icon above o.bond2
+    o.layering3 = o.icon above o.bond1
+    
+    ensure equal(norm(v1), norm(v2))
+    ensure equal(angleBetween(v1, v2), toRadians(104.5))
+    ensure equal(norm(v1), 100)
```
joshsunshine, line 54: This seems reasonable and the diagram looks better.

PR #1063, opened 2022-07-14
fix: multiple matching

### LINKED ISSUE #1062: Multiple Matching Issue
**Describe the bug**
Sometimes, the same things in substance can be matched multiple times by the same style rule, due to the style matcher accepting multiple different variable mappings.

**To Reproduce**
Here is the program trio to reproduce the issue, on the current newest version on the `main` branch: https://fake3dlinalg.penrose-72l.pages.dev/try/?gist=fd9cf13ae1b6faebedc2bab08bb23ba4.

For convenience, the code is as follows:
```
-- Substance
Hydrogen H1, H2
Oxygen O
Bond( O, H1 )
Bond( O, H2 )

-- Style
canvas {
    width = 50
    height = 40
}

forall Oxygen o; Hydrogen h1; Hydrogen h2
where Bond(o,h1); Bond(o,h2) {
    myText = Text {
        string: "Water!"
        fillColor: rgba(0, 0, 0, 255)
    }
}

-- Domain
type Atom
type Hydrogen <: Atom
type Oxygen <: Atom
predicate Bond( Atom a1, Atom a2 )
```

**Expected behavior**
The style program is supposed to detect water molecules with `forall Oxygen o; Hydrogen h1; Hydrogen h2 where Bond(o,h1); Bond(o,h2)`. Whenever it sees a water molecule, it should print, in giant fonts, "Water" on the canvas. The substance program only has one water molecule, but this style rule is invoked twice, and "Water" is printed twice on the canvas.

**Screenshots**
Here is what it currently shows on the canvas:
image

**Speculations about Bug**
Here are some speculations about what's happening. The style compiler generates two possible sets of variable mappings for `forall Oxygen o; Hydrogen h1; Hydrogen h2`:
```
Mapping 1:
  StyVar -> SubVar
  o      -> O
  h1     -> H1
  h2     -> H2

Mapping 2:
  StyVar -> SubVar
  o      -> O
  h1     -> H2
  h2     -> H1
```
It turns out that both mappings satisfy the relational declaration `where Bond(o,h1); Bond(o,h2)`, so the rule is invoked twice.

**Potential Solution**
Each block in the Style program maintains a set of Substance declarations that it matched upon, which we call `MatchedDeclarations`. For instance, for the example provided, we initially start with an empty set:
```
MatchedDeclarations = {}
```

Suppose we now match upon `Mapping 1`. This will add an element into the set. The new element is itself a set: `{Bond(O, H1), Bond(O, H2)}`. Then,
```
MatchedDeclarations = {
  {
    Bond(O, H1), 
    Bond(O, H2)
  }
}
```
This will trigger the rule and print one "Water" on the canvas.

Then, if we match upon `Mapping 2`, this will generate a set `{Bond(O, H2), Bond(O, H1)}`, which is already a member of `MatchedDeclarations`.  This utilizes the fact that
```
{Bond(O, H1), Bond(O, H2)} === {Bond(O, H2), Bond(O, H1)}
```
since sets are orderless. Since such a mapping will not trigger the rule, since the _same_ thing has already matched the rule in the previous step.

**Thoughts**
The most important question, however, is whether or not this really is a _bug_. Is it the expected behavior (according to our users) to print two "Water" on the canvas, or do the users expect only one?

liangyiliang:
# Description

Related issue/PR: #1062 AND #1064 

This PR fixes #1062 and #1064 using the proposed resolution as written in #1064.

See comments of #1064 

```
-- Substance
Hydrogen H1, H2
Oxygen O
Bond( O, H1 )
Bond( O, H2 )

-- Style
canvas {
    width = 50
    height = 40
}

forall Oxygen o; Hydrogen h1; Hydrogen h2
where Bond(o,h1); Bond(o,h2) {
    myText = Text {
        string: "Water!"
        fillColor: rgba(0, 0, 0, 255)
    }
}

-- Domain
type Atom
type Hydrogen <: Atom
type Oxygen <: Atom
predicate Bond( Atom a1, Atom a2 )
```

The current version prints two "Water!" strings on the canvas, even though the Substance program only has one water molecule. After this fix, it only prints one "Water!" as expected.

# Checklist


samestep (approved):
This looks great!

liangyiliang:
Turns out there are some pending questions that need to be resolved - I'm not merging the code yet.

samestep:
For `diagrams/non-convex-non-convex.svg`, why does it look like a couple of the little triangles either changed shape or merged together?

samestep:
For `diagrams/small-graph-disjoint-rect-line-horiz.svg`, why are a bunch of the rectangles disjoint in the old diagram but not the new one?

packages/core/src/compiler/Style.ts
```diff
@@ -1072,9 +1119,42 @@ const filterRels = (
...
+    ([currMatches, currSubsts], subst) => {
+      const matchedRels = matchAllRels(
+        typeEnv,
+        subEnv,
+        subProgFiltered,
+        substituteRels(subst, rels)
+      );
+      const substTargets = im.Set<string>(Object.values(subst));
+      if (matchedRels !== undefined) {
+        const record: im.Record<MatchesObject> = im.Record({
```
samestep, line 1140: Nice, I didn’t know about [`Record`](https://immutable-js.com/docs/v4.1.0/Record/) before!

samestep (approved):
Looks great! Again I’d like to get @wodeni’s approval on this before merging.

liangyiliang:
> For `diagrams/small-graph-disjoint-rect-line-horiz.svg`, why are a bunch of the rectangles disjoint in the old diagram but not the new one?

Looks like that is a local minimum - after resampling, they are disjoint

liangyiliang:
> For `diagrams/non-convex-non-convex.svg`, why does it look like a couple of the little triangles either changed shape or merged together?

They are not merged together - they are still disjoint. The Style block for `Triangle` has
```
    -- Try getting close to the center
    encourage equal(0.0, x.a[0])
    encourage equal(0.0, x.a[1])
```
so they try to clump together. This shows how the objective function changed - it is now assigning less weight to things like `Triangle x; Triangle y`, and more weight to a single `Triangle x`!

PR #1066, opened 2022-07-19
feat: Predicate aliasing

liangyiliang:
# Description

Related issue/PR: #623 

This merges the changes in #623 into the current main branch, while making sure things play well with symmetry and other matching semantics. We added some tests, and updated the _water_ example in the registry to utilize predicate aliasing.

# Motivation
Suppose
```
-- dsl
type Thing
predicate Relation(Thing, Thing)
predicate SpecialRelation(Thing, Thing)
```
There are two types of relations: `Relation` and `SpecialRelation`. If `Relation(t1, t2)` holds, we want to draw a generic, uninteresting line between `t1` and `t2`. However, if both `Relation(t1, t2)` AND `SpecialRelation(t1, t2)` holds, we want to replace that line in `Relation(t1, t2)` with a better, _special_ line.

Our style program would start with
```
forall Thing t1; Thing t2
where Relation(t1, t2) {
  theLine = Line { ... }
}
```
How would we handle `SpeicialRelation(t1, t2)`? We further write,
```
forall Thing t1; Thing t2
where Relation(t1, t2); SpecialRelation(t1, t2) {
  // oops, we do not have access to `theLine` in the previous block!
  // so we cannot modify it to look _special_
}
```
which is problematic. We could associate `theLine` with `t1.theLine`, but it doesn't really make sense, since `theLine` does not belong to `t1` nor `t2` - instead, it belongs to the relation.

We can, however, introduce _predicate aliasing_ in the Style program.
```
forall Thing t1; Thing t2
where Relation(t1, t2) as r1 {
  r1.theLine = Line { ... }
}

forall Thing t1; Thing t2
where Relation(t1, t2) as r1; SpecialRelation(t1, t2) {
  override r1.theLine.xxx = xxx
  // Note that we have access to r1.theLine
}
```
In this case, the predicate `Relation(t1, t2)` has been aliased with `r1`, allowing `r1.theLine` to be used across multiple style blocks.

As a concrete example, consider
```
Thing T1, T2
Relation(T1, T2)
SpecialRelation(T1, T2)
```
In the first style block, shape `Relation(T1, T2).theLine` is registered. In the second style block, when it refers to `r1.theLine`, it points to the previously-registered `Relation(T1, T2).theLine`, allowing the second style block to use things defined under `Relation(T1, T2)` by the first style block, hence allowing `Relation(T1, T2)` to be used across multiple style blocks.

This gives greater flexibility to the Style writers.

Suppose we have
```
-- sub
Hydrogen H
Oxygen O
Bond(H, O)

-- sty, assuming Bond is symmetric
forall Hydrogen h; Oxygen o
where Bond(o, h) as b {
  // ...
}
```

1. We modified the Style parser to support `Bond(o, h) as b`
2. The current matching semantics generates the substitution `h -> H; o -> O`
3. We then augment the substitution to be
```
h -> H
o -> O
b -> Bond_H_O
```
where `Bond_H_O` represents the `Bond(H, O)` in the substance program. Note that we always use the version of the predicate in the Substance program.

4. Within this style block, `b -> Bond_H_O` is available to use. If, in another style block, substance `Bond(H, O)` is matched again and bounded to another alias `b2`, then `b2` refers to the same `Bond_H_O` in the translation map.

See https://ee249bd2.penrose-72l.pages.dev/try/?gist=610d61025fee9b822cf35d76c59e04f5
which is the example in the registry.

# Checklist


joshsunshine:
@liangyiliang can you add some of the motivation from the Helena's earlier PR into this one?

packages/examples/src/registry.json
```diff
@@ -168,12 +174,6 @@
       "style": "fake-3d-linear-algebra",
       "domain": "fake-3d-linear-algebra",
       "variation": "ConceptualMeerkat694"
-    },
-    {
-      "substance": "water-molecule",
-      "style": "symmetric-bond",
-      "domain": "symmetric-bond",
-      "variation": "SharkskinWalrus24396"
```
joshsunshine, line 176: I think this simple example is nice too. Please add it back @liangyiliang

joshsunshine (approved):

Issue #1064, opened 2022-07-15
Multiple Matching, continued


liangyiliang:
Hello! This issue continues upon some of the discussions in #1062. This hopefully gives insights on the implementation details to fix the #1062 issue.
## Program 1
Consider the following program trio:
```
-- DSL
type Atom

-- SUB
Atom A1
Atom A2

-- STY, canvas preamble omitted
forall Atom a1; Atom a2 {
  myText = Text {
    string : "Hello"
  }
}
```
In this program, how many "Hello" strings should be printed onto the canvas? The current implementation prints two, because the selector `forall Atom a1; Atom a2` admits two mappings:
```
MAPPING 1
a1 --> A1
a2 --> A2

MAPPING 2
a1 --> A2
a2 --> A1
```
There are no relational constraints, so both mappings invoke the style rule. However, I think quite naturally, it should print `Hello` out once, since there really is only one pair of `Atom` objects. _**Does this make sense? Are there cases where having this rule only invoked once causes problems?**_

## Program 2
Now consider this program:
```
-- Substance
Atom A1, A2
Bond( A1, A2 )
Atom A3, A4

-- Style, canvas preamble omitted
forall Atom a1; Atom a2; Atom a3            -- Note the additional a4
where Bond(a1, a2) {
    myText = Text {
        string: "Hello"
    }
}

-- Domain
type Atom
predicate Bond( Atom, Atom )
```
How many times should `Hello` be printed, in this case? _I think it should be printed twice_, with the following mappings being accepted:
```
MAPPING 1
a1 --> A1
a2 --> A2
a3 --> A3

MAPPING 2
a1 --> A1
a2 --> A2
a3 --> A4
```
Note that in the style block, `a3` does not appear in the relational constraints, so `a3` can be mapped to anything, `A3` or `A4`. These mappings result in different things being mapped, not just different order of things, so I think it should print twice.
The current implementation of the code (in PR #1063), however, prints it only once. _Should printing twice be the desired behavior?_

Thanks!

liangyiliang:
To ensure default set-matching, we have
* The first program should print "Hello" once.
* The second program should print "Hello" twice.

To do this, the proposed solution in #1062 is amended, as follows.

Each style block maintains a set of tuples, each tuple storing
* A set of _substitution targets_, and
* A set of _matched relations_.
I shall explain both.
## Premise
Suppose we have Style program
```
forall Atom a; Atom b; Atom c; Atom d
[...]

PR #1072, opened 2022-07-26
feat: faster matching

liangyiliang:
# Description

Related issue/PR: #566 

We updated the style-substance matching semantics so that there are less substitutions to consider.

# Writeup

[faster-matching.pdf](https://github.com/penrose/penrose/files/9214420/faster-matching.pdf)
One thing to add to my writeup. Say I am looking at the Style relation `Bond(a, b)`, and, by looking at the Substance `Bond(A, B), Bond(B, C), Bond(C, D)`, I decided that
`(a, b) ==> { (A, B), (B, C), (C, D) }`.
The actual information that I store is:
```
(a, b) ==> {
  ((A, B), { Bond(A, B) })
  ((B, C), { Bond(B, C) })
  ((C, D), { Bond(C, D) })
}
```
In particular, for each substitution `(a, b) ==> (A, B)` , I also store a copy of the Substance relation that this Style relation matches.

Then, when merging two potential substitution lists, I still perform the Cartesian product, where I
merge together the two `Substs`, and merge the corresponding Substance relations.

This information allows us to _deduplicate_ our substitutions. The _deduplication_ steps are described [here](https://github.com/penrose/penrose/issues/1064#issuecomment-1189228125).

# Checklist


packages/core/src/compiler/Style.ts
```diff
@@ -1370,38 +1171,457 @@ const matchDeclLine = (
...
+  const first = listPSubsts.first();
+  if (first) {
+    // first_ converts List<[Subst, SubStmt]> into List<[Subst, Set<SubStmt>]>
+    const first_: im.List<
+      [Subst, im.Set<SubStmt<A> | undefined>]
+    > = first.map(([subst, stmt]) => [
+      subst,
+      im.Set<SubStmt<A> | undefined>().add(stmt),
+    ]);
+    const substs = listPSubsts.shift().reduce((currSubsts, pSubsts) => {
```
samestep, line 1610: I know `shift` is linear-time for JS arrays; is it the same for `im.List`, or is it faster (constant or logarithmic time)?

packages/core/src/compiler/Style.ts
```diff
@@ -1286,23 +1062,47 @@ const combine = (s1: Subst, s2: Subst): Subst => {
 // TODO check for duplicate keys (and vals)
 // (x) operator combines two lists of substitutions: [subst] -> [subst] -> [subst]
 // the way merge is used, I think each subst in the second argument only contains one mapping
```
samestep, line 1064: Does this `TODO` still apply?

samestep (approved):
Looks great, thanks so much for doing this! Really clean approach, and nice use of Immutable `Set`s for clarity in the code.

I left a couple comments for things I was curious about. One other question: why exactly did some of the `diagrams/` change? I'm guessing they all have the same set of substitutions for their selectors; is the difference just that now some of those substitutions are in a different order? Is there a way to verify that this is the case?

packages/core/src/compiler/Style.ts
liangyiliang: Duplication of keys should be prohibited by Subst inherently because it is a JS/TS object. Was there some other implementation of Subst that allowed this?

Duplication of values is checked by another function

packages/core/src/compiler/Style.ts
liangyiliang: I'm not sure whether or not it's linear time. But it shouldn't have a big effect, given that this will be run only once for each block, and the list it operates on has size O(num of Stu Relations + num of Sty Variables).

samestep: Makes sense!

packages/core/src/compiler/Style.ts
samestep: Not sure ¯\_(ツ)_/¯ but in any case, if this no longer applies then it's probably best to remove the comment?

liangyiliang: Sounds good I'll fix that.

joshsunshine (approved):

Issue #785, opened 2022-01-07
Expose match index and total number of matches per rule in Style


keenancrane:
## Problem

Many diagrams are most naturally expressed in terms of ordinal and/or cardinal numbers.

_Ordinal._ For instance, in this diagram we would like to use an enumeration of marked angles to determine the number of tick marks:

<img src="https://user-images.githubusercontent.com/3604525/148614549-ab0acfd5-92e2-4f46-a5d7-51817c3532eb.png" width="300px">

_Cardinal._ In this diagram, we would like to use the total number of tick marks to determine the spacing between ticks:

<img src="https://user-images.githubusercontent.com/3604525/148614693-2fcfd4fe-f1aa-4c50-b659-bb5285163c86.png" width="400px">

However, the current design of Style makes it hard to specify these kinds of diagrams: even though a given rule may get invoked several times, there is currently no way to determine
- the ordinal for the current match, nor
- the total number of matches for that rule.

## Proposed solution

The proposal is to provide access to the ordinal and match count for each Style rule.  In particular, within a rule block we will allow the programmer to access the integer constants

- `match_id` — gives the (1-based) ordinal of the current match, among all matches for the current block.
- `match_count` — gives the total number of matches for this rule.

For instance, consider the Substance program

```
Point i, j, k
```

and the Style program

```
forall Point p {
   shape p.labelText = Text {
      string: toString(match_id) + "/" + toString(match_count)
      center: (?,?)
      -- ...more attributes...
   }
   shape p.labelWidget = Circle {
      center: p.labelText.center
      -- ...more attributes...
   }
}
```

(Note: I'm aware we don't have string conversion or concatenation syntax... this is just for demonstration purposes. :-))

Together, these programs would produce a drawing like this one:
<img src="https://user-images.githubusercontent.com/3604525/148615962-152c8911-39d3-4f57-92ff-df0affea456a.png" width="300px">

It seems reasonable that attempting to override these values, e.g.,

```
override match_id = 5
```

should produce an error.

### Implementation

According to @wodeni we already keep track of similar information internally, i.e., an ID for each match, and the total number of matches.  Sounds like it may need to be broken down per-rule though.

liangyiliang:
This sounds really interesting and reasonable! Informations such as `match_count` and `match_id` are what we can attach to our `Subst[]` object and individual `Subst` objects.

[...]

PR #1074, opened 2022-08-09
feat: match metadata

liangyiliang:
# Description

Related issue/PR: #785 

We expose some metadata about a specific _substitution_ (or matching) to the interior of a style block. In particular, two metadata are available:
* `match_id` is the 1-indexed ordinal of the specific matching, and
* `match_total` is the total number of matches.

For example:
```
-- dsl
type MyType

-- sub
MyType T1, T2, T3, T4

-- sty, canvas specifications omitted
forall MyType t {
  // can use `match_id` and `match_total` here.
}
```
This _Style_ block is matched four times, so `match_total` equals to `4` within this style block. Furthermore, `match_id` can be `1`, `2`, `3`, or `4` depending on which matching we are processing.

The current algorithm to process a Style block is:
```
for i = 1 .. N {
  processBlock(subst[i], blockStatements)
}
```
where `N` is the total number of substitutions, and `blockStatements` is the list of statements within this Style block.

The new algorithm is:
```
for i = 1 .. N {
  blockStatements.addInFront("match_id = " + i)
  blockStatements.addInFront("match_total = " + N)
  processBlock(subst[i], blockStatements)
}
```

Essentially, we add two fake AST nodes at the beginning of the block statements, assigning the metadata variables to their correct values, depending on the substitution.

See: [example](https://3dd0308b.penrose-72l.pages.dev/try/?gist=e07d1ec3de4be8b5577fc1b568479ad8)

# Checklist


packages/core/src/compiler/Style.ts
```diff
@@ -1790,15 +1791,32 @@ const resolveLhsName = (
...
 ): Result<ResolvedPath<C>, StyleError> => {
   const { start, end, name, members, indices } = path;
-  return indices.length > 0
-    ? err({ tag: "AssignAccessError", path })
-    : ok({ start, end, ...resolveLhsName(block, assignment, name), members });
+  if (indices.length > 0) {
+    return err({ tag: "AssignAccessError", path });
+  } else {
+    const resolvedLhsName = resolveLhsName(block, assignment, name);
+    /*if (reservedVariableNames.includes(toString(name))) {
```
wodeni, line 1806: Is this useful?

wodeni: If not, you can revert to the ternary right?

packages/core/src/compiler/Style.ts
```diff
@@ -1872,6 +1890,45 @@ const blockId = (
   }
 };
 
+const makeFakeIntPathAssign = (name: string, value: number): PathAssign<C> => {
```
wodeni, line 1893: TODO: need an API for synthetic node construction

packages/core/src/types/errors.ts
```diff
@@ -134,6 +134,11 @@ export interface DeconstructNonconstructor {
   deconstructor: Deconstructor<A>;
 }
 
+export interface ReadonlyVariableMutationError {
+  tag: "ReadonlyVariableMutationError";
```
wodeni, line 138: Is this in use?

wodeni (commented):
Some minor comments below

packages/core/src/compiler/Style.ts
liangyiliang: Yup I can do that

packages/core/src/types/errors.ts
liangyiliang: Removing that

PR #1075, opened 2022-08-14
feat: allow anonymous shapes in Style rules

### LINKED ISSUE #985: Allow anonymous shapes in Style rules
The Style language currently requires all shapes to be given an explicit name, even if these shapes never need to be referenced in subsequent expressions or rules.  For instance:

```
forall Node x {
   shape x.dotShape = Circle {
      center: (?,?)
      r: 10.0
      fillColor: Colors.black
   }
}
```

These "dummy names" (in this case, `x.dotShape`) make code less legible, and increase cognitive load.  It also forces the programmer to spend extra effort picking a dummy name that doesn't conflict with existing names, even though this name will never be used.

A simple solution is to allow anonymous names, i.e., allow declarations like

```
forall Node x {
   shape Circle {
      center: (?,?)
      r: 10.0
      fillColor: Colors.black
   }
}
```

(Of course, if this shape does need to be referenced in later iterations of the program, it's easy enough to give it an explicit name.)

**Additional context**
This design nicely parallels the design we already have for constraints and objectives, which are allowed to be anonymous.  In fact, it sort of violates the principle of least surprise that constraints/objectives can be anonymous, but shapes cannot.

I would guess that anonymous shapes can be implemented exactly as we've done for anonymous constraints/objectives: automatically generate a name for each anonymous shape that only gets referenced internally within the compiler (but is never exposed to the user).

ZiyaoWei:
# Description

Fixes #985 

https://penrose.cs.cmu.edu/try/?gist=80a8c2a48fc1cdf67e01be918cfba404 works locally

# Checklist

### DIFF
--- a/packages/core/src/parser/Style.ne
+++ b/packages/core/src/parser/Style.ne
@@ -403,6 +403,7 @@ anonymous_expr
   -> layering {% id %}
   |  objective {% id %}
   |  constraint {% id %}
+  |  gpi_decl {% id %}
 
 # NOTE: inline computations on expr_literal (including expr_literal)
 expr -> arithmeticExpr {% id %}


samestep (approved):
Looks perfect, thank you! Also, if you want to open a followup PR editing the existing diagrams in our registry (the ones @keenancrane was referring to in the original issue) to make use of this new syntax, that would definitely be appreciated :)

PR #1080, opened 2022-09-05
feat: Make parse errors more readable

### LINKED ISSUE #1052: Improve parsing error messages
Last month @wodeni and @JonathanAldrich and I had a Slack conversation about #961, lamenting how errors from the parser can be "quite cryptic", causing us to make the parser less strict than it could (should?) be and instead shift some of that error-checking into the compiler. I don't know if there's a good general solution to this issue, but while showing Penrose to someone earlier this week, I realized that part of the problem is simply information overload. For instance, consider changing `=` to `==` after `x.icon` in the `tree-venn` example:
```
Error: Syntax error at line 7 col 10:

    x.icon ==
           ^
Unexpected eq token: "==". Instead, I was expecting to see one of the following:

A ws token based on:
    __$ebnf$1 → __$ebnf$1 ● %ws
    __ →  ● __$ebnf$1
    layering → layering$ebnf$2 path ● __ "above" __ path
    anonymous_expr →  ● layering
    statement →  ● anonymous_expr
    statements → _ ● statement _c_ nl statements
    statements → _c_ nl ● statements
    block → "{" ● statements "}"
    header_block → header ● block _ml
    header_blocks$ebnf$1 → header_blocks$ebnf$1 ● header_block
    header_blocks →  ● header_blocks$ebnf$1
    input → _ml ● header_blocks
A ws token based on:
    _$ebnf$1 → _$ebnf$1 ● %ws
    _ →  ● _$ebnf$1
    path$ebnf$2$subexpression$1 →  ● _ access_ops
    path$ebnf$2 →  ● path$ebnf$2$subexpression$1
    path → binding_form path$ebnf$1 ● path$ebnf$2
    path_assign → path_assign$ebnf$1 __ ● path _ "=" _ assign_expr
    statement →  ● path_assign
    statements → _ ● statement _c_ nl statements
    statements → _c_ nl ● statements
    block → "{" ● statements "}"
    header_block → header ● block _ml
    header_blocks$ebnf$1 → header_blocks$ebnf$1 ● header_block
    header_blocks →  ● header_blocks$ebnf$1
    input → _ml ● header_blocks
A "below" based on:
    layering → layering$ebnf$1 path __ ● "below" __ path
    anonymous_expr →  ● layering
    statement →  ● anonymous_expr
    statements → _ ● statement _c_ nl statements
    statements → _c_ nl ● statements
    block → "{" ● statements "}"
    header_block → header ● block _ml
    header_blocks$ebnf$1 → header_blocks$ebnf$1 ● header_block
    header_blocks →  ● header_blocks$ebnf$1
    input → _ml ● header_blocks
A "above" based on:
    layering → layering$ebnf$2 path __ ● "above" __ path
    anonymous_expr →  ● layering
    statement →  ● anonymous_expr
    statements → _ ● statement _c_ nl statements
    statements → _c_ nl ● statements
    block → "{" ● statements "}"
    header_block → header ● block _ml
    header_blocks$ebnf$1 → header_blocks$ebnf$1 ● header_block
    header_blocks →  ● header_blocks$ebnf$1
    input → _ml ● header_blocks
A "=" based on:
    path_assign → path_assign$ebnf$1 __ path _ ● "=" _ assign_expr
    statement →  ● path_assign
    statements → _ ● statement _c_ nl statements
    statements → _c_ nl ● statements
    block → "{" ● statements "}"
    header_block → header ● block _ml
    header_blocks$ebnf$1 → header_blocks$ebnf$1 ● header_block
    header_blocks →  ● header_blocks$ebnf$1
    input → _ml ● header_blocks
A "[" based on:
    access_op →  ● "[" _ expr _ "]"
    access_ops →  ● access_op
    path$ebnf$2$subexpression$1 → _ ● access_ops
    path$ebnf$2 →  ● path$ebnf$2$subexpression$1
    path → binding_form path$ebnf$1 ● path$ebnf$2
    path_assign → path_assign$ebnf$1 __ ● path _ "=" _ assign_expr
    statement →  ● path_assign
    statements → _ ● statement _c_ nl statements
    statements → _c_ nl ● statements
    block → "{" ● statements "}"
    header_block → header ● block _ml
    header_blocks$ebnf$1 → header_blocks$ebnf$1 ● header_block
    header_blocks →  ● header_blocks$ebnf$1
    input → _ml ● header_blocks
```

This is way too much! And to someone who doesn't know about grammars, all this text is completely inscrutable and useless. The special symbols like ● just make it look scary. But we can actually extract a pretty decent error message from this:
```
Error: Syntax error at line 7 col 10:

    x.icon ==
           ^
Unexpected token: "==". Instead, I was expecting to see one of the following:

    below
    above
    =
    [
```

This is actually useful, and by glancing at the 4-item list, the user can immediately see that they need to write `=` instead of `==`.

Another example, from https://github.com/penrose/penrose/issues/985#issuecomment-1150528695:
```
Error: Syntax error at line 7 col 18:

      shape Circle {
                   ^
Unexpected lbrace token: "{". Instead, I was expecting to see one of the following:

A ws token based on:
    _$ebnf$1 → _$ebnf$1 ● %ws
    _ →  ● _$ebnf$1
    path$ebnf$2$subexpression$1 →  ● _ access_ops
    path$ebnf$2 →  ● path$ebnf$2$subexpression$1
    path → binding_form path$ebnf$1 ● path$ebnf$2
    path_assign → path_assign$ebnf$1 __ ● path _ "=" _ assign_expr
    statement →  ● path_assign
    statements → _ ● statement _c_ nl statements
    statements → _c_ nl ● statements
    block → "{" ● statements "}"
    header_block → header ● block _ml
    header_blocks$ebnf$1 → header_blocks$ebnf$1 ● header_block
    header_blocks →  ● header_blocks$ebnf$1
    input → _ml ● header_blocks
A "=" based on:
    path_assign → path_assign$ebnf$1 __ path _ ● "=" _ assign_expr
    statement →  ● path_assign
    statements → _ ● statement _c_ nl statements
    statements → _c_ nl ● statements
    block → "{" ● statements "}"
    header_block → header ● block _ml
    header_blocks$ebnf$1 → header_blocks$ebnf$1 ● header_block
    header_blocks →  ● header_blocks$ebnf$1
    input → _ml ● header_blocks
A "[" based on:
    access_op →  ● "[" _ expr _ "]"
    access_ops →  ● access_op
    path$ebnf$2$subexpression$1 → _ ● access_ops
    path$ebnf$2 →  ● path$ebnf$2$subexpression$1
    path → binding_form path$ebnf$1 ● path$ebnf$2
    path_assign → path_assign$ebnf$1 __ ● path _ "=" _ assign_expr
    statement →  ● path_assign
    statements → _ ● statement _c_ nl statements
    statements → _c_ nl ● statements
    block → "{" ● statements "}"
    header_block → header ● block _ml
    header_blocks$ebnf$1 → header_blocks$ebnf$1 ● header_block
    header_blocks →  ● header_blocks$ebnf$1
    input → _ml ● header_blocks
```

This can similarly be shortened:
```
Error: Syntax error at line 7 col 18:

      shape Circle {
                   ^
Unexpected token: "{". Instead, I was expecting to see one of the following:

    =
    [
```

If Nearley provides a structured representation of errors then it would be best to just write our own printer for it which presents error messages in this way, but if not then we can probably just write a simple text processing function that reduces it to this simpler form.

samestep:
# Description

Resolves #1052.

[The `reportError` function](https://github.com/kach/nearley/blob/6e24450f2b19b5b71557adf72ccd580f4e452b09/lib/nearley.js#L382-L386) @wodeni mentioned makes use not just of the `Parser`, but also of the specific `token` that caused the error. I didn't know how to access that token from the error we `catch`, although after looking more closely at [the Nearley code](https://github.com/kach/nearley/blob/6e24450f2b19b5b71557adf72ccd580f4e452b09/lib/nearley.js#L343-L346), it seems to be available via the `token` field. Thus, to me it seemed that we would need to do some textual processing of the error message regardless, just to get the line and column number.

In any case, the `prettyParseError` function I've written here is fairly small and easy to read, and (because of the way it is written) should degrade gracefully to the identity function if Nearley's error message formatting changes.

Replace one of the `=` tokens with `==` in `packages/examples/src/set-theory-domain/venn.sty`, then run this command from within the `packages/automator/` directory:

```sh
yarn start draw tree.sub venn.sty setTheory.dsl out --src-prefix=../examples/src/set-theory-domain --variation=PlumvilleCapybara104
```

## Before

See the first example in #1052.

## After

```
Error: Syntax error at line 7 col 10:

    x.icon ==
           ^
Unexpected token: "==". Instead, I was expecting to see one of the following:

    ws token
    "below"
    "above"
    "="
    "["
```

# Checklist

### DIFF
--- a/packages/core/src/parser/ParserUtil.ts
+++ b/packages/core/src/parser/ParserUtil.ts
@@ -191,3 +191,26 @@ export const lastLocation = (parser: nearley.Parser): SourceLoc | undefined => {
     return undefined;
   }
 };
+
+export const prettyParseError = (e: unknown): string => {
+  const lines: string[] = [];
+
+  let inStart = true;
+  const alreadySeen = new Set<string>();
+  for (const line of `${e}`.split("\n")) {
+    const match = line.match(/^A (.*) based on:$/);
+    if (match !== null) {
+      inStart = false;
+      const option = match[1];
+      if (!alreadySeen.has(option)) {
+        alreadySeen.add(option);
+        lines.push(`    ${option}`);
+      }
+    } else if (inStart) {
+      lines.push(line);
+    }
+  }
+
+  lines.push("");
+  return lines.join("\n");
+};
--- a/packages/core/src/parser/StyleParser.test.ts
+++ b/packages/core/src/parser/StyleParser.test.ts
@@ -200,11 +200,44 @@ where C := intersect ( A, B, Not(f) ) ;\
     const prog = `
     forall Set x; Set y where y := Baz(x) as foo {}
     `;
-    expect(parseStyle(prog).isErr()).toEqual(true);
+    const parsed = parseStyle(prog);
+    if (parsed.isErr()) {
+      expect(parsed.error.message)
+        .toEqual(`Error: Syntax error at line 2 col 43:
+
+      forall Set x; Set y where y := Baz(x) as
+                                            ^
+Unexpected as token: "as". Instead, I was expecting to see one of the following:
+
+    ws token
+    nl token
+    ";"
+    comment token
+    multiline_comment token
+    "with"
+    "{"
+`);
+    } else {
+      throw Error("expected parse error");
+    }
   });
   test("cannot set subVars as aliases", () => {
     const prog = "forall Set x; Set y where IsSubset(x,y) as `A` {}";
-    expect(parseStyle(prog).isErr()).toEqual(true);
+    const parsed = parseStyle(prog);
+    if (parsed.isErr()) {
+      expect(parsed.error.message)
+        .toEqual(`Error: Syntax error at line 1 col 44:
+
+  forall Set x; Set y where IsSubset(x,y) as \`
+                                             ^
+Unexpected tick token: "\`". Instead, I was expecting to see one of the following:
+
+    ws token
+    identifier token
+`);
+    } else {
+      throw Error("expected parse error");
+    }
   });
 
   test("label field check", () => {
--- a/packages/core/src/compiler/Domain.ts
+++ b/packages/core/src/compiler/Domain.ts
@@ -3,7 +3,7 @@ import im from "immutable";
 import { every, keyBy, zipWith } from "lodash";
 import nearley from "nearley";
 import domainGrammar from "parser/DomainParser";
-import { idOf, lastLocation } from "parser/ParserUtil";
+import { idOf, lastLocation, prettyParseError } from "parser/ParserUtil";
 import { A, C } from "types/ast";
 import {
   Arg,
@@ -57,7 +57,7 @@ export const parseDomain = (
       return err(parseError(`Unexpected end of input`, lastLocation(parser)));
     }
   } catch (e) {
-    return err(parseError(<string>e, lastLocation(parser)));
+    return err(parseError(prettyParseError(e), lastLocation(parser)));
   }
 };
 
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -12,7 +12,7 @@ import { alg, Edge, Graph } from "graphlib";
 import im from "immutable";
 import _, { range } from "lodash";
 import nearley from "nearley";
-import { lastLocation } from "parser/ParserUtil";
+import { lastLocation, prettyParseError } from "parser/ParserUtil";
 import styleGrammar from "parser/StyleParser";
 import seedrandom from "seedrandom";
 import {
@@ -2912,8 +2912,8 @@ export const parseStyle = (p: string): Result<StyProg<C>, ParseError> => {
     } else {
       return err(parseError(`Unexpected end of input`, lastLocation(parser)));
     }
-  } catch (e: unknown) {
-    return err(parseError(<string>e, lastLocation(parser)));
+  } catch (e) {
+    return err(parseError(prettyParseError(e), lastLocation(parser)));
   }
 };
 
--- a/packages/core/src/compiler/Substance.ts
+++ b/packages/core/src/compiler/Substance.ts
@@ -2,7 +2,7 @@ import { dummyIdentifier } from "engine/EngineUtils";
 import im from "immutable";
 import { findIndex } from "lodash";
 import nearley from "nearley";
-import { idOf, lastLocation } from "parser/ParserUtil";
+import { idOf, lastLocation, prettyParseError } from "parser/ParserUtil";
 import substanceGrammar from "parser/SubstanceParser";
 import { A, ASTNode, C, Identifier } from "types/ast";
 import {
@@ -69,7 +69,7 @@ export const parseSubstance = (
       return err(parseError(`Unexpected end of input`, lastLocation(parser)));
     }
   } catch (e) {
-    return err(parseError(<string>e, lastLocation(parser)));
+    return err(parseError(prettyParseError(e), lastLocation(parser)));
   }
 };
 


joshsunshine:
@samestep can you add examples of buggy code and the error messages before and after this change to the PR description?

samestep:
@joshsunshine Done.

packages/core/src/parser/ParserUtil.ts
```diff
@@ -191,3 +191,26 @@ export const lastLocation = (parser: nearley.Parser): SourceLoc | undefined => {
...
   }
 };
+
+export const prettyParseError = (e: unknown): string => {
+  const lines: string[] = [];
+
+  let inStart = true;
+  const alreadySeen = new Set<string>();
+  for (const line of `${e}`.split("\n")) {
+    const match = line.match(/^A (.*) based on:$/);
```
wodeni, line 201: good old `(.*)` saving the day :p

packages/core/src/parser/StyleParser.test.ts
```diff
@@ -200,11 +200,44 @@ where C := intersect ( A, B, Not(f) ) ;\
...
+    const parsed = parseStyle(prog);
+    if (parsed.isErr()) {
+      expect(parsed.error.message)
+        .toEqual(`Error: Syntax error at line 2 col 43:
+
+      forall Set x; Set y where y := Baz(x) as
+                                            ^
+Unexpected as token: "as". Instead, I was expecting to see one of the following:
+
+    ws token
```
wodeni, line 212: Seems like this depends on how we name the tokens. `whitespace` and `newline` will certainly be more readable than `ws` and `nl`.

wodeni (approved):
LGTM! The token naming issue below should be a separate PR IMO.

wodeni:
> [The `reportError` function](https://github.com/kach/nearley/blob/6e24450f2b19b5b71557adf72ccd580f4e452b09/lib/nearley.js#L382-L386) @wodeni mentioned makes use not just of the `Parser`, but also of the specific `token` that caused the error. I didn't know how to access that token from the error we `catch`, although after looking more closely at [the Nearley code](https://github.com/kach/nearley/blob/6e24450f2b19b5b71557adf72ccd580f4e452b09/lib/nearley.js#L343-L346), it seems to be available via the `token` field. Thus, to me it seemed that we would need to do some textual processing of the error message regardless, just to get the line and column number.

Looks like the source location is encoded as a single `offset` in the error from looking at https://github.com/kach/nearley/blob/6e24450f2b19b5b71557adf72ccd580f4e452b09/lib/nearley.js#L296. Doesn't look like this is used in the actual error reporting?

packages/core/src/parser/StyleParser.test.ts
samestep: Yep, I agree; thankfully this is a pretty easy fix!

samestep:
> Looks like the source location is encoded as a single `offset` in the error from looking at https://github.com/kach/nearley/blob/6e24450f2b19b5b71557adf72ccd580f4e452b09/lib/nearley.js#L296. Doesn't look like this is used in the actual error reporting?
> 
> If `offset` is indeed accurate, we might not even have to deal with the nearley error string (and don't have to use regexes!)

Ah, interesting. Do you have a strong preference? I personally feel like this regex approach is easier, but I'm willing to switch to the other approach if you want.

wodeni:
> > Looks like the source location is encoded as a single `offset` in the error from looking at https://github.com/kach/nearley/blob/6e24450f2b19b5b71557adf72ccd580f4e452b09/lib/nearley.js#L296. Doesn't look like this is used in the actual error reporting?
> > If `offset` is indeed accurate, we might not even have to deal with the nearley error string (and don't have to use regexes!)
> 
> Ah, interesting. Do you have a strong preference? I personally feel like this regex approach is easier, but I'm willing to switch to the other approach if you want.

Well going from offset to coordinates will also involve [newline counting](https://github.com/kach/nearley/blob/6e24450f2b19b5b71557adf72ccd580f4e452b09/lib/nearley.js#L216), but I guess it's slightly better because we don't have to rely on nearley's pretty printer. For the time being I don't think this is super high priority though.

samestep:
> Well going from offset to coordinates will also involve [newline counting](https://github.com/kach/nearley/blob/6e24450f2b19b5b71557adf72ccd580f4e452b09/lib/nearley.js#L216), but I guess it's slightly better because we don't have to rely on nearley's pretty printer. For the time being I don't think this is super high priority though.

OK! Hmm, yeah I'm leaning toward just merging this, then?

joshsunshine:
> ws token
What does this mean?

samestep:
> > `ws token`
> 
> What does this mean?

It means "whitespace token". As @wodeni pointed out, in a followup PR we can change the names of our Nearley rules so it will literally say "whitespace token" instead of "ws token".

wodeni:
> > `ws token`
> 
> 
> 
> What does this mean? 
> 
> 

It's the name for the whitespace token internally. See the naming discussion above

joshsunshine (approved):

PR #1085, opened 2022-09-09
fix: Generate code using heap instead of stack

### LINKED ISSUE #1070: Stack overflow with too many substance objects
**Describe the bug**
I've noticed that Penrose crashes with a stack overflow with programs that have too many substance types or relations. I don't know whether the problem is the number of types or the number of relations.

**To Reproduce**
Edit the [`hundred-points-around-star`](https://github.com/penrose/penrose/blob/main/packages/examples/src/shape-distance/hundred-points-around-star.sub) example to include 1000 points and `Around` predicates and the system will crash with a Stack Overflow.

samestep:
# Description

Fixes #1070 by eliminating all local variables in the function constructed by `genCode`, instead using a `bools` array and a `nums` array. Preliminary performance results suggest that this makes Penrose almost twice as slow.

Currently this fails all the tests that use `polyRoots`, because I have not yet handled `ad.Vec`.

To make this at least somewhat efficient, we need to separate `boolean`s from `number`s. There are two obvious approaches:

1. Keep `makeGraph` the same, then separate them out in `genCode`.
2. Separate them out in `makeGraph`, keeping `genCode` the same.

The first approach is what I used for [my experiment where I changed `genCode` to produce C instead of JS](https://github.com/penrose/experiments/blob/05d231a8f5d8faf6fb1dcf642fde1c6d78f08634/2022-optimizer-performance/penrose.patch#L306-L336). I ended up going with the second approach here because it means I don't have to make a mapping between IDs in the graph and indices in these `bools` and `nums` arrays. I'm really not sure which is better, though; the second approach makes some of the logic in `makeGraph` messier.

As an aside, since I was touching code nearby, I cleaned up our unnecessarily verbose definition for `ad.Edge`.

In one terminal:

```sh
yarn
cd packages/examples/src/shape-distance/
node -e 'for (let i = 100; i < 1000; i++) { console.log(`Point p${i}\nAround(p${i},g)`); }' \
  >> hundred-points-around-star.sub
roger watch
```

Then in a separate terminal:

```
yarn start
```

Open http://localhost:3000/try/ in your browser, select the following files, and click "compile":

- `hundred-points-around-star.sub`
- `shape-distance.sty`
- `shapes.dsl`

Before this PR, you should get this error:

```
Uncaught RangeError: Maximum call stack size exceeded
    at eval (eval at genCode (Autodiff.ts:1094:13), <anonymous>:3:12)
    at Autodiff.ts:1095:22
    at Optimizer.ts:812:46
    at minimize (Optimizer.ts:659:56)
    at step (Optimizer.ts:152:19)
    at stepStateSafe (index.ts:75:15)
    at step (DiagramPanel.tsx:78:28)
```

After this PR, it should just work:

points-around-star

# Checklist

### DIFF
--- a/packages/core/src/engine/Autodiff.ts
+++ b/packages/core/src/engine/Autodiff.ts
@@ -365,8 +365,12 @@ const children = (x: ad.Expr): Child[] => {
   }
 };
 
-const indexToID = (index: number): ad.Id => `_${index}`;
-const idToIndex = (id: ad.Id): number => parseInt(id.slice(1), 10);
+const idToNumsIndex = (id: ad.Id): number =>
+  id.startsWith(ad.numsName)
+    ? parseInt(id.slice(ad.numsName.length + 1, id.length - 1), 10)
+    : // in this case we're looking at an edge pointing to a `Bool` node, which
+      // has no gradient, so it doesn't matter how we sort it
+      -1;
 
 const getInputs = (
   graph: ad.Graph["graph"]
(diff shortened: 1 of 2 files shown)


samestep:
> Preliminary performance results suggest that this makes Penrose almost twice as slow.

Closing for this reason.

PR #1092, opened 2022-09-13
perf: Port the optimizer to WebAssembly

samestep:
# Description

This PR improves optimizer performance by an order of magnitude, building on [my optimizer performance experiment](https://github.com/penrose/experiments/tree/main/2022-optimizer-performance) from earlier this year.

This PR adds the following build tools to our setup:

- [Cargo](https://doc.rust-lang.org/cargo/) for Rust stuff
- [`wasm-bindgen`](https://github.com/rustwasm/wasm-bindgen) to generate JavaScript/TypeScript scaffolding around Wasm compiled from Rust
- [ts-rs](https://github.com/Aleph-Alpha/ts-rs) to generate TypeScript types from Rust [Serde](https://serde.rs/) types

Quoting [from Stack Overflow](https://stackoverflow.com/a/45285081):

> Some browsers limit the size of modules that can be compiled synchronously because that blocks the main thread.

One implication of this is that compiling an autodiff graph becomes an `async` operation, so in particular, `compileStyle` (and thus also `compileTrio`) becomes `async`.

Another implication is that loading/initializing the optimizer _itself_ is an `async` operation. We considered three possible architectures to accommodate this:

1. Don't initialize the optimizer Wasm module at load time, and instead, make each of the functions exposed by the Wasm module wrapper have an implicit precondition that the module must have already been loaded; this has the benefit that we don't need top-level `await` and we don't need to make all those functions `async`, but the drawback that we now need to remember to properly initialize the module before every possible usage or we'll have race conditions.
2. Change the interface of the Wasm module wrapper so that rather than exposing things directly, it exposes just one `async` function that initializes the Wasm module and returns an object with methods to access all the underlying Wasm functions; this has the advantage of not requiring top-level `await` and not giving things implicit preconditions, but the viral disadvantage that now every consumer of the Wasm module needs to either be `async` or take in this object as a parameter.
3. Use top-level `await`, and update our downstream packages to use ESM instead of CJS; this has the advantage of no race conditions, no implicit preconditions, and no viral changes to downstream function signatures, but the disadvantage that `docs-site` needs to support top-level `await`.

This PR takes approach (1), but we would like to switch to (3) in the future if possible; see the further discussion below.

# Checklist

# Open questions

- To reduce the time to compile the WebAssembly to native code, a followup should identify sets of objectives/constraints that are topologically similar (usually those sharing a source location in the Style program) and compile those to a single WebAssembly function, instead of duplicating their logic across many functions.
- Because [a `WebAssembly.Module` "can be efficiently shared with Workers, and instantiated multiple times"](https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Module), a followup should allow the optimizer to be run in a web worker; see this issue: #1095
- If a generated gradient function crashes while computing polynomial roots, it does not undo the change it made to the stack pointer. I don't know of a good way to handle this. But on the other hand, I don't know whether it's possible for such a crash to occur.
- While developing this PR, I added `${FORCE_COLOR+--color=true}` to `@penrose/optimizer`'s `esbuild` script, and `${FORCE_COLOR+--color=always}` to its `cargo` scripts, to get color output in Nx. I've removed those because they're not portable to Windows, but it would be nice to come up with a portable way to get this back in the future; see this issue: #1176
- Originally I expected that this PR would allow us to handle a thousand points in our `hundred-points-around-star` example (which would previously yield a stack overflow), but instead, after this PR it simply crashes the whole page. See #1070
- In the medium term, it would be nice to remove all our instances of `await ready;` by using a top-level `await` in `@penrose/optimizer`. As far as I can tell, this would require us to get rid of Webpack first; see #1172
- I removed the autodiff `debug` node, because I don't think we ever use it, and it would have taken nontrivial effort to port to WebAssembly. I can add it back if we want, but didn't want to waste time if we don't care.
- This PR uses a prerelease version of `wasm-bindgen` that includes the `--keep-lld-exports` flag and does not emit `BigInt` literals. Once version 0.2.84 is released, we should just switch to that, but for now, the `CONTRIBUTING.md` instructions just say to install from GitHub at a specific commit, and our CI/CD scripts pull a Linux binary which I compiled and put in [a GitHub Gist](https://gist.github.com/samestep/2cf703cfd81691f2cb3c23422fce7e56).
- Although we mostly compile Wasm asynchronously for reasons described above, there are two exceptions: in tests, and for convex partitioning. The former is fine because we only run tests in Node and not in the browser, but the latter may become an issue if we end up using convex partitioning on polygons with too many points or whose point locations are computed via sufficiently complicated expressions.
- As you can verify by running `strings packages/optimizer/target/wasm32-unknown-unknown/release/penrose_optimizer.wasm`, the built WebAssembly binary contains absolute paths (starting with `/Users/samueles/.cargo/registry/src/github.com-1ecc6299db9ec823/` in my case) to local files on whatever machine built the binary. See this Rust issue: rust-lang/rust#40552


kyleliangus:
Almost there!

kyleliangus:
Keep it up!

joshsunshine:
> While developing this PR, I added ${FORCE_COLOR+--color=true} to @penrose/optimizer's esbuild script, and ${FORCE_COLOR+--color=always} to its cargo scripts, to get color output in Nx. I've removed those because they're not portable to Windows, but it would be nice to come up with a portable way to get this back in the future.

@samestep can you write an issue about this? Perhaps @liangyiliang can implement a workaround for Windows since he is our main Windows user.

joshsunshine:
> Originally I expected that this PR would allow us to handle a thousand points in our hundred-points-around-star example (which would previously yield a stack overflow), but instead, after this PR it simply crashes the whole page. See https://github.com/penrose/penrose/issues/1070

@samestep can you update 1070 so it explains the new issue.

samestep:
> > While developing this PR, I added `${FORCE_COLOR+--color=true}` to `@penrose/optimizer`'s `esbuild` script, and `${FORCE_COLOR+--color=always}` to its `cargo` scripts, to get color output in Nx. I've removed those because they're not portable to Windows, but it would be nice to come up with a portable way to get this back in the future.
> 
> @samestep can you write an issue about this? Perhaps @liangyiliang can implement a workaround for Windows since he is our main Windows user.

Done: #1176

> > Originally I expected that this PR would allow us to handle a thousand points in our hundred-points-around-star example (which would previously yield a stack overflow), but instead, after this PR it simply crashes the whole page. See #1070
> 
> @samestep can you update 1070 so it explains the new issue.

Done: https://github.com/penrose/penrose/issues/1070#issuecomment-1360162984

wodeni (approved):
LGTM! Thanks for going through the PR with me @samestep :p.

PR #1094, opened 2022-09-14
feat: Style inline color widgets

wodeni:
# Description

Related issue/PR: N/A

@keenancrane asked:

> Is a pop-up color picker something that could be easily integrated into the framework you're using for the IDE?

This PR adds support for `ColorProvider` in our monaco editor for Style, which displays a color icon and on-hover color picker for all `rgba` function calls.

Since we do not have a language server, this PR is implemented in the simplest way: parsing the document with a regex for valid `rgba` function calls. Currently, this feature does not support `hsva` or any other functions that return colors. In addition, the arguments to `rgba` have to be floating-point literals; inline expression are not supported yet.

image

# Checklist

# Open questions

Ideally, we should have a language server that supports [Color decorators](https://code.visualstudio.com/api/language-extensions/programmatic-language-features#show-color-decorators). The server should tokenize and parse the Style program, and the color provider reasons over tokens and the AST instead of matching on text to find source ranges for the decorators.

### DIFF
--- a/packages/components/src/editing/languages/StyleConfig.ts
+++ b/packages/components/src/editing/languages/StyleConfig.ts
@@ -1,6 +1,6 @@
 import { Monaco } from "@monaco-editor/react";
 import { compDict, constrDict, objDict, shapedefs } from "@penrose/core";
-import { IRange, languages } from "monaco-editor";
+import { editor, IRange, languages } from "monaco-editor";
 import { CommentCommon, CommonTokens } from "./common";
 
 export const StyleConfig: languages.LanguageConfiguration = {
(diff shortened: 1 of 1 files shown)


packages/components/src/editing/languages/StyleConfig.ts
```diff
@@ -163,20 +163,56 @@ export const SetupStyleMonaco = (monaco: Monaco) => {
...
+        true
+      );
+      console.log(colorMatches);
+
+      return colorMatches.map(({ matches, range }) => ({
+        color: {
+          red: +matches![1],
+          green: +matches![2],
+          blue: +matches![3],
+          alpha: +matches![4],
```
samestep, line 193: Whoa! What is this syntax? :joy: I've never seen this before

wodeni: this is lazy nimo trying to cast to float ignoring if there's an error with the matches. Adding in a null check to get rid of the `!` at least.

packages/components/src/editing/languages/StyleConfig.ts
```diff
@@ -163,20 +163,56 @@ export const SetupStyleMonaco = (monaco: Monaco) => {
...
+      const { red, green, blue, alpha } = colorInfo.color;
+      return [
+        {
+          label: `rgba(${red}, ${green}, ${blue}, ${alpha})`,
+        },
+      ];
+    },
+
+    provideDocumentColors: (model) => {
+      const colorRegex = /rgba\(\s*(.*)\s*,\s*(.*)\s*,\s*(.*)\s*,\s*(.*)\s*\)/;
```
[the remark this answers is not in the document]
wodeni: Yeah a bit. Refining this takes more effort than necessary IMO (float regex can be messy). My thought is to allow the color icon to show up at least (even with the wrong color), so the user can at least access the picker.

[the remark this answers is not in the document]
samestep: Makes sense!

packages/components/src/editing/languages/StyleConfig.ts
samestep: 👌 seems legit

packages/components/src/editing/languages/StyleConfig.ts
```diff
@@ -163,20 +163,56 @@ export const SetupStyleMonaco = (monaco: Monaco) => {
...
+      const colorMatches = model.findMatches(
+        colorRegex.source,
+        false,
+        true,
+        false,
+        null,
+        true
+      );
+      console.log(colorMatches);
```
samestep, line 186: Should remove this before merging.

PR #1104, opened 2022-09-20
fix: Change the derivative of abs to be sign

samestep:
# Description

This PR changes the derivative of `Math.abs(v)` to be `Math.sign(v)` instead of the less-correct `v / (Math.abs(v) + 0.00001)`.

Here's a graph comparing the two derivatives:

```mathematica
Plot[{v/(Abs[v] + 0.00001), Sign[v]}, {v, -.001, .001}]
```

image

# Checklist

# Open questions

- As mentioned in #907:

  > There are several places where `EPS_DENOM = 10e-6` is added to denominators to try to avoid divide-by-zero issues. This is obviously inelegant, and also doesn't resolve the issue in general because it can even cause divide-by-zero where it wouldn't have happened otherwise, if the original denominator was a very small negative number. Is there a better way to approach this?


samestep:
I'm worried about the change in `diagrams/persistent-homology-persistent-homology.svg` caused by this PR.

wodeni:
> I'm worried about the change in `diagrams/persistent-homology-persistent-homology.svg` caused by this PR.

Can you elaborate why? Looks like `diagrams/wos-nested-estimator-walk-on-spheres.svg` changed quite a bit visually, but I can't tell if there's anything wrong with that one.

samestep:
> Can you elaborate why? Looks like `diagrams/wos-nested-estimator-walk-on-spheres.svg` changed quite a bit visually, but I can't tell if there's anything wrong with that one.

Oh! You're right, I don't know why I was worried. I thought the triangle was changing from equilateral to non-equilateral so I got concerned, but I now see that it's the other way around.

PR #1111, opened 2022-09-26
feat: support path lists in `layer` expressions

wodeni:
# Description

Related issue/PR: N/A

There are a lot of duplicated layering statements in Style programs such as `euclidean.sty`. Mostly saying one thing should be below/above multiple things:

```
  t.PQ above P.icon
  t.QR above P.icon
  t.RP above P.icon
```

This PR lets `layer` expressions take a list of paths on the right-hand side of the layering operator. For instance, the above can be abbreviated to `P.icon below t.PQ, t.QR, t.RP`. 

* The Style parser parses all `layer` expressions as `path ("below"|"above") path_list`. There's only one path allowed on the left to simplify the semantics. `path_list` can be a list of just one element.
  * The parser is changed to not special-case on `below` and `above` cases. In general, I try to maintain the original code structure in the AST and not do any computation (e.g. sorting `below`, `above` pairs) in the parser. This is mostly for better language services.
* In the Style compiler, we convert the layering list expressions to individual layering pairs.

# Checklist


packages/core/src/compiler/Style.ts
```diff
@@ -2070,7 +2070,7 @@ const findPathsExpr = <T>(expr: Expr<T>): Path<T>[] => {
       return expr.properties.flatMap((prop) => findPathsExpr(prop.value));
     }
     case "Layering": {
-      return [expr.below, expr.above];
+      return [expr.left, ...expr.right];
```
samestep, line 2073: The reason `diagrams/` are being updated is because of this line. If you change it to the following, behavior is the same as before:
```suggestion
      if (expr.layeringOp === "below") return [expr.left, ...expr.right];
      else return [...expr.right, expr.left];
```
But I don't think we should actually add this complexity just to maintain the exact same behavior. Could you update the `variation` for `diagrams/siggraph-teaser-euclidean-teaser.svg` to make it look nice again, though?

wodeni: Agreed. Perhaps it's a bit odd that a users changes from `above` to `below` and sees diagram changes in the IDE. How significant is the ordering here? Can we normalize it somehow?

samestep: I think this should be part of a larger discussion about what kinds of stability we should try to support beyond raw determinism.

wodeni: Sounds good. Let's talk more about this later. Merging this PR now.

PR #1114, opened 2022-09-27
feat: hexadecimal color literals in Style

wodeni:
# Description

Related issue/PR: #1094 

The non-standard `rgba` format Style used had been a pain to work with. This PR adds native support for hexadecimal color literals as specified in [web standards](https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color), i.e.:

```
#RGB        // The three-value syntax
#RGBA       // The four-value syntax
#RRGGBB     // The six-value syntax
#RRGGBBAA   // The eight-value syntax
```

Both upper- and lower-case hex digits are supported.

Style writers can use standard hex colors with highlighting and color picker support in the IDE. 

image

* `RGBA` is the ground-truth storage format in the Style compiler, and `#rrggbbaa` is the output format used by the color picker.
* Add tests to validate values of hex literals
* Updated `continousmap.sty` to use hex literals
* Side mission: found a bunch of very duplicated code in `Style.test.ts`. Pulled them into `loadProgs` to simplify things.

# Checklist

# Open questions

### DIFF
--- a/packages/core/src/compiler/Style.test.ts
+++ b/packages/core/src/compiler/Style.test.ts
@@ -5,18 +5,17 @@ import * as S from "compiler/Style";
 import { buildAssignment } from "compiler/Style";
 import { compileSubstance } from "compiler/Substance";
 import im from "immutable";
-import { A, C } from "types/ast";
+import { C } from "types/ast";
 import { Either } from "types/common";
 import { Env } from "types/domain";
 import { PenroseError } from "types/errors";
 import { State } from "types/state";
 import { StyProg } from "types/style";
 import { Layer } from "types/styleSemantics";
-import { SubProg, SubRes, SubstanceEnv } from "types/substance";
-import { andThen, Result, showError, unsafelyUnwrap } from "utils/Error";
+import { SubstanceEnv } from "types/substance";
+import { andThen, Result, showError } from "utils/Error";
 import { foldM, toLeft, ToRight } from "utils/Util";
 import { compileDomain } from "./Domain";
-import { envOrError, subEnvOrError } from "./Substance.test";
 
 // TODO: Reorganize and name tests by compiler stage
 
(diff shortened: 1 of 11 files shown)


packages/core/src/compiler/Style.test.ts
```diff
@@ -56,17 +59,65 @@ export const loadProgs = ({
...
+
+  // third pass: compile all expressions in topological sorted order
+  const translation = S.translate(
+    { makeInput },
+    canvas,
+    graph,
+    assignment.diagnostics.warnings
+  );
+
+  const state = S.compileStyle(
```
wodeni, line 109: I haven't really copied the entirety of `compileStyle` in here so this is technically running things twice. Wondering if there's a better way to do this.

packages/core/src/parser/ParserUtil.ts
```diff
@@ -17,6 +17,7 @@ export const basicSymbols: moo.Rules = {
   rparen: ")",
   apos: "'",
   comma: ",",
+  hex_literal: /#(?:[0-9A-Fa-f]{3,})/,
```
samestep, line 20: Why do you use `(?:)` here?

packages/core/src/utils/Util.ts
```diff
@@ -177,14 +177,75 @@ export const toScreen = (
 
 //#region color
 
-export const toHexRGB = (color: [number, number, number]): string => {
+export const hexToRgba = (
+  hex: string
+): [number, number, number, number] | null => {
```
samestep, line 182: By our conventions from #873:
```suggestion
): [number, number, number, number] | undefined => {
```

packages/core/src/parser/ParserUtil.ts
wodeni: Probably from a previous attempt at capturing just the hex string. Removing now.

packages/core/src/compiler/Style.test.ts
wodeni: Yeah, I basically just want to run the compiler up to a certain pass. @liangyiliang had some tests that ran it all the way, but many others only need the translation.

PR #1117, opened 2022-10-06
feat: improve `euclidean.sty`

wodeni:
# Description

Related issue/PR: #1133 

This PR improves `euclidean.sty`, the Style program with the most example Substances in `@penrose/examples`. We also use this Style for the presets in `synthesizer-ui`. 

Below are some notes on each example in `geometry-domain/textbook-problems`.

<google-sheets-html-origin><style type="text/css"></style>

ID | Style issues
-- | --
c01p01 | * [ ] `On` works but the point can go outside of the ray length if the linelike shape is mkray* [ ] Labels overlap
c01p10 | * [x] mistakenly uses `pointOnLineDist`. Changed to `atDist` instead* [x] add label disjoint to line segment
c02p01 | * [x] equal angle markers are too small. Bumping the constant is okay for this one, but really it should take the angle size into account to look pretty
c03p01 | * [ ] ParallelMarker1 is still there. Should use match_id* [ ] `KelpLemur0794` yield very short line with lots of constraints violated* [x] slope-based parallel constraint behaves weirdly. Changed to dot product to simplify, not sure if behavior improves
c04p01 | * [x] if collinear is just the cross product, then points like to overlap. Instead of reverting `collinear`, I added two `inRange` constraints in the Style program for `Collinear`* [ ] the question: mathematically collinearity doesn't imply point ordering. Seems like this should be captured in Substance instead* [x] removed the `centerArrow` function and the util function it calls* [x] set `Simple` to be `animate: true` so it doesn't block the UI
c04p12 | * [x] using segments to construct a triangle, not `Triangle`* [x] `equalAngle` uses `dot`, leading to incorrect outputs* [x] `angleBetween` gives small values, leading to bad optimization output* [x] the unknown angle `x` is not labeled
c05p01 | * [x] use of `p.icon.center` instead of `p.vec`. Not sure which way is better tbh
c05p13 | This is the incenter example, so no problems!
c06p06 | * [x] `Parallelogram` is wayy to generic. Use other predicates instead * [x] make sure labels don't go into a `Quadralateral`
c07p02 |  
c07p06 | pretty good diagram. no problems
c07p10 |  
c07p22 | - [ ] the triangles might overlap with each other. This is definitely not a concept that our domain program understands. Is this mathemattical, or presentational?
c08p08 | - [ ] optimization gets a little slow
c08p18 |  
c09p02 |  
c10p08 | - [ ] `angleBetween` is unsigned, `AngleBisector` really needs to know if the bisector is somehow "between" the angle sides.
c11p07 | - [x] `Diameter` not behavintg well due to bad constraints- [x] `Chord` and `Diameter` labels inside circle
c11p12 | - [x] `Radius labeling
c11p25 |  
c12p12 | - [x] the transitive EqualLengthMarker. Found a workaround by redeclaring shapes?

# Checklist

# Open questions


packages/examples/src/geometry-domain/geometry.dsl
```diff
@@ -80,8 +78,7 @@ predicate Orthocenter(Point, Triangle)
 predicate Centroid(Point, Triangle)
 predicate Circumcenter(Point, Triangle)
 
--- Unimplemented
-predicate Supplementary(Angle) -- TODO broken
+
```
samestep, line 81:
```suggestion
```

samestep (commented):
A few minor comments for now; will try to come back and review this in more detail later. Thanks for all the work you've put into this!

packages/core/src/contrib/Constraints.ts
```diff
@@ -116,21 +117,19 @@ const constrDictSimple = {
...
-    return max(0, sub(add(ops.vnorm(v1), ops.vnorm(v2)), ops.vnorm(v3)));
+    return absVal(ops.cross2(v1, v2));
   },
 
   /**
    * Require that three points be collinear.
    * Does not enforce a specific ordering of points, instead it takes the arrangement of points that is most easily satisfiable.
    */
-  collinearUnordered: (c1: ad.Num[], c2: ad.Num[], c3: ad.Num[]) => {
+  collinearOrdered: (c1: ad.Num[], c2: ad.Num[], c3: ad.Num[]) => {
```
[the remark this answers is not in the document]
wodeni: After testing the unordered, cross-product version of `collinear`, it seemed to be a more reasonable default. That's why I demoted my original triangle-inequality constraint to `collinearOrdered`.

packages/examples/src/geometry-domain/textbook_problems/ex.sub
```diff
@@ -21,22 +18,19 @@ Angle aXYZ := InteriorAngle(X, Y, Z)
...
-EqualAngleMarker1(aUTS, aZXY)
+EqualAngleMarker(aUTS, aZXY)
 EqualAngle(aUTS, aZXY)
-EqualAngleMarker2(aUTS, aXYZ)
+EqualAngleMarker(aUTS, aXYZ)
 EqualAngle(aUTS, aXYZ)
-EqualAngleMarker3(aTUS, aYZX)
-AutoLabel All
\ No newline at end of file
+EqualAngleMarker(aTUS, aYZX)
```
samestep: I have a question about this diagram: is the Substance program wrong, or does the Style program just fail to display it properly? Currently the diagram shows that `aUTS` and `aXYZ` are equal, and that `aTUS` and `aYZX` are equal, but it doesn't show that `aUTS` and `aZXY` (and thus also `aXYZ`) are equal, because the one-arc equality marker on `aUTS` gets covered by the two-arc equality marker on the same angle, leaving a lone one-arc equality marker on `aZXY`. Is this simply due to current limitations of the Style language, which we are working on resolving via support for transitivity?

packages/examples/src/geometry-domain/geometry.dsl
```diff
@@ -46,7 +46,7 @@ function Diameter(Circle c, Point p, Point q) -> Segment
 -- Unimplemented
 -- function Sum(Angle, Angle) -> Angle
 -- function Intersection(Linelike, Linelike) -> Point
--- function Altitude(Triangle, Angle) -> Segment
+-- function Altitude(Triangle, Angl/e) -> Segment
```
samestep, line 49: Stray slash?
```suggestion
-- function Altitude(Triangle, Angle) -> Segment
```

packages/examples/src/geometry-domain/textbook_problems/c11p25.sub
```diff
@@ -1,11 +1,12 @@
 Point A, B, C, D
 Circle c
 Segment AB := Chord(c, A, B)
-Segment CD := Chord(c, C, D)
+Segment DC := Chord(c, D, C)
```
samestep, line 4: I'm curious, is there a particular reason you flipped this?

packages/core/src/contrib/Functions.ts
```diff
@@ -896,12 +924,10 @@ export const compDict = {
...
     tickLength: ad.Num
   ): PathDataV<ad.Num> => {
-    if (typeof numTicks !== "number") {
-      throw Error("numTicks must be a constant");
-    }
     const path = new PathBuilder();
     // calculate scalar multipliers to determine the placement of each tick mark
-    const multipliers = tickPlacement(spacing, numTicks);
+    // TODO: why does the typecast work here?
+    const multipliers = tickPlacement(spacing, numTicks as number);
```
samestep, line 930: This typecast should not work... what happens when you remove it?

packages/core/src/contrib/ConstraintsUtils.ts
```diff
@@ -192,6 +193,42 @@ export const overlappingCircleLine = (
   return sub(d, add(r, o));
 };
 
+/**
+ * Require that circle `s1` overlaps line `s2` with some overlap `overlap`.
+ */
+export const overlappingTextLine = (
```
samestep, line 199: lmao I think I need a Penrose diagram to help me understand how this function works

packages/core/src/contrib/Functions.ts
```diff
@@ -658,6 +658,34 @@ export const compDict = {
...
     return path.getPath();
   },
+
+  repeatedArcs: (
+    _context: Context,
+    innerStart: ad.Pt2,
+    innerEnd: ad.Pt2,
+    outerStart: ad.Pt2,
+    outerEnd: ad.Pt2,
+    innerRadius: ad.Pt2,
```
samestep, line 668: Why is the radius a vector?

packages/core/src/contrib/ConstraintsUtils.ts
wodeni: Mostly copied from the signed distance functions by @joshsunshine :p. Still blocked on this until the Minkowski issue with text width/height is resolved

packages/core/src/contrib/Functions.ts
wodeni: Arcs use ellipses, so they are actually `rx, ry`

samestep: Gotcha; is the intention that you _can_ use this function to create elliptical arcs, or would it be better to change the interface to just take in an `ad.Num` (while otherwise keeping the implementation the same)?

wodeni: I don't really love the fact that we have these functions so the API is inconsequential to me. Once we have some form of collection in Style these should really go away

samestep: Sounds good!

wodeni: But yes in theory these could be defined on an ellipse with different radii

packages/core/src/contrib/Functions.ts
wodeni: That's probably it. I should add it back (or really just do numOf)

packages/examples/src/geometry-domain/textbook_problems/ex.sub
wodeni: Kind of both. A typo in the Substance program first, which accidentally made 3 angles equal to each other. The Style program doesn't handle `EqualAngleMarker` the same way as `EqualLengthMarker` (i.e. setting the mark paths s.t. that look the same). 

I fixed both of the above, but that approach is still flawed when there are `A=B=C` and `D=E` in Substance, which I don't think I can fix due to Style lang limitations.

packages/examples/src/geometry-domain/textbook_problems/c11p25.sub
wodeni: Tricky case this one: `MkQuadrilateral` don't enforce vertex ordering and I haven't found a great way to express this as a constraint. The temporary solution is to use `Parallel`, which implicitly favors one direction and thereby makes the vertices ordered correctly

PR #1127, opened 2022-10-12
fix: some symmetric predicates don't match

### LINKED ISSUE #1126: Symmetric predicate not matching
**Describe the bug**
Sometimes, symmetric predicates are not matched.

**To Reproduce**

[Here](https://penrose.cs.cmu.edu/try/?gist=b238c2fa771be44b65e5adee6d5cf4ab)

**Expected behavior**
It should print something to the canvas; instead it doesn't.

**Reason**
When the algorithm looks at Equal(x, y) , it should try both `x = A, y = B`  and `x = B, y = A` . However, the algorithm currently handles symmetric predicates by first trying for `x = A, y = B` . If it finds an exact match in Substance, which it does, it stops here and does not look for `x = B, y = A`, even though `x = B, y = A` works better with other predicates.

liangyiliang:
# Description

Resolves #1126

Some symmetric predicates do not match. For example,
```
forall Set x, y, z
where Equal(x, y); Equal(y, z) {
    Text {
        string: "Hello"
    }
}
```
should match against
```
Set A, B, C
Equal(B, A)
Equal(B, C)
```
where `Equal` is symmetric. Using mapping `x --> A, y --> B, z --> C`, the `where`-clause becomes `Equal(A, B)` and `Equal(B, C)`. `Equal(A, B)` should match against `Equal(B, A)` due to symmetry, and `Equal(B, C)` also successfully matches. But, this matching is not admitted.

The reason is that when we first consider `Equal(x, y)`, we will only consider the version that is present in the Substance program. If that does not exist, then we will consider the alternative, symmetric version. But sometimes we actively seek for the symmetric version that is _not_ present in the substance.

The fix involves always considering both versions of a symmetric predicate, regardless of whether the original version exists in the substance. Then, using Cartesian products, generate all possible matchings _under_ symmetry.

# Checklist

# Open questions

The search space blows up exponentially with the number of symmetric predicates due to additional stages of cartesian product. This problem existed prior to this fix, but with consideration of _both_ versions of symmetric predicates, the problem exacerbates.

### DIFF
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -162,13 +162,22 @@ const safeContentsList = <T>(x: { contents: T[] } | undefined): T[] =>
 
 const toString = (x: BindingForm<A>): string => x.contents.value;
 
-// https://stackoverflow.com/questions/12303989/cartesian-product-of-multiple-arrays-in-javascript
-const cartesianProduct = <T>(...a: T[][]): T[][] =>
-  a.reduce(
-    (tuples: T[][], set) =>
-      tuples.flatMap((prefix: T[]) => set.map((x: T) => [...prefix, x])),
-    [[]]
-  );
+const cartesianProduct = <Tin, Tout>(
+  t1: Tin[],
+  t2: Tin[],
+  consistent: (t1: Tin, t2: Tin) => boolean,
+  merge: (t1: Tin, t2: Tin) => Tout
+): Tout[] => {
+  const product: Tout[] = [];
+  t1.map((t1: Tin) => {
+    t2.map((t2: Tin) => {
+      if (consistent(t1, t2)) {
+        product.push(merge(t1, t2));
+      }
+    });
+  });
+  return product;
+};
 
 const oneErr = (err: StyleError): StyleDiagnostics => {
   return { errors: im.List([err]), warnings: im.List() };
@@ -979,15 +988,18 @@ const merge = (
   const s1Arr = s1.toArray();
   const s2Arr = s2.toArray();
 
-  const result: [Subst, im.Set<SubStmt<A>>][] = cartesianProduct(s1Arr, s2Arr)
-    .filter(([[aSubst], [bSubst]]) => {
+  const result: [Subst, im.Set<SubStmt<A>>][] = cartesianProduct(
+    s1Arr,
+    s2Arr,
+    ([aSubst], [bSubst]) => {
       // Requires that substitutions are consistent
       return consistentSubsts(aSubst, bSubst);
-    })
-    .map(([[aSubst, aStmts], [bSubst, bStmts]]) => [
+    },
+    ([aSubst, aStmts], [bSubst, bStmts]) => [
       combine(aSubst, bSubst),
       aStmts.union(bStmts),
-    ]);
+    ]
+  );
   return im.List(result);
 };
 
@@ -1204,13 +1216,12 @@ const matchStyArgsToSubArgs = (
   if (first !== undefined) {
     const substs: Subst[] = substsForEachArg.reduce(
       (currSubsts, substsForArg) => {
-        return cartesianProduct(currSubsts, substsForArg)
-          .filter(([aSubst, bSubst]) => {
-            return consistentSubsts(aSubst, bSubst);
-          })
-          .map(([aSubst, bSubst]) => {
-            return combine(aSubst, bSubst);
-          });
+        return cartesianProduct(
+          currSubsts,
+          substsForArg,
+          (aSubst, bSubst) => consistentSubsts(aSubst, bSubst),
+          (aSubst, bSubst) => combine(aSubst, bSubst)
+        );
       },
       first
     );
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -1167,7 +1167,7 @@ const matchStyArgToSubArg = (
 
 /**
  * Match a list of Style arguments against a list of Substance arguments.
- * @returns If all arguments match, return a `Subst` that maps the Style variable(s) against Substance variable(s). If any arguments fail to match, return `undefined`.
+ * @returns If all arguments match, return a `Subst[]` that contains mappings which map the Style variable(s) against Substance variable(s). If any arguments fail to match, return [].
  */
 const matchStyArgsToSubArgs = (
   styTypeMap: { [k: string]: StyT<A> },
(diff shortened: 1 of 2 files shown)


packages/core/src/compiler/Style.ts
```diff
@@ -1104,7 +1104,7 @@ const matchStyArgToSubArg = (
   varEnv: Env,
   styArg: PredArg<A> | SelExpr<A>,
   subArg: SubPredArg<A> | SubExpr<A>
-): Subst | undefined => {
+): Subst[] => {
```
samestep, line 1107: Could you clarify the reasoning behind this change? Is it just to make it easier to use this function later (avoiding an `if` statement)?

samestep: Or does it sometimes return an array with more than one element?

liangyiliang: It does sometimes return an array with more than one matching! This is because of possible nested symmetric predicates. In particular, when matching a StyArg against a SubArg, the arg could be a predicate application, which might involve symmetry!

liangyiliang:
> Thanks for doing this! Reviewing now. Before looking at the code though, I have a question:
> 
> > The search space blows up exponentially with the number of symmetric predicates due to additional stages of cartesian product. This problem existed prior to this fix, but with consideration of both versions of symmetric predicates, the problem exacerbates.
> 
> Could you give an example of how this is exacerbated by this PR? For example, a specific trio in our registry, comparing the size of the search space before this PR vs after?

I wan't completely aware of any specific example in the registry. But, here is one I can make up on the fly:

Suppose the Style program is
```
forall Set x, y, z
where Equal(x, y); Equal(y, z) {
    Text {
        string: "Hello"
    }
}
```
and the Substance program is
```
Set A, B, C
Equal(A, B)
Equal(B, C)
```

# Previous Algorithm
The previous code would give
```
(x, y) -> (A, B), (B, C)
(y, z) -> (A, B), (B, C)
```
Upon taking the Cartesian product, we have
```
(x, y, y, z) -> (A, B, A, B), (B, C, A, B), (A, B, B, C), (B, C, B, C)
```
(four potential mappings) of which only `(x, y, y, z) -> (A, B, B, C)` is consistent.

# New Algorithm
Under the new algorithm, we would have
```
(x, y) -> (A, B), (B, A), (B, C), (C, B)
(y, z) -> (A, B), (B, A), (B, C), (C, B)
```
The Cartesian product would give 4 * 4 = 16 potential mappings, which entails a larger search space.

packages/core/src/compiler/Style.ts
```diff
@@ -1162,47 +1162,67 @@ const matchStyArgToSubArg = (
...
-      return undefined;
-    } else {
-      return { ...rSubst, ...argSubst };
-    }
-  }, initRSubst);
-  return res;
+    return argSubsts;
+  });
+
+  // We do Cartesian product here.
```
wodeni, line 1195: Not gonna improve the computational complexity but maybe good for space complexity: instead of performing the Cartesian product right away, let `cartesianProduct` take in a lambda to filter out the inconsistent substs. My relational algebra is super rusty, but if the condition is `consistentSubsts` is this a... natural join??

packages/core/src/compiler/Style.ts
```diff
@@ -1219,50 +1239,52 @@ const matchStyApplyToSubApply = (
...
-        rSubst = matchStyArgsToSubArgs(
-          styTypeMap,
-          subTypeMap,
-          varEnv,
-          flippedStyArgs,
-          subRel.args
-        );
-      }
+
+    // Consider the symmetric, flipped-argument version
```
wodeni, line 1259: The refactoring in this function is not super clear to me. Looks like the original version already considered flipped args for symmetry. What are you changing here at a high level?

Also `rSubst1` and `rSubst2` are not very meaningful names.

wodeni (commented):
Still trying to understand how the algorithm is different from before. Had some questions below:

packages/core/src/compiler/Style.ts
liangyiliang: Yup I can explain that.

The original algorithm is:
```
rSubst1 = get subst
if (rSubst1 is empty) {
  if (isSymmetric) {
    flip arguments
    rSubst2 = get subst
  }
}
```

whereas the new algorithm is
```
rSubst1 = getSubst
if (isSymmetric) {
  flip arguments
  rSubst2 = getSubst
}
```

In particular, the original algorithm considers the symmetric version if and only if
* the predicate is symmetric, and
* the original version of the predicate does not match

On the other hand, the new algorithm considers the symmetric version if the predicate is symmetric, regardless of whether the original version of the predicate matches.

liangyiliang: Sure I can change the names

liangyiliang: Suppose a predicate is symmetric. The original algorithm checks for the symmetric version of the predicate IF the original version fails to match. The new algorithm tries the symmetric version regardless.

This is captured by the fact that we removed `if (rSubst === undefined)`

packages/core/src/compiler/Style.ts
liangyiliang: Hmmm yeah that's something I can do

packages/core/src/compiler/Style.ts
wodeni: Yeah makes a lot of sense. Basically this is the place where you got rid of the original short-circuiting. Thanks!

samestep (approved):
Looks good to me, thanks Leo! I'm approving, but did want to note that my understanding of this part of the compiler is still not very deep, so I did not actually review most of your code changes in this PR in depth.

packages/core/src/compiler/Style.ts
wodeni: Do you mind adding the filter lambda in? I think this will future-proof us from memory problems down the road even if we don't end up optimizing the matching algorithm.

liangyiliang: Oops forgot to do that - a new push will be imminent, sorry

packages/core/src/compiler/Style.ts
```diff
@@ -162,13 +162,22 @@ const safeContentsList = <T>(x: { contents: T[] } | undefined): T[] =>
...
-    [[]]
-  );
+const cartesianProduct = <Tin, Tout>(
+  t1: Tin[],
+  t2: Tin[],
+  consistent: (t1: Tin, t2: Tin) => boolean,
+  merge: (t1: Tin, t2: Tin) => Tout
+): Tout[] => {
+  const product: Tout[] = [];
+  t1.map((t1: Tin) => {
```
wodeni, line 172: you can literally do nested `for` loops here, which might be faster?

wodeni (approved):
LGTM! Just a small suggestion on `for` (or `forall`) vs. `map` below.

packages/core/src/compiler/Style.ts
liangyiliang: Just repushed using native for loops.

PR #1130, opened 2022-10-25
fix: Rename the padding argument of the overlapping constraint to overlap and negate its semantic

### LINKED ISSUE #857: Padding for overlapping constraint is unintuitive
Currently (as mentioned in https://github.com/penrose/tutorials/pull/4) the `padding` of the `overlapping` constraint must be set to a negative number in order to cause a nonnegative overlap:

https://github.com/penrose/penrose/blob/526a635d2f741f82a067365774e04e201f930f7e/packages/core/src/contrib/Constraints.ts#L186-L194

This is unintuitive; probably the sign on this parameter should be flipped. This also relates to #396; perhaps a different name besides `padding` should be used.

ZiyaoWei:
# Description

Resolves #857 

As mentioned in #857, currently the `padding` argument for the `overlapping` constraint has to be negative to work which is confusing.

Rename the `padding` to `overlap`, and negate its semantic so positive values means there's an overlap.

See the updated tutorial2

# Checklist

# Open questions

Currently this doesn't propagate the change all the way down, e.g., into the Minkowski Sum related functions. Should we also do that?

### DIFF
--- a/packages/core/src/contrib/Constraints.ts
+++ b/packages/core/src/contrib/Constraints.ts
@@ -188,7 +188,7 @@ const constrDictGeneral = {
   overlapping: (
     [t1, s1]: [string, any],
     [t2, s2]: [string, any],
-    overlap = 0.0
+    overlap: ad.Num = 0.0
   ) => {
     // Same shapes
     if (t1 === "Circle" && t2 === "Circle")
@@ -231,9 +231,9 @@ const constrDictGeneral = {
   disjoint: (
     [t1, s1]: [string, any],
     [t2, s2]: [string, any],
-    padding = 0.0
+    padding: ad.Num = 0.0
   ) => {
-    return neg(constrDictGeneral.overlapping([t1, s1], [t2, s2], -padding));
+    return neg(constrDictGeneral.overlapping([t1, s1], [t2, s2], neg(padding)));
   },
 
   /**
@@ -244,9 +244,11 @@ const constrDictGeneral = {
   touching: (
     [t1, s1]: [string, any],
     [t2, s2]: [string, any],
-    padding = 0.0
+    padding: ad.Num = 0.0
   ) => {
-    return absVal(constrDictGeneral.overlapping([t1, s1], [t2, s2], -padding));
+    return absVal(
+      constrDictGeneral.overlapping([t1, s1], [t2, s2], neg(padding))
+    );
   },
 
   /**


samestep:
> I'm not sure how deep the change should propagate down, should it go all the way to, e.g., the Minkowski levels?

Good question! I'm not sure; @joshsunshine @keenancrane any thoughts on this?

wodeni:
Also tagging @jiriminarcik. I think the constraint API change makes a lot of sense, but not sure about deeper changes in Minkowski functions.

samestep:
@ZiyaoWei Really good catch. If I understand correctly, this is because of the particular way we approximate the signed distance to a convex polygons, which is accurate for points inside the polygon and inaccurate for points outside the polygon. Then when you consider the `overlapping` constraint, adding positive padding to the example rectangles causes them to overlap with each other, so the origin lies inside their Minkowski sum, so the approximation made by `overlappingPolygons` is accurate and matches the more correct calculation made by `overlappingAABBs`; on the other hand, if you make `padding` negative, it causes the rectangles to not overlap, so the origin lies outside their Minkowski sum, so `overlappingPolygons` becomes incorrect while `overlappingAABBs` remains correct. @jiriminarcik can you confirm that this diagnosis is correct?

Anyway, long story short, no need to worry about `ConstraintsUtils.test.ts` for this PR; just flip the positive `padding` values to negative `overlap` values.

packages/core/src/contrib/Constraints.ts
```diff
@@ -233,7 +233,7 @@ const constrDictGeneral = {
     [t2, s2]: [string, any],
     padding = 0.0
   ) => {
-    return neg(constrDictGeneral.overlapping([t1, s1], [t2, s2], padding));
+    return neg(constrDictGeneral.overlapping([t1, s1], [t2, s2], -padding));
```
samestep, line 236:
```suggestion
    return neg(constrDictGeneral.overlapping([t1, s1], [t2, s2], neg(padding)));
```

packages/core/src/contrib/Constraints.ts
```diff
@@ -246,7 +246,7 @@ const constrDictGeneral = {
     [t2, s2]: [string, any],
     padding = 0.0
   ) => {
-    return absVal(constrDictGeneral.overlapping([t1, s1], [t2, s2], padding));
+    return absVal(constrDictGeneral.overlapping([t1, s1], [t2, s2], -padding));
```
samestep, line 249:
```suggestion
    return absVal(constrDictGeneral.overlapping([t1, s1], [t2, s2], neg(padding)));
```

samestep (approved):
Amazing, thank you so much for doing this @ZiyaoWei! Just a couple minor suggestions and then I think this is ready to be merged :) We really appreciate your contributions; no worries if they're sporadic, you're doing this on your own time anyways. Best of luck with your new job!

packages/core/src/contrib/Constraints.ts
```diff
@@ -181,46 +181,46 @@ const constrDictGeneral = {
...
-   * (e.g. if `s1` should be overlapping `s2` with margin `padding`).
+   * Require that shape `s1` overlaps shape `s2` with some overlap `overlap`.
+   * based on the type of the shape, and with an optional `overlap` between them
+   * (e.g. if `s1` should be overlapping `s2` with margin `overlap`).
    */
   overlapping: (
     [t1, s1]: [string, any],
     [t2, s2]: [string, any],
-    padding = 0.0
+    overlap = 0.0
```
samestep, line 191: @ZiyaoWei Good point; the fact that the `overlap` parameter on the `overlapping` constraint is of type `number` instead of `ad.Num` is actually a mistake, which you can fix via this change:

```suggestion
    overlap: ad.Num = 0.0
```

Then once you do that, it should allow you to make the other changes I mentioned above. (Actually, we really should consistently specify type `ad.Num` in all the places where we provide a numerical default value.)

samestep (changes requested):

samestep:
That sounds great @ZiyaoWei, thank you!

Issue #524, opened 2021-03-10
Headless PDF Export


maxkrieger:

wodeni:
Maybe related: https://github.com/vega/vega/tree/master/packages/vega-canvas

PR #1134, opened 2022-11-02
feat: export rasterized diagrams in `editor`

wodeni:
# Description

Related issue/PR: #524

As much as we love vectorized images, rasterized images are sometimes needed for compatibility with other tools. This PR adds support for PNG export via canvas.

* Attempted [canvg] first, but the support for `<marker>` is not great. See example: 
image
* Following the approach from [SO post](https://stackoverflow.com/questions/28226677/save-inline-svg-as-jpeg-png-svg) (thanks!), we use canvas's `drawImage` instead. 
* The exported png is true-to-scale. I did add a `scale` argument to `DownloadPng` so the UI can surface scale options in the future.
* The main catch is: __drawImage__ requires the SVG to have non-percentage width and height attributes.
  * https://bugzilla.mozilla.org/show_bug.cgi?id=700533
  * https://stackoverflow.com/questions/3768565/drawing-an-svg-file-on-a-html5-canvas 

Some example images:
wos-offcenter-estimator-walk-on-spheres(1)

tree-venn(11)
tree-venn-3d(2)
siggraph-teaser-euclidean-teaser(2)

# Checklist

- [x] I have reviewed any generated changes to the `diagrams/` folde

# Open questions

* This solution is currently client-side only. Need to see if we can use canvas as expected in `automator`, for example.

### DIFF
--- a/packages/editor/src/components/DiagramPanel.tsx
+++ b/packages/editor/src/components/DiagramPanel.tsx
@@ -41,6 +41,57 @@ export const DownloadSVG = (
   document.body.removeChild(downloadLink);
 };
 
+/**
+ * (browser-only) Downloads any given exported PNG to the user's computer
+ * @param svg
+ * @param title the filename
+ * @param width canvas and SVG width
+ * @param height canvas and SVG height
+ */
+export const DownloadPNG = (
+  svg: SVGSVGElement,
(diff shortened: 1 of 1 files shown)


maxkrieger (approved):
LGTM

Issue #1096, opened 2022-09-15
Many of our dev scripts don't work on Windows


samestep:
From https://github.com/penrose/penrose/pull/1081#issuecomment-1247468152:

1. running `yarn install-global` in `packages/roger/` on Windows does not actually put a `roger` command on the PATH
2. setting `NODE_OPTIONS='--max-old-space-size=8192'` is not valid PowerShell syntax
3. running `vite` gives the error `Error: listen EACCES: permission denied 127.0.0.1:3000`

wodeni:
>     2. setting `NODE_OPTIONS='--max-old-space-size=8192'` is not valid PowerShell syntax

`cross-env` handles this.

>     3. running `vite` gives the error `Error: listen EACCES: permission denied 127.0.0.1:3000`

Guessing this needs OS config, not just from us.

>     4. running `git clean` gives `warning: could not open directory` with the reason `Filename too long`

Weird.

samestep:
> `cross-env` handles this.

Ah cool! Easy fix, then.

> Guessing this needs OS config, not just from us.

Yeah that one is probably just weirdness on my own machine.

> Weird.

Specifically, running `git clean -dnx` gives these warnings:

```
warning: could not open directory 'node_modules/@docusaurus/core/node_modules/@babel/helper-create-class-features-plugin/node_modules/@babel/helper-compilation-targets/node_modules/@babel/helper-function-name/node_modules/@babel/types/lib/ast-types/generated/': Filename too long
warning: could not open directory 'node_modules/@docusaurus/core/node_modules/@babel/helper-create-class-features-plugin/node_modules/@babel/helper-compilation-targets/node_modules/@babel/helper-function-name/node_modules/@babel/types/lib/builders/generated/': Filename too long
warning: could not open directory 'node_modules/@docusaurus/core/node_modules/@babel/helper-create-class-features-plugin/node_modules/@babel/helper-compilation-targets/node_modules/@babel/helper-function-name/node_modules/@babel/types/lib/builders/typescript/': Filename too long
warning: could not open directory 'node_modules/@docusaurus/core/node_modules/@babel/helper-create-class-features-plugin/node_modules/@babel/helper-compilation-targets/node_modules/@babel/helper-function-name/node_modules/@babel/types/lib/constants/generated/': Filename too long
[...]

PR #1138, opened 2022-11-07
build: Use cross-env

samestep:
# Description

This PR should fix the second item from #1096, using the same technique as #497.

# Checklist

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -7,7 +7,7 @@
   "module": "dist/index.es.js",
   "scripts": {
     "watch": "vite build --watch",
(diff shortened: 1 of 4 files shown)


wodeni (approved):
Looks good. Might need to add `cross-env` in `devDependencies` in some of the packages?

samestep:
@wodeni I don't think so, right? Because it's already in `devDependencies` in the root `package.json`.

wodeni:
> @wodeni I don't think so, right? Because it's already in `devDependencies` in the root `package.json`.

Yeah makes sense. yarn workspace should take care of it hopefully.

PR #1142, opened 2022-11-11
build: Replace Turborepo with Nx

samestep:
# Description

This PR follows up on #1081 by replacing [Turborepo](https://turbo.build/) with [Nx](https://nx.dev/). The primary motivation is to make it nicer to run scripts with extra arguments, and to make the outputs of those scripts look nicer.

Because we no longer make our own custom DSL for generating the Turborepo config, this loses us a couple minor niceties. It's not that bad though, and Nx allows us to retain the property of putting script metadata right next to script definitions.

Before:
```
yarn test -- -- -- Autodiff
```

After:
```
yarn test Autodiff
```

Also, compare the output: the latter is much nicer than the former, because it doesn't prefix all the output with `@penrose/core:test:` and it preserves Jest's colors and other formatting.

# Checklist


juristr:
@samestep cool to see you adopt Nx :)

Since Penrose is an open-source project, you'd also be eligible to free remote caching with Nx Cloud. Is that something you're interested in? We can help set that up (basically changing the task runner in `nx.json`).

samestep:
@juristr Ah cool, thank you! We're not interested in remote caching at this time, but I'll definitely let you know if that changes.

juristr:
Of course! 

I'd love to get more feedback if you're willing to share.
My [Twitter DMs](https://twitter.com/juristr) are open. Would be cool to connect :)
(Also [Mastodon](https://mas.to/@juristr))

Also if there's anything Nx or Lerna related where I can help, feel free to reach out!

Issue #644, opened 2021-08-16
Turn off dragging


joshsunshine:
**Is your feature request related to a problem? Please describe.**
Dragging barely works and it consistently confuses users. 

**Describe the solution you'd like**
Disable dragging in browser-ui and on panes.penrose.ink.

wodeni:
Closing because the new IDE (https://github.com/penrose/penrose/pull/992) has dragging disabled already.

PR #1143, opened 2022-11-11
feat: "follow the cursor" dragging mode

wodeni:
# Description

Related issue/PR: #644, #1115 

The current dragging behavior is counter-intuitive. After a user drags a shape in interactive mode, the optimizer steps all variable in the state, including the one the user interacted with. If a user drags a shape from point A to B, the shape might end up in point C. This PR allows the system to programmatically specify "frozen" variable during optimization, which gives zero gradient and therefore stays during optimization.

This feature was initially developed in #1115. Since it's an experimental PR, I'm separating this feature out, which is stable and ready to be merged.

Note: this mode of optimization requires the frozen variables to be in the feasible region. In the context of dragging, if the user drags a shape off screen and the system freezes the `x, y` of the shape, the optimizer won't find a solution. To remedy this, this PR also constrain the draggable area to the canvas. 

* Add annotation to input variables
* Freeze dragged inputs in subsequent optimization
* Constrain dragging so the shapes always stay entirely on canvas: uses `getBBox` with `stroke: true` to get shape bboxes and compute offsets after `mouseDown`. Reference: https://www.petercollingridge.co.uk/tutorials/svg/interactive/dragging/

See https://github.com/penrose/penrose/pull/1107 for instructions on how to enable interactive mode.

https://user-images.githubusercontent.com/11740102/201419814-162128aa-1030-4fd3-af65-5dede682e926.mov

# Checklist

# Open questions

* More of a research question:

### DIFF
--- a/packages/core/src/engine/Optimizer.ts
+++ b/packages/core/src/engine/Optimizer.ts
@@ -109,6 +109,7 @@ const epConverged2 = (
  */
 export const step = (state: State, steps: number): State => {
   const optParams: Params = { ...state.params };
+  const { frozenValues } = state;
   const { optStatus, weight } = optParams;
   let xs: number[] = state.varyingValues;
 
@@ -135,7 +136,10 @@ export const step = (state: State, steps: number): State => {
         ...state,
         params: {
           ...state.params,
-          currObjectiveAndGradient: objectiveAndGradient(initConstraintWeight),
+          currObjectiveAndGradient: objectiveAndGradient(
+            initConstraintWeight,
+            frozenValues
+          ),
           weight: initConstraintWeight,
           UOround: 0,
           EPround: 0,
@@ -246,7 +250,8 @@ export const step = (state: State, steps: number): State => {
         optParams.UOround = 0;
 
         optParams.currObjectiveAndGradient = optParams.objectiveAndGradient(
-          optParams.weight
+          optParams.weight,
+          frozenValues
         );
 
         log.info(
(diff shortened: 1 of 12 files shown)


packages/core/src/types/state.ts
```diff
@@ -25,6 +25,7 @@ export interface State {
   canvas: Canvas;
   computeShapes: ShapeFn;
   params: Params;
+  frozenValues: number[];
```
samestep, line 28: Could you add a comment about what this means? Are these indices into the `varyingValues` array?

packages/core/src/engine/Optimizer.ts
```diff
@@ -815,7 +816,14 @@ export const genOptProblem = (
       gradf: xs.map((x, i) => {
         // fill in any holes in case some inputs weren't used in the graph, and
         // also treat pending values as constants rather than optimizing them
-        return i in gradient && !("pending" in inputs[i]) ? gradient[i] : 0;
+        if (!(i in gradient)) {
+          return 0;
+        } else {
+          const meta = inputs[i];
+          return meta.tag === "Optimized" && !frozenVaryingValues.includes(i)
```
samestep, line 823: Actually, given that we use `includes` here, why not make `frozenVaryingValues` a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instead of an array?

packages/core/src/renderer/dragUtils.ts
```diff
@@ -12,42 +13,50 @@ export const dragUpdate = (
...
+  const { inputs, constrFns, objFns } = state;
   const updated: State = {
     ...state,
-    params: { ...state.params, optStatus: "NewIter" },
+    params: genOptProblem(
+      inputs,
+      objFns.map(({ output }) => output),
+      constrFns.map(({ output }) => output),
+      frozenVaryingValues
+    ),
```
samestep, line 31: I'd rather not call `genOptProblem` on every drag, because that calls `makeGraph` and `genCode`, both of which are expensive. Instead of doing this, could you pass `frozenValues` from the state to the optimizer when you run it? There shouldn't be any need to recompile, because the computation graph itself remains exactly the same after a drag.

PR #1146, opened 2022-11-17
refactor: `numsOf` for evaluating autodiff numbers

wodeni:
# Description

Related issue/PR: #907 

Aside from the main optimizer, other parts of the system sometimes need to get the numerical values of autodiff nodes. Notably, `convexPartitions` needs to evaluate points in a polygon before the optimizer runs. Similarly, any computation in `Function.ts` that needs JS `number` from `ad.Num` will need to run such evaluation. Currently, we go through these steps to achieve this:

1. Construct a `secondaryGraph` with `inputs` that are sources of the output nodes of interest
2. Run `genCode` on the graph, and evaluate the generated code with the input values.  

This PR modifies `numsOf` and `numOf` in `contrib/Util.ts` to perform 1 so we can just call `numsOf(xs)` or `numOf(x)` without doing the steps manually.

# Checklist

### DIFF
--- a/packages/core/src/contrib/Utils.ts
+++ b/packages/core/src/contrib/Utils.ts
@@ -189,12 +189,24 @@ export const closestPt_PtSeg = (
   return ops.vadd(start, ops.vmul(t1, dir));
 };
 
-export const numsOf = (xs: ad.Num[]) => {
+/**
+ * Get numerical values of nodes in the computation graph. This function calls `secondaryGraph` to construct a partial computation graph and run `genCode` to generate code to evaluate the values.
+ *
+ * @param xs nodes in the computation graph
+ * @returns a list of `number`s corresponding to nodes in `xs`
+ */
+export const numsOf = (xs: ad.Num[]): number[] => {
   const g = secondaryGraph(xs);
+  const inputs = [];
+  for (const v of g.nodes.keys()) {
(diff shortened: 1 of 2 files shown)


wodeni:
One question: does it matter if we give the generated function more inputs than it needs? My assumption is it'll just ignore the additional inputs?

packages/core/src/contrib/Utils.ts
```diff
@@ -189,12 +189,24 @@ export const closestPt_PtSeg = (
   return ops.vadd(start, ops.vmul(t1, dir));
 };
 
-export const numsOf = (xs: ad.Num[]) => {
+/**
+ * Get numerical values of nodes in the computation graph. This function calls `secondaryGraph` to construct a partial computation graph and run `genCode` to generate code to evaluate the values.
```
samestep, line 193:
```suggestion
 * Get numerical values of nodes in the computation graph. This function calls `secondaryGraph` to construct a partial computation graph and runs `genCode` to generate code to evaluate the values.
```

samestep (approved):
Looks great, thank you! One typo fix in the docstring.

samestep:
> One question: does it matter if we give the generated function more inputs than it needs? My assumption is it'll just ignore the additional inputs?

Correct.

PR #1147, opened 2022-11-17
feat: detect and report cyclic assignments in Style

wodeni:
# Description

Related issue/PR: #760 

Style doesn't allow the user to assign values cyclically. For instance, `a = a + 1` is not a valid Style expression. Currently, the system fails silently whenever that happens. This PR adds an explicit check and report such errors in a human-readable way.

* Accumulate source location when building up the dependency graph in the first compilation pass
* Retrieve the source location and pretty-print them in error reporting

TODO

# Checklist

# Open questions

Generally the Style compiler doesn't keep source locations around for long. Is there a more principled way to keep track of them?

### DIFF
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -2881,7 +2881,6 @@ export const translate = (
     })
   );
   if (cycles.length > 0) {
-    // TODO: improve the error
     return {
       ...trans,
       diagnostics: oneErr({ tag: "CyclicAssignmentError", cycles }),
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -1745,15 +1745,7 @@ const insertExpr = (
   updateExpr(path, assignment, "Assign", (field, prop, fielded) => {
     const warns: StyleWarning[] = [];
     if (prop === undefined) {
-      const source = processExpr(
-        {
-          ...block,
-          start: expr.start,
-          end: expr.end,
-          locals: assignment.locals,
-        },
-        expr
-      );
+      const source = processExpr({ ...block, locals: assignment.locals }, expr);
       if (source.isErr()) {
         return err(source.error);
       }
@@ -1765,7 +1757,7 @@ const insertExpr = (
       if (expr.tag === "GPIDecl") {
         return err({ tag: "NestedShapeError", expr });
       }
-      const shape: FieldSource | undefined = fielded.get(field);
+      const shape = fielded.get(field);
       if (shape === undefined) {
         return err({ tag: "MissingShapeError", path });
       }
@@ -1779,12 +1771,7 @@ const insertExpr = (
         dict: fielded.set(field, {
           ...shape,
           props: shape.props.set(prop, {
-            context: {
-              ...block,
-              start: expr.start,
-              end: expr.end,
-              locals: assignment.locals,
-            },
+            context: { ...block, locals: assignment.locals },
             expr,
           }),
         }),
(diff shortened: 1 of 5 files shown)


joshsunshine:
I've never head the phrase "cyclic assignment" before? Is this a standard name for statements like `x = x + 1`?

wodeni:
> I've never head the phrase "cyclic assignment" before? Is this a standard name for statements like `x = x + 1`?

Feel free to suggest a name! Don't really know a standard name. "In-place" may be reasonable but a bit jargon-y. The error message does print out all cycles, so perhaps that helps with readability.

samestep:
> Generally the Style compiler doesn't keep source locations around for long. Is there a more principled way to keep track of them?

I would say this isn't really true; if I remember correctly, while working on #984 I put a lot of effort into using concrete syntax nodes wherever possible and passing those all the way through the compiler, as well as adding source location info to other places as well (this is why pretty much all the existing error messages print source location info now). But I'm sure there is room for improvement; not sure what exactly a more principled way would look like.

packages/core/src/utils/Error.ts
```diff
@@ -433,7 +442,8 @@ canvas {
 };
 
 const showCycles = (cycles: string[][]) => {
-  const pathString = (path: string[]) => path.join(" -> ");
+  // repeats the cycle start again
+  const pathString = (path: string[]) => [...path, path[0]].join(" <-> ");
```
samestep, line 446: Could you explain the reasoning behind changing this from a single arrow to a double arrow? The former feels more accurate to me.

samestep:
@wodeni I rewrote your code to make use of the `SourceRange` info that was already present instead of adding it to the `Context` type, and also added a bit more code to handle the case where the node is `ShapeType | undefined` and not `WithContext<NotShape>`, instead of just casting. Does this look reasonable? If so then this looks good to me after removing the TODO comments.

packages/core/src/types/errors.ts
```diff
@@ -356,6 +357,12 @@ export interface CanvasNonexistentDimsError {
   type?: Expr<A>["tag"];
 }
 
+export interface CyclicAssignmentError {
+  tag: "CyclicAssignmentError";
+  // TODO: improve types, currently the generated id and source location
```
samestep, line 362: Probably remove this, unless you had a specific improvement in mind?

```suggestion
```

packages/core/src/compiler/Style.ts
```diff
@@ -2867,6 +2867,27 @@ export const translate = (
...
+        id,
+        src:
+          e === undefined || typeof e === "string"
+            ? undefined
+            : { start: e.expr.start, end: e.expr.end },
+      };
+    })
+  );
+  if (cycles.length > 0) {
+    // TODO: improve the error
```
samestep, line 2884: Same here; unless you had something specific in mind

```suggestion
```

samestep (approved):

packages/core/src/utils/Error.ts
wodeni: mostly trying to see how it'd look on the screen. the reasoning is that you can go in any direction because it's a cycle

wodeni:
> @wodeni I rewrote your code to make use of the `SourceRange` info that was already present instead of adding it to the `Context` type, and also added a bit more code to handle the case where the node is `ShapeType | undefined` and not `WithContext<NotShape>`, instead of just casting. Does this look reasonable? If so then this looks good to me after removing the TODO comments.

Oh thanks! Yeah looks reasonable to me. I wasn't super sure about how much info `DepGraph` stores. The memory of reviewing #984 got a little blurry :P.

packages/core/src/types/errors.ts
wodeni: I was hoping to make `cycles: Expr<C>[][]`, but I guess there might be some exceptions to this?

packages/core/src/utils/Error.ts
samestep: Ah gotcha; that's not true though, like if you have the cycle `1 -> 2 -> 3 -> 1`, that's different from the cycle `1 -> 3 -> 2 -> 1`. Is it OK if we change it back?

packages/core/src/types/errors.ts
samestep: Ah gotcha; yeah we can't actually do that, because some `DepGraph` nodes represent shapes rather than shape properties, and in general there isn't a single expression representing a shape, because `override` allows people to replace a single property of a shape while leaving the rest of the shape unchanged:

https://github.com/penrose/penrose/blob/654571a52fd347c4078db95da277e7c231bacbf6/packages/core/src/types/styleSemantics.ts#L149-L154

samestep:
@wodeni feel free to merge this!

PR #1165, opened 2022-12-09
build: Don't use estrella to build core

samestep:
# Description

While looking at #423, @wodeni and I found that esbuild has an [analyze](https://esbuild.github.io/api/#analyze) option which is not supported by Estrella. Thus, we're removing Estrella as a dependency and instead just using esbuild directly.

# Checklist


samestep:
@wodeni I think this may have broken `core:watch`? Should be fixable by using esbuild's [watch](https://esbuild.github.io/api/#watch) option, but I forgot to do that in this PR.

Issue #430, opened 2020-12-16
Initial parameters for some graphical primitives are surprising


keenancrane:
Currently we initialize some parameters of graphical primitives to random values.  E.g., the declaration

```
   g.shape = Line {
      start: p.dotShape.center
        end: q.dotShape.center
   }
```

gives the line a random thickness and color.  This behavior sort of violates the "principle of least surprise," and makes the programmer wonder if they maybe did something wrong, or if there's some bug in the compiler/renderer that's producing these funky values.

**Proposal:** Just initialize the fields to some sensible but deterministic value.

wodeni:
Addressed by #1169.

PR #1169, opened 2022-12-09
refactor: unify default strokeColor for outline shapes

wodeni:
# Description

Related issue/PR: #430 

This PR makes `#000` the default `strokeColor` for "outline" shapes such as `Line`, `Path`, and `Text`. `Path` is consider an outline shape out of uncertainty that the path might not be closed. 

* Change default props in respective shape files

See deployment `allShapes-allShapes` example.

# Checklist

# Open questions

Looks like this impacted variations of many diagrams. Why? (cc: @samestep) Looks like even when the Style props remain the same, as long as there's a change to the defaults, the resulting diagrams will change.


samestep (approved):
Sorry for taking so long to review this! The code changes look great. For the changes to the generated `diagrams/` we might want to update some `variation`s though... would be nice if Penrose had a way to generate a grid of different variations to let you quickly pick one that looks good.

wodeni:
> Sorry for taking so long to review this! The code changes look great. For the changes to the generated `diagrams/` we might want to update some `variation`s though... would be nice if Penrose had a way to generate a grid of different variations to let you quickly pick one that looks good.

Gotta merge #1287!

samestep (approved):
Thanks for updating the registry diagram variations! I'm good to merge this, but just for future reference, I did want to note the few specific diagrams that still seem to have meaningful changes by this PR:

- `curve-examples/cubic-bezier.svg` (looks upside down now?)
- `geometric-queries/closest-point/test-group.svg` (now the circle and the rectangle overlap)
- `geometric-queries/closest-point/test.svg` (now the polyline and the triangle overlap)
- `geometry-domain/siggraph-teaser.svg` (now the angle $qpr$ is a bit narrow)
- `graph-domain/textbook/sec1/fig10.svg` (the graph is planar but now not drawn so)
- `graph-domain/textbook/sec1/fig12.svg` (the labels RRP4 and RRP41 look ambiguous now)
- `graph-domain/textbook/sec1/fig13.svg` (the Team 2 and Team 5 labels look weird)
- `graph-domain/textbook/sec1/fig8b.svg` (a lot of the labels are now clumped)
- `graph-domain/textbook/sec2/fig11c.svg` (the graph is planar but no longer drawn so)
- `walk-on-spheres/nested-estimator.svg` (the spheres are now all clumped together on one side)

To clarify, I don't even necessarily think that it's worth changing the variations in this PR; I'm mostly just writing these down to note the sorts of ways that diagrams can qualitatively change when we update the variation. I think it's an interesting research problem to think about ways that we can make it more manageable to maintain our registry going forward under this sort of change.

PR #1170, opened 2022-12-09
build: Simplify core build

samestep:
# Description

Resolves #1166 and https://github.com/penrose/penrose/pull/1165#issuecomment-1344512716 by replacing our `packages/core/build.js` script producing two bundles with a much simpler one-line esbuild command that just produces one bundle, allowing us to just tack on `--watch`.

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new ESLint warnings
- [x] I have reviewed any generated changes to the `diagrams/` folder

PR #1160, opened 2022-12-08
build: Switch from CJS to ESM

samestep:
# Description

While working on #1092, I realized that it would be very nice to have access to [top-level await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await). However, that is only supported by ESM and not CJS, so I decided to first open this PR to migrate us to ESM.

# Implementation strategy and design decisions

This was mostly just a lot of trial-and-error, digging through documentation, and playing with config options until it worked. I got a lot of help from the following threads:

- https://github.com/kulshekhar/ts-jest/issues/2057
- https://github.com/nodejs/node/issues/32137
- https://github.com/vitejs/vite/issues/6985
- https://github.com/evanw/esbuild/pull/2067
- https://github.com/TypeStrong/ts-node/issues/1007

By volume, most of the code changes in this PR are just switching imports of CJS modules to only use the default import rather than trying to destructure it, because apparently the latter is not actually supported when importing a CJS module from an ESM module.

One other change: because we need the [`global-jsdom`](https://www.npmjs.com/package/global-jsdom) import to come first in `automator` and `synthesizer`, I upgraded us to TypeScript 4.9.4 and [prettier-plugin-organize-imports](https://github.com/simonhaenisch/prettier-plugin-organize-imports) 3.2.1, to give us grouped imports which were added in [TypeScript 4.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#group-aware-organize-imports). These upgrades caused Prettier to start trying to remove a bunch of React imports, so I added `"organizeImportsSkipDestructiveCodeActions": true` to our `.prettierrc`.

# Examples with steps to reproduce them

I manually verified that this works by running `yarn start` with a top-level `await` in `packages/core/src/index.ts`, but removed that before submitting this PR because we don't actually use it yet. Note that to get that to actually work, you have to also add https://github.com/Menci/vite-plugin-top-level-await as a dependency to all our Vite packages.

<details>
<summary>Expand this to get a concrete diff you can apply to test top-level await.</summary>

```diff
diff --git a/packages/components/package.json b/packages/components/package.json
index 9fb489dcb..4dfd5e83e 100644
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -73,6 +73,7 @@
     "@types/react-dom": "^17.0.10",
     "@vitejs/plugin-react": "^1.0.7",
     "babel-loader": "^8.2.5",
-    "vite": "^2.7.2"
[...]

PR #1173, opened 2022-12-14
build: don't bundle core

### LINKED ISSUE #423: Reduce bundle size of `@penrose/core`

samestep:
# Description

We don't need to, as far as I can tell; anything using `@penrose/core` that wants to bundle it should be able to just bundle it directly, so there's no reason to put all our dependencies into the bundle when they're already dependencies of the package itself. This is a followup to #1160, #1165, and #1170. Closes #423.

To increase consistency with our other packages, I changed the `core` build directory from `build/dist/` to just `dist/`. Previously `packages/synthesizer-ui/README.md` also erroneously said it used `build/` instead of `dist/`, so I corrected that here too.

# Checklist


samestep:
Currently this gives the following error in `editor`:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'prototype')
    at sax.js:221:46
    at ../../node_modules/xmldoc/node_modules/sax/lib/sax.js (sax.js:1574:1)
    at __require2 (chunk-5537U2Q2.js?v=ad4963f6:47:50)
    at xmldoc.js:7:9
    at ../../node_modules/xmldoc/lib/xmldoc.js (xmldoc.js:272:1)
    at __require2 (chunk-5537U2Q2.js?v=ad4963f6:47:50)
    at ../../node_modules/xmldoc/index.js (index.js:3:18)
    at __require2 (chunk-5537U2Q2.js?v=ad4963f6:47:50)
    at ../../node_modules/svg-flatten/src/parse.js (parse.js:1:14)
    at __require2 (chunk-5537U2Q2.js?v=ad4963f6:47:50)
```

wodeni (approved):
Looks good! With this and #1207 merged both `core` and `components` would be bundleless!

re `yarn start`: it's still quite useful to run multiple builds in watch mode. I think we can avoid `tsc-alias` by using `ttypescript`. I can push a commit for that soon.

PR #1119, opened 2022-10-06
refactor: restore the synthesizer and `synthesizer-ui`

wodeni:
# Description

Related issue/PR: #984

After much system improvements, `synthesizer` and `synthesizer-ui` (i.e. edgeworth) are a bit out of date. This PR contains some bug fixes and UI improvements.

* Use `Listing` for the Substance file listing in `synthesizer-ui`
* Synthesizer bug fixes
  * Respect subtyping rules when generating arguments to predicates
  * Fix wrong context passing here and there

# Implementation strategy and design decisions

* Bug fixes are pretty straightforward. For the predicate args issue, I needed a function that finds supertypes from `core`. Ended up using `Digraph` in `Util` for the type graph in `Env` so we have better types for the graph operations 

# Examples with steps to reproduce them

`yarn turbo run start --filter=synthesizer-ui` should run smoothly now!

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new ESLint warnings
- [x] I have reviewed any generated changes to the `diagrams/` folder

PR #1199, opened 2023-01-06
feat: support layout stages in Style

### LINKED ISSUE #766: Perform optimization in user-specified stages
**The problem.**

This issue is a fairly major one, which affects the ability of Penrose to scale to more complex diagrams—not simply at a performance level, but (just as importantly) at a user cognitive level.  There are some big design questions to answer here, but it's something we really do need to think about if we want to scale to the complexity of a lot of real-world diagrams, and to "lower the floor" on what we can expect from Style programmers.

The current paradigm in Style is that the programmer specifies a collection of constraints, via the `ensure` keyword, and objectives, via the `encourage` keyword.  The layout engine then solves a single global optimization problem to try and find a good local minimizer.  There are two basic challenges:

1. **Optimization starts to get expensive.**  As with any optimization problem, the more variables, constraints, and objective terms there are, the harder the machine has to work to evaluate each optimization step.  However, this isn't (yet) our biggest bottleneck.  Our optimization scheme and implementation works pretty well, even for reasonably large problems (at least, once the computation graph has been assembled—see the lingering perf bug in #652).
2. **Constraints and objectives get hard to design.** The much bigger issue, which is more fundamental to the design of our system, is that at some point it becomes hard to strike a balance between many competing constraints and objectives.  Anyone who has ever worked with continuous optimization knows the challenge: you want to simultaneously optimize f(x) and g(x), but if f and g differ on what they consider "good" states x, then they will effectively compete—usually reaching a solution that nobody likes.

image

In the context of Style, the programmer tries to work around this challenge by adjusting the relative weights of various penalties.  However, this approach requires significant experience with/intuition about optimization problems, and often fails to produce a satisfactory solution.  Worse, the solution is brittle: it may work well for the one or two Substance programs used to test the Style, but it's very hard to say whether it will continue to work well for future Substance programs.

**Examples.**

One good example is the [triangle-mesh-2d](https://github.com/penrose/penrose/tree/trimesh2d/examples/triangle-mesh-2d) domain, which draws little 2D diagrams of triangle meshes with labels, as might be used in a geometry processing paper:

image

If labels are omitted, it's pretty easy to get something working that draws reasonable-looking triangles, edges, and vertices.  For instance, here are the first five random samples with labels turned off:

image

If we now add labels, the optimizer starts to struggle—although these diagrams still look "neat," pay close attention to where the vertex labels `i`, `j`, `k`, `l`, and `m` appear:

image

Notice in particular that in about half the cases, labels are placed in ways that make the diagram hard or impossible to interpret.  Formally, however, these diagrams satisfy all the given constraints:

- no label is intersected by an edge
- no two labels overlap
- etc.

They also try as hard as possible to minimize the given objectives, e.g.,

- keep the label of a vertex close to the "dot" for that vertex.

In other words, the problem is not with the layout engine—it is with the design of the Style program.  At this point we could "blame the programmer" (in this case, me), but the programmer has already worked very hard to get the Style even to this state, leaning heavily on past expertise coding graphics and optimization problems.  We need a better solution.

One key observation is that, in practice, many diagrams are not laid out "all at once."  Consider for instance this reaction diagram, which comes from the `structural-formula` example:

image

Amazingly enough, the Penrose layout engine *does* manage to lay out this diagram by solving one big optimization problem.  But it was a real struggle to write the Style program that makes this happen—requiring a lot of experimentation with different versions of constraints, different weighting functions, etc.  In the end, the results are still not entirely satisfactory.  For instance, the program "encourages" water molecules to have a bond angle of 104.5 degrees, and CO2 molecules to have a bond angle of 180 degrees.  But because of interactions with other elements of the diagram, these molecules are (literally) bent out of shape.

Yet there's little good reason to optimize the diagram all at once.  A reasonable strategy would be:

- **Pass 1.** First optimize the shape of individual molecules.
- **Pass 2.** Next, keep the individual molecule configurations "frozen" and just optimize their arrangement within each gray box (stil letting the box size vary within some bounds, to acommodate the molecule shapes).
- **Pass 3.** Next, arrange the boxes and large titles on the canvas.
- **Pass 4.** Finally, optimize the location of small per-molecule labels.

One might argue that there is a missed opportunity here: perhaps an *even better* diagram could be found by letting the molecule shapes and box arrangements adjust themselves according to the label placement, so that all diagram elements ultimately live in perfect harmony.  And that's true... if you have an oracle (or solver) that can do a great job of handling the whole diagram all at once.  But, once again, it's not just a question of efficiency: *designing* constraints and objectives becomes (very) hard when you're laying out the diagram all at once, because everything affects everything.

**The proposal.**

In a perfect world, we would have an oracle that finds an awesome solution to every constrained optimization problem, ever—*and* a language design that makes it easy to specify a ton of simultaneous constraints/objectives and still get the "expected result."  These are lofty goals (and great research problem) that we can continue to strive for.  In the mean time, there are some pragmatic solutions that have to do with the design of the Style language.

The key observation, again, is that real layout problems are often solved by laying out individual pieces, freezing these pieces, then optimizing the arrangement of the pieces.  Beyond diagram layout, this is the way microprocessor design is often done.  For instance, in a multicore architecture you do layout for a single core, then (more or less…) repeat that core many times.  You certainly don't do a full custom layout for all cores simultaneously—that would be nuts.

Likewise, **the proposal for Style is to add facilities for the Style programmer to specify optimization passes.**

More specifically, a Style program would associate each `ensure` statement, `encourage` statement, and optimized variable with a "pass" (as in the combustion example above).  These passes would be given a definite ordering.  _Each pass effectively determines a different, independent optimization problem._ The layout engine would then solve these problems sequentially: solve the first pass.  Use this solution as the initial state for the next pass.  Repeat.

This setup enables Style programmers to design one pass at a time.  Once they're happy with the first pass, they can move on to a subsequent pass knowing that subsequent objectives and constraints won't mess with the beautiful results they achieved in the first pass.  As a side benefit, it reduces the size/cost of each optimization problem, meaning that they will likely get quicker feedback on design iterations.

---

A key remaining question is: how should optimization passes be specified at the Style language level?

**Implicit Passes.** One possibility is to use syntax reminiscent of the current `layer` paradigm, where objects are determined by a *relative ordering*.  For instance, one might write

```
constraint x.onCanvas = ensure contains( Global.bbox, x.icon )
encourage near( x.icon, x.label ) after x.onCanvas
```

These statements implicitly define two passes: one that enforces the `contains` constraint, and another that tries to minimize the `near` objective.  It also implies that the former should be optimized before the latter.  However, much like layering, keeping track of these passes in one's head becomes very challenging after a while.  Also, this specification makes it difficult to explore different orderings for the optimization passes.

**Named Passes.** Another possibility is to explicitly declare passes, and give them an explicit ordering.  Individual constraints and objectives can then be added to these passes.  For instance, at the top of a Style program (where programs already often have some kind of `Global` or `const` block) one might write

```
Global {
   pass shapePass -- this pass lays out the basic shapes
   pass labelPass -- this pass positions labels on top of the fixed shapes
   perform shapePass before labelPass
}
```

In subsequent rules, one can then refer to these passes:
```
forall Type x; Type y
where Predicate(x,y) {
   ensure near( x.icon, y.icon ) in Global.shapePass
}
```
This mechanism is both easier to follow (because there's a name that hints at the purpose of each pass), and also allows one to easily experiment with different optimization "schedules", e.g., `perform pass1 before pass2` versus `perform pass2 before pass1`.

We also need a way to specify which values are variable vs. fixed (i.e., "frozen") in each pass.  Currently we use the `?` syntax for this purpose, which we can again extend using the `in pass` mechanism:

```
forall Type x {
   x.center = (?,?) in Global.pass1
   x.center = fixed in Global.pass2
}
```

Here `fixed in <pass>` indicates that, upon reaching the pass `<pass>`, `x.center` should not be optimized, but rather should keep whatever values it was assigned in the previous pass.  This kind of mechanism seems necessary, because by default we assume that all variables that are not given *constant* values should be optimized.  There are some further semantics to work out here—for instance, if `x.center` is fixed in `pass2`, and then a subsequent `pass3` comes after `pass2`, should `x.center` remain fixed?  Or should it now *automatically* be variable again?  Etc.  These are not "deep" design questions, but do need to be thoughtfully worked out.

---

**Implementation.**

It's reasonable to ask whether staged optimization will be a nightmare to implement.  I don't think it has to be.  Roughly speaking, "all" we have to do is parse the Style program once for each pass, and ignore any directives that don't apply to that pass.  Otherwise, we build the computational graph exactly as we always do, and solve the optimization problem exactly as we always do.  Some very minor glue is needed to move from one stage to the next—basically assigning constant values to any "frozen" variables, by simply copying in the optimized values from the previous pass.

---

On the whole there's a fair bit to discuss here in terms of both design and implementation.  But on the whole extending Style to allow multiple optimization passes seems well within reach, within the existing design and implementation of our system.  And would go a long way toward improving scalability, both in temrs of system performance, but (more importantly) in terms of programmers' cognitive ability to write Style programs.

wodeni:
# Description

Related issue/PR: #1115, #1192 
Resolve #766

This PR includes a more general version of staged layout with a preliminary Style syntax. The user can annotate varying values (i.e. `?`), constraints, and objectives with layout stages. At layout time, the optimizer executes layout stages in order. For instance, the user may want to layout labels before laying out shapes (e.g. #1115), and using the feature, they can separate varying values and constraints into distinct sets to be optimized in stages. 

## Preliminary syntax

The layout stages are optionally defined at the top level in Style: 

```
layout = [ShapeLayout, LabelLayout, Overall]
```

If undefined, `layout` gets a default anonymous stage, i.e. the default mode where all inputs and constraints are jointly optimized at once. 

In Style blocks, inputs and constraints can be associated with stages via the `in` keyword:
```
forall Object o {
  o.shape = Square {
    center : (? in ShapeLayout, ? in ShapeLayout) -- `in` can be used after any `?`
    w: ? in [ShapeLayout, Overall] -- use parens after `in` for multiple stages
  }
  o.text = Text {
    center: (? in LabelLayout, ? in LabelLayout)
  }
  ensure contains(o.shape, o.text) in (LabelLayout, Overall)
  encourage equal(o.shape.center[1]. o.text.center[1]) in LabelLayout
  -- if let unspecified, the constraint/objective/input automatically participate in all stages, 
  -- i.e. [ShapeLayout, LabelLayout, Overall]
  ensure onCanvas(o.shape)  
}
```

* #1119 introduced masks for inputs, constraints, and objectives
* Add `stages` to AST nodes
* Add `in` syntax for varying values and constraints
* Add `layout = [Stage1, Stage2, ...]` syntax at the top level
* Process stage pipeline and annotations in the Style compiler

https://8a4052a7.penrose-72l.pages.dev/try/?gist=73da256d6f5b9f1217b257c66fe453c4

# Checklist


samestep (approved):
LGTM modulo error handling in `getLayoutStages` :)

joshsunshine:
@wodeni Can you please add a registry example or two that demonstrates/tests the features of this change before it is merged?

keenancrane:
What stage(s) include variables/objectives/constraints that are not assigned an explicit stage? All of them?

samestep:
> What stage(s) include variables/objectives/constraints that are not assigned an explicit stage? All of them?

Yeah, currently all of them.

PR #1206, opened 2023-01-10
feat: Functions/Constraints for curves

jiriminarcik:
# Description

Added the following **constraints**:
- isLocallyConvex: the shape should be locally convex (all angles between consecutive edges would have the same sign)
- isConvex: the enclosed area should be convex (implemented as "isLocallyConvex and has turning number ±1")
- isEquilateral: all edges should have the same length
- isEquiangular: all angles between consecutive edges should be equal

and **functions**:
- totalCurvature: intergral of curvature over the curve
- elasticEnergy: intergral of curvature squared over the curve
- isoperimetricRatio: perimeter squared divided by the area
- perimeter: perimeter of the curve
- turningNumber: index of regular homotopy class
- signedArea: signed enclosed area

and **objectives**:
- isRegular: the shape should be regular (equiangular and equilateral)

Example are in folder `packages/examples/src/curve-examples` with trio `cubic-bezier.sub`, `cubic-bezier.sty` and `curves.dsl` added to the registry.

# Checklist

# Open questions

Do not know any elegant way implement (in a differentiable way) the following constraints/objectives:
- Ensure the curve is simple
- Prescribe a specific number of self-intersections
- Prescribed number of inflection points

Other possible constraints/objectives/functions that can be implemented in future:
- Winding number


keenancrane:
Can we change `isEquidistant` to `isEquilateral`?  The latter is more understandable, because this constraint has to do with edge length (rather than, say, distance to a fixed point).

keenancrane:
Can we change `algebraicArea` to `signedArea`?  Just a little less obtuse (even if "signed area" still requires some understanding…)

keenancrane:
Small thing: the PR says
```
turningNumber: prescribed regular homotopy class
```
but perhaps in any future documentation it should just be
```
turningNumber: index of regular homotopy class
```
In other words, the `turning number` function doesn't prescribe anything; only if someone says `ensure equal( turningNumber(x.curve), 0 )`.  (<— which is a super cool example if it works!)

keenancrane:
Seems maybe natural to combing `isEquilateral` and `isEquiangular` into `isRegular`, i.e., same angles and same side lengths (but specific side lengths, as well as rigid motion undetermined).  This of course just gives you a regular n-gon, but could be interesting if you want to say `encourage isRegular` rather than `ensure isRegular`.

keenancrane:
Overall this is super cool!  Can't wait until we do the 3D extension. ;-)

keenancrane:
The PR mentions that perimeter is only an approximation for Bézier curves.  Does this mean that other quantities, like elastic energy, are not approximations?  For instance, the bending energy for a cubic Bézier curve is the integral of a rational function (degree 2 over degree 4).  I'm not sure I know how to integrate that exactly.

packages/examples/src/curve-examples/cubic-bezier.sty
```diff
@@ -0,0 +1,68 @@
...
+
+    points = [c.p1, c.p2, c.p3, c.p4]
+    shape curve = Path {
+        d: cubicCurveFromPoints("open", points)
+        strokeWidth: 3.5
+        ensureOnCanvas: true
+    }
+
+    ensure equal(perimeter(points, false), 900)
+    ensure equal(algebraicArea(points, false), 1e5)
```
wodeni, line 24:
```suggestion
    ensure equal(signedArea(points, false), 1e5)
```

jiriminarcik:
> The PR mentions that perimeter is only an approximation for Bézier curves. Does this mean that other quantities, like elastic energy, are not approximations? For instance, the bending energy for a cubic Bézier curve is the integral of a rational function (degree 2 over degree 4). I'm not sure I know how to integrate that exactly.

No, in this PR everything is just polygons visualised as curves. Will improve it in future PRs, wanted to do the low-hanging fruit first..

PR #1207, opened 2023-01-10
feat: run optimizer in Web Worker   

### LINKED ISSUE #1095: Providing a positive user experience for slow Penrose trios
When selecting a long-running and complex Penrose trio like the `hypergraph-hypergraph` from the example tab in the IDE, the browser window hangs and shows "Loading example..." for several seconds (or more) while the trio compiles and the optimizer runs. This is not a good user experience. The user can't edit the trio while this is happening and has no sense of progress toward completion. 

The state of the system should be constantly visible to the user. To enable that, I propose the following process:
1. Load the trio into the code window tabs before starting any compilation and optimization so at least the user can see the code of the example immediately.
2. Show a progress bar in the diagram window during compilation so the user can see that progress is being made.
3. After compilation, show the first sample in the diagram window.
4. By default, dynamically change the step size so the user can observe optimization progress as it happens. Perhaps the diagram view should change at least 5 times per second. (I chose the value 5 arbitrarily. Feel free to alter it.)

In addition, the compilation and optimization should not freeze the rest of the browser window. A user should be able to interact with the code, switch to another example, etc. during those processes. I suspect the best option to allow for this interaction is to use [Web Workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers) to run compilation and optimization.

### LINKED ISSUE #1194: `components` type declarations get overwritten by Vite build
In `packages/components/`, running `yarn build-decls` generates a bunch of `dist/**.d.ts` and `dist/**.d.ts.map` files, which are then deleted by `yarn build`. We should fix this, either by generating type declarations into a different directory from Vite's build directory, or by generating the `dist/**.js` files via `tsc` instead of `vite build`. The latter may be preferable, as it could possibly allow us to improve our downstream build times by not repeatedly bundling dependencies such as Monaco.

wodeni:
# Description

Resolves #1194 
Resolves #1095 

This PR is a follow-up on #1092 to run our WASM optimizer on web workers. Workers allow us to lay out diagrams without blocking the main thread. 

## TODOs

* [x] port `stepNextStage`
* [ ] implement interrupt protocol

* Fix #1194: use `tsc` to emit `components` as an ES module and add `jsxRuntime: "classic"` in client packages for JSX support.
* In addition to a raw Web Worker file (`worker.ts` in `components/src/worker/`), add a wrapper `OptimizerWorker`, which provides an `async`/`await` API using [`MessageChannel`](https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API/Using_channel_messaging). 
  * For code reuse, added `*Grad` versions of `stepState` and `stepUntilConvergence`. We cannot pass `PenroseState` to workers because they are not [clonable](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm). Therefore, for the worker to use  
* Change `Simple` to use `OptimizerWorker`.
* Update `synthesizer-ui` and `editor` to use `OptimizerWorker`
  * The former keeps multiple worker instances (one per `GridBox`)
  * The latter keeps a global instance initialized in `atom.ts`

# Checklist

# Open questions


joshsunshine:
@wodeni What happens to any existing Web Worker(s) when you hit the resample button, switch to another example, etc.? Do we kill the existing Web Worker(s) before spawning a new one?

wodeni:
Closing in favor of #1528 since @YonahGoldberg has made much improvements to this implementation.

PR #1208, opened 2023-01-10
fix: delete `svg-flatten`

samestep:
# Description

This PR effectively reverts #1026 to remove our dependency on [svg-flatten](https://npmjs.com/svg-flatten), which breaks a lot of things and is a blocker for #1173 and #1207.

Take a look at the walk-on-spheres examples; I changed them back to the original `Text` labels @keenancrane was using in #1019, with a few minor improvements (Unicode subscripts)

# Checklist

# Open questions

- We can add back `Equation` stroke in the future if we want; let's just do it in a way that doesn't pull in an ill-behaved dependency.


joshsunshine:
Registry diff looks good to me.

keenancrane:
This reversion made it possible to move forward with some system development, but was a step backward in terms of diagram quality (makes it hard to create legible math labels in many scenarios).

It also seems that it didn't actually help push #1207 forward, which is still not merged.  So, seems it's just a bundling issue, a la #1173.

Is there an alternative to `svg-flatten` that we can use to reinstate the functionality from #1026?  (Especially given the hard work from @cmumatt!)  Or some alternative way of including it so that it doesn't cause problems for bundling?  I don't really understand the issue here.

Issue #1123, opened 2022-10-10
Uncaught error in the Style compiler when selector relations refer to nonexistent variables


wodeni:
**Describe the bug**

When a Style relation (e.g. a binding) refers to variable not defined in the header but defined in the Substance file, an error occurs in the Style compiler. 

```
Uncaught (in promise) TypeError: e.name is undefined
```

**To Reproduce**

https://penrose.cs.cmu.edu/try/?gist=a7a33751a606a5688cb7c7cdf7aea9e7

1. go to the link above
2. press compile
3. observe error in the console

**Expected behavior**

A `varNotFound` error in the selector checker and surfaced to the UI.

This behavior is really caused by the Style compiler, specially the selector checking code, incorrectly merging the `varEnv` coming from the Substance checker and the `selEnv` in the Style compiler. Specially, while merging the two is necessary to match on both Style and Substance variables, the compiler needs to distinguish variables coming from (1) the Substance program itself (for checking against Substance names) and (2) the head of the current selector. 

A TODO item in the source code also described this: https://github.com/penrose/penrose/blob/ce4cb51cdb22b67d01766bd744073f191cc0a262/packages/core/src/compiler/Style.ts#L569

liangyiliang:
That's something that I think I can work on

PR #1239, opened 2023-01-18
fix: style relation checker using substance variables

liangyiliang:
# Description

Related issue/PR: #1123

This PR fixes the uncaught error in the Style compiler when selector relations refer to a variable not declared in the selector, but exists in the substance program. It now throws an error. We added a test case on this error. Finally, we improved upon how the "suggested" variables in the error message distinguish between Style and Substance variables.

The style compiler checks through the relations in the `where` clause, primarily using functions within the Substance checker, which checks the consistency of expressions with respect to a `SubstanceEnv`. In particular, it checks that each variable referred in the expressions are declared inside `SubstanceEnv`.

In order to call the Substance checker, we would need to provide a `SubstanceEnv` that contains all the variables that we want the relations to be able to access. Previously, we just take the current `SubstanceEnv` (as used in the Substance program), and add all the variables declared in the style header. The problem is that now this `SubstanceEnv` contains both variables declared in the style header (which we want to be able to access) and variables not declared in the style header but exists in the substance program.

The solution is simply to create a new, empty `SubstanceEnv` and add in the variables declared in the style header.

[This](https://merging-envs.penrose-72l.pages.dev/try/?gist=a7a33751a606a5688cb7c7cdf7aea9e7) is the same example as used in the original issue, running in the version in this PR. Previously, this did not trigger an error, and only fails silently under the browser console. Now it does trigger a sensible error.

# Checklist

### DIFF
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -571,7 +571,10 @@ const mergeMapping = (
   return {
     ...varEnv,
     vars: varEnv.vars.set(bindingForm.contents.value, toSubstanceType(styType)),
-    varIDs: [dummyIdentifier(bindingForm.contents.value, "Style"), ...varEnv.varIDs],
+    varIDs: [
+      dummyIdentifier(bindingForm.contents.value, "Style"),
+      ...varEnv.varIDs,
+    ],
   };
 };
 
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -568,22 +568,11 @@ const mergeMapping = (
   }
   const [, bindingForm] = res;
 
-  switch (bindingForm.tag) {
-    case "SubVar": {
-      // G || (x : |T) |-> G
-      return varEnv;
-    }
-    case "StyVar": {
-      // G || (y : |T) |-> G[y : T] (shadowing any existing Sub vars)
-      return {
-        ...varEnv,
-        vars: varEnv.vars.set(
-          bindingForm.contents.value,
-          toSubstanceType(styType)
-        ),
-      };
-    }
-  }
+  return {
+    ...varEnv,
+    vars: varEnv.vars.set(bindingForm.contents.value, toSubstanceType(styType)),
+    varIDs: [dummyIdentifier(bindingForm.contents.value, "Style"), ...varEnv.varIDs],
+  };
 };
 
 // TODO: don't merge the varmaps! just put g as the varMap (otherwise there will be extraneous bindings for the relational statements)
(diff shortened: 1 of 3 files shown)


samestep:
Just to clarify, does #1123 become fixed via this PR? If so, could you modify your PR description autoclose it using GitHub's special syntax? https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

liangyiliang:
> Just to clarify, does #1123 become fixed via this PR? If so, could you modify your PR description autoclose it using GitHub's special syntax? https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

Done!

samestep:
> Done!

Thanks! And in future you can do this by just saying "Closes #1123" instead of "Related issue/PR: #1123". Actually, @wodeni do you think we should change this in `.github/PULL_REQUEST_TEMPLATE.md`? I've seen a bunch of people get tripped up by this in the past.

wodeni (approved):

packages/core/src/compiler/Style.ts
```diff
@@ -567,22 +567,31 @@ const mergeMapping = (
...
+          ...varEnv.varIDs,
+        ],
+      };
+    case "SubVar":
+      return {
+        ...varEnv,
+        vars: varEnv.vars.set(
+          bindingForm.contents.value,
+          toSubstanceType(styType)
+        ),
```
samestep, line 589: This code looks duplicated from the above code for `StyVar`; does it make sense to pull it to above the `switch`?

samestep (approved):
Looks great, thanks Leo! Just one minor comment.

packages/core/src/compiler/Style.ts
liangyiliang: Done!

packages/core/src/compiler/Style.ts
```diff
@@ -567,22 +567,29 @@ const mergeMapping = (
...
-    case "StyVar": {
-      // G || (y : |T) |-> G[y : T] (shadowing any existing Sub vars)
+    case "StyVar":
       return {
         ...varEnv,
-        vars: varEnv.vars.set(
-          bindingForm.contents.value,
-          toSubstanceType(styType)
-        ),
+        vars: vars,
```
samestep, line 578:
```suggestion
        vars,
```

packages/core/src/compiler/Style.ts
```diff
@@ -567,22 +567,29 @@ const mergeMapping = (
...
+        varIDs: [
+          dummyIdentifier(bindingForm.contents.value, "Style"),
+          ...varEnv.varIDs,
+        ],
+      };
+    case "SubVar":
+      return {
+        ...varEnv,
+        vars: vars,
```
samestep, line 587:
```suggestion
        vars,
```

PR #1241, opened 2023-01-18
build: fix `npm pack`

samestep:
# Description

Version 2.0.0 of [`@penrose/core`](https://www.npmjs.com/package/@penrose/core) and [`@penrose/components`](https://www.npmjs.com/package/@penrose/components) are missing a lot of files, specifically everything under `dist/` that isn't a direct child of `dist/`, because I tested them using `yarn pack` instead of `npm pack`, and the two commands handle globs differently. This PR fixes the issue.

# Checklist


wodeni (approved):
Surprising that they have different syntax :O. Thanks for fixing this!

PR #1243, opened 2023-01-18
chore: add `new-version` script for version bumps

wodeni:
# Description

`package.json` contains some scripts that we no longer use. Since #1221 stopped canary releases and we are switching to a more active release schedule, this PR adds a `yarn new-version` script for safely bumping versions of all packages in `@penrose` and generating `CHANGELOG` automatically. Note that this script does not create any GitHub releases nor git tags. It only changes `package.json` and `CHANGELOG.md` files.


wodeni:
[The Nx semver plugin](https://github.com/jscutlery/semver) also has opt-out tagging and pushing like lerna, so I don't really think there's a huge difference if we can find a way to use Nx for that. Adding the lerna script because I've only done this with lerna.

samestep (approved):
Awesome thanks!

PR #1250, opened 2023-01-19
docs: update `core` usage in README

wodeni:
# Description

Now that `core` is modules-only and bundle-less, the user can't just import `core` through CDNs such as unpkg because it's difficult to resolve dependencies. The unpkg `?module` suffix doesn't seem to work in our case. Therefore, this PR updates the README to include an example that uses JSPM, a service that resolves ES modules properly and generates import maps.

# Checklist

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -23,12 +23,32 @@ In a HTML document, use [JavaScript modules](https://developer.mozilla.org/en-US
   <body>
     <!--
     JSPM Generator Import Map
-    Edit URL: https://generator.jspm.io/#U2VhYGBkDM0rySzJSU1hcChIzSvKL07VT84vSnUw1DPWMwAA/5OhSyIA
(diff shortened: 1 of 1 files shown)


packages/core/README.md
```diff
@@ -15,18 +15,40 @@ import "global-jsdom/register"; // must come before the Penrose import
...
+<html>
+  <body>
+    <!--
+    JSPM Generator Import Map
+    Edit URL: https://generator.jspm.io/#U2VhYGBkDM0rySzJSU1hcChIzSvKL07VT84vSnUw1DPWMwAA/5OhSyIA
+  -->
+    <script type="importmap">
+      {
+        "imports": {
+          "@penrose/core": "https://ga.jspm.io/npm:@penrose/core@1.3.1-alpha.363/build/dist/index.js"
```
samestep, line 31: This seems wrong, right? Shouldn't this say `2.1.0`?

samestep (changes requested):

samestep (approved):
Looks good to me!

PR #1257, opened 2023-01-23
feat: inline comparison operators

### LINKED ISSUE #1168: Inline comparison operators for constraints and objectives

Currently, only calls to the constraint library is allowed in `ensure` and `encourage` statements. The syntax can be a bit restrictive and counter-intuitive. For instance, to express `norm(c.vec) > 200`, one can only write `ensure lessThan(200, norm(c.vec))`. 

Support boolean expressions such as `==`, `>`, `<`, `<=`, `>=` in constraint/objective statements as a starter. To improve the expressiveness, we may also support chained expressions such as `10 <= norm(c.vec) < 200`.

liangyiliang:
# Description

Resolves #1168 .

We added the support for inline comparison operators. In particular,

* `a > b` is the syntax sugar for `greaterThan(a, b)`
* `a == b` is the syntax sugar for `equal(a, b)`
* `a < b` is the syntax sugar for `lessThan(a, b)`

This applies to both constraints and objectives.

We modified the parser so that constraints and objective declarations now look like
```
<constr> ::= ensure <body> <staged layout>+
   <obj> ::= encourage <body> <staged layout>+

  <body> ::= <identifier> ( <expr_list> )  // Function Call
           | <expr> <op> <expr>            // Inline Comparison

    <op> ::= ==     // Equal
           | <      // Less Than
           | >      // Greater Than
```

In other words, the <body> of a constraint or objective now can either be a `Function Call` or an `Inline Comparison`.

In the Style compiler where we handle constraints and objectives, we proceed as before when we encounter a `Function Call`. When we encounter an `Inline Comparison`, on the other hand, we handle it as if it is a `Function Call`, with the corresponding function name (`lessThan`, `greaterThan`, or `equal`) and the two operands as parameters to the function.

Tests have been added to the parser and compiler.

The docs site has been updated to include this syntactic sugar. Furthermore, we modified the `tree-venn` and `tree-tree` example to utilize this syntactic sugar.

See examples `tree-venn` and `tree-tree` with applications of this syntactic sugar.

# Checklist


packages/core/src/parser/StyleParser.test.ts
```diff
@@ -420,8 +420,14 @@ const {
   -- encourage
   A.fn = encourage obj("string1", true, "string\\n", false)
   A.fn2 = encourage obj( a, b )
+  encourage a == b
```
wodeni, line 423: Might wanna add some complex expressions for sanity check. Something like:

```suggestion
  encourage a == b
  encourage a + a / b > b*b
  encourage MathPI()*4 > abs(sqrt(b))
```

wodeni (approved):
Looks good! This will actually improve the editing experience for Style dramatically for me (and I'm sure many others!). Just had a minor comment about parser tests below. Also I'll try and push some refactoring on the geometry domain

wodeni:
Refactored all `lessThan` and `equal` in `euclidean.sty` to use the new inline operators! Local tests seemed to work very well. Some variations might need tweaks for gallery, which I'll do after CI reports the diffs.

wodeni:
BTW it'd be good to open issues about (1) chaining and (2) other logical operators (eg `||`, `>=` etc). This PR technically doesn't resolve all the things in #1168, but I somewhat prefer closing issues sooner and open more specific followups for the health of the repo.

samestep:
Looks super cool, thanks! Looking at the code now... quick question though, why does the `tree-venn` example change?

packages/core/src/__tests__/overall.test.ts
```diff
@@ -161,9 +161,13 @@ describe("Energy API", () => {
     if (res.isOk()) {
       // NOTE: delibrately not cache the overall objective and re-generate for original and filtered states
       const state = res.value;
-      const smallerThanFns = state.constrFns.filter(
-        (c) => c.ast.expr.name.value === "smallerThan"
-      );
+      const smallerThanFns = state.constrFns.filter((c) => {
+        if (c.ast.expr.body.tag === "FunctionCall") {
+          return c.ast.expr.body.name.value === "smallerThan";
```
samestep, line 166: The PR description says `lessThan`; is there a difference between that and `smallerThan`?

packages/core/src/compiler/Style.ts
```diff
@@ -2128,10 +2130,21 @@ const findPathsExpr = <T>(expr: Expr<T>): Path<T>[] => {
...
+      const body = expr.body;
+      switch (body.tag) {
+        case "FunctionCall": {
+          return body.args.flatMap(findPathsExpr);
+        }
+        case "InlineComparison": {
+          return [body.arg1, body.arg2].flatMap(findPathsExpr);
+        }
+      }
+      break;
```
samestep, line 2147: I see that our particular cocktail of TypeScript and ESLint causes you to have to insert a dead `break;` here due to nested `switch` statements; you could instead consider pulling this case out into its own function, to avoid the dead code.

packages/core/src/compiler/Style.ts
```diff
@@ -2805,6 +2818,41 @@ const stageExpr = (
...
+    const mapInlineOpToFunctionName = (op: "<" | "==" | ">"): string => {
+      switch (op) {
+        case "<":
+          return "lessThan";
+        case "==":
+          return "equal";
+        case ">":
+          return "greaterThan";
+      }
+    };
```
samestep, line 2834: This function could be top-level; doesn't matter much though.

samestep (approved):

samestep:
> We changed `smallerThan(x.icon, y.icon)` with `x.icon.r < y.icon.r`, equivalent to `lessThan(x.icon.r, y.icon.r)`.
> 
> `smallerThan` has a default nonzero padding of 0.4. `lessThan`'s default padding is 0.

Gotcha; but why does that change the colors?

samestep:
I guess the reason I'm asking is that I feel like the colors in the variation as of this PR don't look great; if that example does have to change, could you modify the variation to get nicer colors?

packages/core/src/__tests__/overall.test.ts
liangyiliang: Hmmm oops I need to look more into this ...

I think this test case is specifically looking for `smallerThan` not `lessThan`. `smallerThan` takes two shapes, whereas `lessThan` takes two numbers.

liangyiliang:
> Gotcha; but why does that change the colors?

I think the overall energy function changed, which means a different optimization ...

But in reality now that I think about it, I shouldn't really touch `smallerThan` - since I think the idea of having a padding makes sense ... I will revert the change to `tree-venn`

packages/core/src/__tests__/overall.test.ts
liangyiliang: Fixed - now it is only looking for `smallerThan`.

packages/core/src/compiler/Style.ts
liangyiliang: I decided to use `if` statements instead

samestep (approved):
Great, thanks Leo!

joshsunshine (approved):
Looks great!

Issue #1258, opened 2023-01-23
Full moon example is unstable


joshsunshine:
**Describe the bug**
The `full-moon-full-moon` trio in the registry seems to optimize extremely slowly for some variations. All of the variations of this example, including ones that appear to work, yield `Error: <circle> attribute r: A negative value is not valid.` errors in the console. However, for the variations that work variations the negative value is usually extremely small (e.g. -9.240216684717542e-7). For the variations that fail, the values are much more negative (e.g. -562.1579801179566). The IDE is completely  unresponsive after running these variations, even after waiting 5+ minutes.

**To Reproduce**
Load `full-moon-full-moon` in the IDE and hit resample a few times.

PR #1259, opened 2023-01-23
chore: remove full moon trio

joshsunshine:
# Description
Temporary removes the full-moon trio as a workaround for issue #1258.


joshsunshine:
I left the substance, style, and domain, blocks in place in`registry.json` since I anticipate adding this trio back once we fix #1259.

PR #1264, opened 2023-01-25
feat: compile diagrams in `editor` after detected changes in `roger`

### LINKED ISSUE #1048: When a file is updated in the IDE by Roger, automatically trigger a recompile
No.

When a file is updated in the IDE by Roger, automatically trigger a recompile

Currently it is necessary to manually click "Compile" after Roger updates a file.

YonahGoldberg:
# Description

Resolves #1048 

When editing a file watched by Roger, if you save the file it now automatically triggers compilation of the diagram.

This was really a quick change, except it was a bit subtle to switch to using 
`snapshot.getPromise` instead of `snapshot.getLoadable` to obtain the current workspace state. The latter would only compile diagrams with the newly changed code after two saves in the IDE. To be honest, I'm still not completely sure why this is the case. I think it has something to do with `snapshot.getPromise` being preferable in an async callback, but probably someone else on the Penrose team would know better.

# Checklist

### DIFF
--- a/packages/examples/src/animation/center-shrink-circle.sub
+++ b/packages/examples/src/animation/center-shrink-circle.sub
@@ -1,2 +1,2 @@
-Object o 
+Object o
 AutoLabel All
--- a/packages/examples/src/animation/center-shrink-circle.sub
+++ b/packages/examples/src/animation/center-shrink-circle.sub
@@ -1,2 +1,2 @@
-Object o
+Object o 
 AutoLabel All
--- a/packages/editor/src/App.tsx
+++ b/packages/editor/src/App.tsx
@@ -147,11 +147,6 @@ function App() {
   const isPortrait = useMediaQuery({ query: "(orientation: portrait)" });
   const isTabletOrMobile = useMediaQuery({ query: "(max-width: 1224px)" });
   const compileDiagram = useCompileDiagram();
-  // const value = useRecoilValue(currentWorkspaceState);
-
-  // useEffect(() => {
-  //   toast.error(value.files.domain.contents);
-  // }, [value]);
 
   const ws = useRef<WebSocket | null>(null);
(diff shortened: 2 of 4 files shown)


wodeni (approved):
This looks great! 

Thinking about next steps: It would be great if we can do the same in the editor pane. This requires a bit more care as the editor pane doesn't have an explicit "save" button and we may not want to render compile errors all the time.

wodeni:
> To be honest, I'm still not completely sure why this is the case. I think it has something to do with snapshot.getPromise being preferable in an async callback, but probably someone else on the Penrose team would know better.

Can't say that I know better, but [the docs did say the same](https://recoiljs.org/docs/api-reference/core/Snapshot/#reading-snapshots). Notably, [docs on `Loadable`](https://recoiljs.org/docs/api-reference/core/Loadable/) says:

> If the state has a value then it returns a value, if it has an error then it throws that error, and if it is still pending then it suspends execution or rendering to propagate the pending state.

packages/examples/src/animation/center-shrink-circle.sub
```diff
@@ -1,2 +1,2 @@
-Object o 
+Object o
```
samestep, line 1: This file change is causing a merge conflict because of #1280; could you either remove this edit or pull from `main` and then push to your branch again with the new `.substance` file extension? (I would do it myself but your branch is on your fork instead of in this source repo 😛)

samestep (approved):
Thanks for doing this! Unfortunately I don't have the Recoil expertise to comment on the details. Regarding @wodeni's comment:

> Thinking about next steps: It would be great if we can do the same in the editor pane. This requires a bit more care as the editor pane doesn't have an explicit "save" button and we may not want to render compile errors all the time.

I agree this would be nice as a followup after we merge this PR. Seems related to #1272?

PR #1265, opened 2023-01-25
fix: enforcing ordering in `collinearOrdered` constraint

### LINKED ISSUE #1262: Constraint `collinearOrdered` does not enforce ordering
**Describe the bug**

The constraint `collinearOrdered` is supposed to enforce ordering, but it does not.

**To Reproduce**

[Here](https://penrose.cs.cmu.edu/try/?gist=e1790dbaf54393efc4904b3470b5fe0b)

**Expected behavior**

The order at which the letters appear should either be `abc` or `cba` since we used `collinearOrdered`, but in some variations, we see `bca` and `cab`, violating this requirement.

**Screenshots**
image

wodeni:
# Description

Resolves #1262.

Somehow the implementation changed to `minN` of all triangle inequalities instead of just one. This PR reverts that change since `collinear` already has an implementation based on cross product.

The `wet-floor` example has been changed to use `collinearOrdered` since that's what we intended.

# Checklist

### DIFF
--- a/packages/core/src/contrib/Constraints.ts
+++ b/packages/core/src/contrib/Constraints.ts
@@ -28,7 +28,6 @@ import {
   lt,
   max,
   min,
-  minN,
   mul,
   neg,
   squared,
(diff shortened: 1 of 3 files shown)


liangyiliang (approved):
Thank you so much! This works great!

Also, when the `wet-floor` example was written, `collinear` was order-sensitive whereas `collinearUnordered` was order-insensitive, I will also change the occurence of `collinear` in `wet-floor` to `collinearUnordered`

PR #1286, opened 2023-02-02
build: monorepo-aware reloading in `editor`

wodeni:
# Description

Previously, we had trouble watching changes from packages in our monorepo. For instance, a change in `@penrose/core` will not trigger a reload when running `yarn start` because it only watches source file changes within `@penrose/editor`. 

After #1160, all of our packages are ES modules and vite is capable of treating ESM dependencies as source. This PR configures `@penrose/editor` such that it watches changes from both `core` and `components`.

* Top-level await: since we are treating `core` as source, `editor`'s vite config also needs to know about how to deal with top level await. Added the [suggested plugin](https://github.com/penrose/penrose/tree/main/packages/core).
* Absolute path resolution: vite is not going to run custom plugins such as `ttsc` in `core`. Therefore, we need to make sure `core` has minimal configuration. This PR removes `ttsc` from `core` and changed all absolute path imports to relative. 
  * it doesn't look that bad. Most imports are just `../<module>/file`, which is pretty logical.

Run `yarn start`. Touch any files in `core` or `components` to trigger HMR in vite.

# Checklist

# Open questions

[This issue](https://github.com/vitejs/vite/issues/6479) was very helpful for this PR, but I couldn't reproduce the regex-based approach because `examples` and `optimizer` have different folder structures. I'm inclined to keep the aliases explicit because it's much more transparent.


samestep:
> * Absolute path resolution: vite is not going to run custom plugins such as `ttsc` in `core`. Therefore, we need to make sure `core` has minimal configuration. This PR removes `ttsc` from `core` and changed all absolute path imports to relative. 
>   * it doesn't look that bad. Most imports are just `../<module>/file`, which is pretty logical.

Related: #1189

samestep (approved):
Amazing, thanks for doing this Nimo! One thing to think about: now the only source package this doesn't work for is `@penrose/optimizer` which has a more complicated build process that Vite doesn't understand, which is why we still need `"dependsOn": ["^build"]` for the three dev tasks `packages/editor/package.json`. But those tasks don't actually depend on `core:build` or `components:build`, just `optimizer:build`. Doesn't really matter though :)

PR #1287, opened 2023-02-02
feat: show multiple diagram instances on a grid in `editor`

wodeni:
# Description

This PR introduces a new `Grid` panel to `editor`, which show multiple diagram instances at once. The number of instances is 10 by default, and configuration through the `option` panel. 

`synthesizer-ui` first introduced a grid of Penrose-generated diagrams in a web app. This PR pulls relevant components (i.e. `Gridbox` and `Grid`) into `components` and uses them in both `editor` and `synthesizer-ui`.

# Implementation strategy and design decisions

* `Grid` contains `Gridbox`, which contains `Simple`. The properties are passed through, with `Gridbox` inheriting all `Simple`'s props and `Grid` having a `gridboxProps` option for each `Gridbox`. 
* `Gridbox` can be `stateful`, which means that it caches the trio it receives from props _and_ caches `state` from `Simple`'s `onFrame` callback.
* Styling: use `styled-components` to style `Grid` and `Gridbox`. The client may provide a theme using `<ThemeProvider>`. 
* `editor`'s state management: the IDE wraps `GridPanel` around `Grid` and maintains a `diagramGridState`. All state updates are done through Recoil. 
* `Image` and `Line` arrowhead id collisions: since SVG has a global namespace, when there are multiple diagrams on the page with duplicated defs, they disappear. This PR introduces an util function `makeIDsUnique`, which changes all definitions and corresponding references of [IRI references](https://dev.w3.org/SVG/profiles/1.2T/publish/diff/linking.html) in the SVG such that they are unique. (Resolves #1288)

# Examples with steps to reproduce them

See preview deployment:
grid

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new ESLint warnings
- [x] I have reviewed any generated registry diagram changes

# Open questions

* Caching in `editor`: the grid doesn't cache anything other than a list of `variation`s, which causes re-rendering whenever the grid size is decreased and then increased. No re-rendering happen when the `gridSize` is decreasing because there are no prop changes.

Issue #1282, opened 2023-01-31
Consolidate shape types


samestep:
https://github.com/penrose/penrose/blob/6e835968280a784a91c4a2ca47a226516a3067d0/packages/core/src/shapes/Shapes.ts#L24-L34

https://github.com/penrose/penrose/blob/6e835968280a784a91c4a2ca47a226516a3067d0/packages/core/src/types/shape.ts#L4-L15

https://github.com/penrose/penrose/blob/6e835968280a784a91c4a2ca47a226516a3067d0/packages/core/src/types/shapes.ts#L65-L67

https://github.com/penrose/penrose/blob/6e835968280a784a91c4a2ca47a226516a3067d0/packages/core/src/types/value.ts#L6-L13

Issue #715, opened 2021-12-20
Vector functions are inconvenient for 2D


samestep:
#698 adds an `isPt2` [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) to `types/ad.ts` which checks whether the `length` of a `VarAD[]` is `2`. (It serves a purpose similar to that of the `toPt` function in `contrib/Functions.ts`.) Some usages of this are due to `properties` using `VectorV` because there isn't a `Value` variant for `Pt2` (see also #701), but several of them are for `VarAD[]` outputs that came from calling `ops` functions from `engine/Autodiff.ts` on `Pt2` inputs.

It would be nice if the return _types_ of those functions somehow preserved the dimension of their input types.

---

As of c0a01c87fb631fc6c4b9af42d4228a8194a61d34, here is a list (comprehensive within each file but perhaps not within the entire repo) of exported functions that produce vectors with the same dimension as (part of) their input(s):

- `contrib/`
  - `Functions.ts`
    - `compDict`†
      - `normalize`
      - `ptOnLine`
      - `midpoint`*
      - `midpointOffset`*
      - `innerPointOffset`*
      - `unit`
      - `mul`
      - `vmul`
      - `rot90`*
- `engine/`
  - `Autodiff.ts`
    - `makeADInputVars`
    - `ops`
      - `vadd`
      - `vsub`
      - `vmul`
      - `vneg`
      - `vdiv`
      - `vnormalize`
      - `vmove`
      - `rot90`*
      - `vrot`*
    - `fns`
      - `center`
- `utils/`
  - `OtherUtils.ts`
    - `scalev`
    - `addv`
    - `subv`
    - `negv`
    - `linePts`
    - `getStart`
    - `getEnd`
  - `Util.ts`
    - `loadImages`
    - `ops`
      - `vadd`
      - `vsub`
      - `vmul`
      - `vneg`
      - `vdiv`
      - `vnormalize`
      - `vmove`
      - `rot90`*
    - `intersection`*

Notes:

- functions marked with an asterisk * don't work at all on vectors with a dimension other than 2
- functions in objects marked with a dagger † produce `Value` subtypes, not raw arrays

The above functions make use of three dimension-preserving functions from the JavaScript standard library and lodash; in other words, since we don't have control of the type signatures of these functions, resolving this issue may require us to either write our own versions of them or otherwise modify the implementations of some vector functions to avoid using them:

[...]

PR #1294, opened 2023-02-02
feat: basic group shape

### LINKED ISSUE #783: Add `Group` shape to Style
**What's the problem?**

A longstanding deficiency in Style is that we don't provide a mechanism for defining a "group" shapes.  A group is simply an unordered collection of shapes, including other groups.  In SVG, groups are implemented as nested SVG tags.  In general, groups play several important roles in vector graphics:

- **Scene graph.** In conjunction with spatial transformations, groups enable one to define a [scene graph](https://en.wikipedia.org/wiki/Scene_graph).  A scene graph is basically a hierarchical description of how shapes in a scene are arranged.  For instance, to describe the configuration of a human character in a 3D environment, you might think of the torso as the root node; the neck and upper limbs are children of the torso; the head and lower limbs are children of the next and upper limbs; etc.  This graph can be implemented via grouping: the outermost group contains a shape for the torso, and the groups for the limbs and head/neck.  The group for a limb contains a shape for the upper part of a limb and a group for whatever is in the lower limb (including fingers and toes, etc.).  If we now associate a transformation with each group node, that node gets applied to the shapes in that node, and recursively to the child shapes.  For instance, rotating the group containing the torso rotates everything; rotating the group containing the upper limb also rotates the lower limb.  Etc.

image

- **Clipping.** Even without transformations, groups play an important role in _clipping_.  Here, one shape is used to map out a collection of shapes.  A basic use case might be trimming objects to the bounds of the canvas.  In SVG, clipping is achieved via [the `clipPath` tag](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Clipping_and_masking).  Note that this construct does not itself rely on groups—for instance, you could apply the same clip path repeatedly to a bunch of shapes.  But the most natural thing to do is group everything that will be clipped, then apply a clip path _once_ to that group (see example below).  There is also a similar-but-different notion of [compound paths](https://www.vektorgarten.de/illustrator-compound-path-shape.html).  Such paths are used, for example, to make a hole in the letter "O".

image

- **Editability.** Like named layers (see #768), groups are also essential for making vector graphics files easily editable.  Currently Penrose just spits out a big soup of SVG shapes.  These shapes are difficult to edit in external programs like Illustrator and Inkscape (currently you have to load up the file and group elements by hand!).  Yet individual shapes very often have a natural organization.  In fact, if a `Group` shape were available in Style, it would be easy for Style programmers to define grouping based on Style rules.  Since these rules are triggered by the semantics of the Substance program, the resulting diagrams would be nicely grouped into semantically-meaningful collections of shapes.  (Or any other organization the Style writer wants to provide to a human editor---for instance, all vertices, all edges, and all triangles in a mesh could be put in separate groups).  Here's an example of an SVG that has been nicely structured by grouping, and has annotations via `id`s and `title`s.  Here I've annotated the drawing with information about spatial placement—more natural in Penrose would be to annotate it with the automatically-generated names from substance, or more descriptive names assigned by the Style programmer:

```
<svg>
   <g id="wholeDrawing">
      <rect x="654" y="174" width="208" height="208"><title>A big rectangle at the top.</title></rect>
      <g id="leftHalf">
         <rect x="535" y="427" width="138" height="138"><title>A medium rectangle at left.</title></rect>
         <g id="leftBottom">
            <rect x="459" y="598" width="85" height="85">
            <rect x="569" y="598" width="85" height="85">
         </g>
      </g>
      <g id="rightHalf">
         <rect x="838" y="427" width="138" height="138"><title>A medium rectangle at right.</title></rect>
         <g id="rightBottom">
            <rect x="883" y="598" width="85" height="85"/>
            <rect x="994" y="598" width="85" height="85"/>
         </g>
      </g>
   </g>
</svg>
```

---

**What's the proposal?**

The basic proposal is to define a new shape called `Group`.  The basic fields of a `Group` are:

1. `shapes:` — a list of shapes contained in the group.
2. `transformation:` — a spatial transformation applied to the group and everything in it.  (For now this field can be unused—we shouldn't block on implementing shapes just because we haven't completely fleshed out the design for transformations.  But a placeholder would be nice to have.)
3. `name:` — just like any other shape, a group has a name.

The list syntax for `shapes` can be whatever makes sense/is easiest to implement in Style; we just need to be able to specify a list of shapes of any length.  Just to provide a definite proposal, we could write something like

`shapes: [ x.myRect, x.myCircle, x.myPath ]`

_Optional:_ Beyond these basic fields, we could also allow arbitrary SVG attributes.  These attributes would be applied to any child that does not already define this attribute.  For instance, if we write

```
shape x.g = Group {
   shapes: [ x.myRect, x.myCircle, x.myPath ]
   stroke-width: 3.0
}
```

then the shapes `x.myRect`, `x.myCircle`, and `x.myPath` would inherit a `stroke-width` of 3.0—unless they define their own stroke width.  Basically attributes defined lower in the tree take precedence.  (And as always, fields overriden in lower blocks take precedence over fields set in earlier Style rules—group precedence and rule precedence are entirely orthogonal.) **Note:** This functionality is easily supported via SVG—for example, the following code will color the outer circle green and the two inner circles blue:

```
<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
	 viewBox="0 0 1920 1080" style="enable-background:new 0 0 1920 1080;" xml:space="preserve">
<g color="green">
	<circle cx="941.67" cy="363.67" r="156.67" fill="currentcolor"/>
	<g color="blue">
		<circle cx="818.33" cy="552" r="156.67" fill="currentcolor"/>
		<circle cx="1038.33" cy="568.67" r="156.67" fill="currentcolor"/>
	</g>
</g>
</svg>
```

image

---

**Implementation.**

Implementation amounts to doing just a couple things:

* Define a new `Group` shape (as usual), with the properties above.
* Modify the renderer to emit grouped shapes in group order, within nested `<g>` tags.

One place where implementation gets tricky is interaction with layering.  The way this works out in basically every vector graphics setting is that all objects within a group are either above, or below, all objects within another disjoint group.  Otherwise, objects within the same group (including other groups) have some layer ordering.  For instance, if we have a group

`A = [ a1, a3, a2 ]`

and another group

`B = [ b3, b1, b2 ]`

and say that group `A` is above group `B`, then the overall ordering will end up being

`b3 < b1 < b2 < a1 < a3 < a2`.

There is no way to "interleave" ordering across groups, e.g., it's impossible for `a3` to be above `b2` *and* `a1` to be below `b3`.  The reason for this is that objects in an SVG will still get layered from top to bottom.  Since shapes within a group all appear in one contiguous chunk of tags, there's no way to interleave the layer orderings.  (Some thought still required here to totally flesh out layer order, and how to fail gracefully in a world with `Groups`…)

liangyiliang:
# Description

Resolves #783 .

We added a basic `Group` shape that contains multiple already-declared shapes. Each group is now rendered as a `<g>` tag which contains the internal shapes.

As an example,
```
forall T t {
    t.s1 = Circle {}
    t.s2 = Rectangle {}
    t.s3 = Text {
        string: "Hello"
    }
    t.g = Group {
        shapes : [t.s1, t.s2, t.s3]
    }
    t.s4 = Circle {}
    t.g2 = Group {
        shapes : [t.s4, t.g]
    }
}
```
can now generate 
```
<svg ...>
  <g>
    <circle ... />
    <g>
      <circle ... />
      <rect ... />
      <text .../>
    </g>
  </g>
</svg>
```

# New Shape

We add a new shape, `Group`, to the shape library. As of now, `Group` shape only has one property, `shapes`, and an automatically generated `name`. Property `shapes` is a list of paths to previously-defined shapes.

# Changes to the Compiler

The Style language now allows users to have a list of shapes, denoted `[path1, path2, ...]` where `path1`, `path2`, ... are paths to shapes.

## GPI or ShapeAD Objects?

We go through and process each Style statement in the `translation` process, at which point only the GPIs of each shape can be accessed. The actual `ShapeAD` objects are available near the end of the compilation process. Therefore, a Group shape, initially, stores a list of GPIs. Near the end of the compilation process, after we already have a list of all ShapeADs, we switch each GPI into the corresponding ShapeAD objects.

## Group Graph

The compiler uses a directed graph, named `GroupGraph`, to represent group-shape relationships between shapes. Formally, a directed edge exists between two shapes ($g \rightarrow s$) if and only if shape $s$ is a child of group $g$.

For example,
- Suppose we have shapes `s1, s2, s3, s4, s5, s6`
- Shapes `s1, s2` belong to group `g1`
- Shapes `s3, s4` belong to group `g2`
- Groups `g1` and `g2`, and shape `s5`, all belong to group `g3`
- Shape `s6` does not belong to any groups.

Then, the group graph should look like
image

We can use this group graph to check whether or not the grouping relations are well-defined, which means

- the grouping relations should not be cyclic
- each shape can have at most one parent.

The compiler fires a warning if any of these checks fail. A well-defined grouping relation should look like a tree (as in the example).

## Layering

Suppose all layering directives have the form `layer s above s'` (`layer s below s'` is equivalent to `layer s' above s`). We want the following nice properties: (these are not the actual rules we follow for the algorithm)

- Layering on a group implies the same layering on all members of the group.
- Layering on a member of a group implies the same layering on the entire group.

Concretely, in the group graph above, writing `layer s1 above g2` implies:

- `layer s1 above g3` and `layer s3 above s4` (since `s3, s4` are members of `g2`)
    - `layer g1 above s3` and `layer g1 above s4` (since `s1` is a member of `g1`)
      - `layer s2 above s3` and `layer s2 above s4` (since `s2` is a member of `g1`)
- `layer g1 above g2` (since `s1` is a member of `g1`)

These are the "saturation" rules which saturates the entire layer graph and hence is not efficient. The algorithm leverages upon the observation that layering directives are only useful when two shapes are siblings within the same group. Suppose we encounter the layering directive `layer s above s'`.

When encountering `layer s1 above s2`, the algorithm attempts to "walk up" the group graph from both `s1` and `s2` towards the root, until when the algorithm finds `s1'` and `s2'` which are siblings. Then, the algorithm applies the same layering directive to `s1'` and `s2'`.

In the concrete example of `layer s1 above g2`, the algorithm would "walk up" `s1` to get `g1`. Since `g1` and `g2` are siblings, we add `layer g1 above g2` to the layering graph. As another example, `layer s1 above s6` involves walking up `s1` twice to get `g3` which is a sibling of `s6`, so we add `layer g3 above s6` to the layer graph.

A few more (more trivial) examples:

- `layer s3 above s4` gets added to the layer graph directly, since `s3` and `s4` are already siblings.
- `layer g3 above g3` also gets added to the layer graph directly, since we consider `g3` to be a "sibling" of itself. This will cause a cycle, which is perfectly expected.
- `layer s4 above g3` would involve walking up `s4` twice to get `g3`. Similarly, we add `layer g3 above g3`.

The final layer graph would only contain edges between siblings.

## Keeping Groups Together

```
s1 = // some shape
s2 = // some shape
g = Group {
  shapes: [s1, s2]
}
s3 = // some shape

layer s1 above s2
layer s3 above s2
layer s3 below s1
```
The layering directives would force the layering to be `bottom s2 s3 s1 top`. This itself does not cause a loop, but does break up the group `g`. The compiler should generate a warning.

Running the algorithm, we have

- `layer s1 above s2` just gets added to the layer graph because `s1, s2` are siblings.
- `layer s3 above s2` requires one "walk up" from `s2` to `g`, which is a sibling of `s3`, so `layer s3 above g` gets added to the layer graph.
- `layer s3 below s1` requires one "walk up" from `s1` to `g`, which is a sibling of `s3`, so `layer s3 below g` gets added to the layer graph.

We hence have both `layer s3 above s2` and `layer s3 below s1`, which causes a loop and a warning should fire.

## Changes to Compilation

The style compilation results in a `State` object whose `shapes` field contains a flat, ordered list of all shapes. With groups, the list of shapes should no longer be a flat list. It still stores `ShapeAD` objects, but each `Group` shape's members should be actual `ShapeAD` objects. That is, the list of shapes should contain only `ShapeAD` objects for the top-level shapes, and all lower-level shapes should be within their respective parent groups. Furthermore, the list of top-level shapes, and for each group, the list of its child shapes, should all be ordered according to the layering. This is easy simply by sorting the elements of the lists based on the layer ordering. The list of top-level shapes should then be passed into the renderer.

Previously, after `translation`, the compiler does the following in order:

- Compute the layer ordering based on the layering graph (`computeShapeOrdering`)
- Using the layer ordering, compute an ordered list of `ShapeAD` objects (`getShapes`)

The change goes,

- Generate a flat list of all shapes (where `Group` shapes store GPIs, not actual shape objects, since shape objects don't exist yet).
- Generate a group graph based on the list of shapes. Each node is a string. This group graph does not respect the layering.
- Generate layer ordering using the layering directives and group graph.
- Annotate the group graph so that each node are labeled with the layer ordering index. This group graph now respects layering within each group.
- Use the annotated group graph to generate a list of top-level `ShapeADs`, this time each group storing its actual sub-shapes. This is what we pass into the renderer.

# Changes to the Renderer

The renderer now takes a list of top-level shapes, and renders them. Whenever it encounters a group, it recursively renders all the member shapes before packaging everything into one `<g> ... </g>` tag.

The `Wet Floor` example now uses groups.

Can also do bbox-related constraints on groups: [here](https://group-shape.penrose-72l.pages.dev/try/?gist=ef5023dd2a243a65dfec4658277ca782)

# Checklist

# Open questions

Turns out, warnings don't show up in the IDE.


packages/core/src/renderer/Renderer.ts
```diff
@@ -155,23 +201,83 @@ export const RenderInteractive = async (
...
-          canvasSize: state.canvas.size,
-          pathResolver,
-        },
-        onDrag,
-        svg
-      )
-    );
-  }
+  const shapes = state.computeShapes(state.varyingValues);
+  const graph = makeGroupGraph(shapes);
```
wodeni, line 205: The `Render*` functions are called quite frequently in clients, so I don't think it's great to re-compute the group graph every single time, especially when they largely stay the same. One alternative is to cache a static group graph with shape names as nodes in `State` and simply index into `shapes` at render time.

packages/core/src/utils/GroupGraph.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+  return graph;
+};
+
+export const findRoot = <T extends Shape | ShapeAD>(
+  graph: GroupGraph<T>
+): string => {
+  const roots = graph.sources();
+  if (roots.length === 0) {
```
wodeni, line 82: Why is this case needed? Also it's possible to dedup this function by `const nodes = roots.length > 0 ? roots : graph.nodes()` and reducing `nodes` instead.

packages/core/src/utils/GroupGraph.ts
```diff
@@ -0,0 +1,103 @@
...
+  GroupGraphNode<T>
+>;
+
+const makeGroupGraphNode = <T extends Shape | ShapeAD>(
+  shape: T,
+  index: number
+): [string, GroupGraphNode<T>, string[]] => {
+  const shapeType = shape.shapeType;
+  const shapeNameVal = shape.properties["name"];
+  if (shapeNameVal.tag !== "StrV") {
```
wodeni, line 20: Unrelated to this PR: this type of boilerplate should go away after we improve the shape types.

packages/core/src/utils/GroupGraph.ts
```diff
@@ -0,0 +1,103 @@
...
+    const [name, node, children] = makeGroupGraphNode(shapes[i], i);
+    rawGraph[name] = [node, children];
+  }
+
+  for (const [name, [node, children]] of Object.entries(rawGraph)) {
+    graph.setNode(name, node);
+    for (const child of children) {
+      graph.setNode(child, rawGraph[child][0]);
+      graph.setEdge({ i: name, j: child, e: undefined });
+    }
```
wodeni, line 73: Why do you need `rawGraph`? Can you insert nodes to the graph immediately after calling `makeGroupGraphNode`?

packages/core/src/compiler/Style.test.ts
```diff
@@ -1044,4 +1131,61 @@ delete x.z.p }`,
...
+        groupGraph.setNode(x, {
+          shape: { shapeType: "Group", properties: {} },
+          index: 0,
+        });
+      });
+      groupGraph.setEdge({ i: "A", j: "X", e: undefined });
+      groupGraph.setEdge({ i: "B", j: "X", e: undefined });
+      const warnings = S.checkGroupGraph(groupGraph);
+      expect(warnings.length).toEqual(1);
+      expect(warnings[0].tag).toEqual("ShapeBelongsToMultipleGroups");
```
wodeni, line 1188: Great that you added these tests!

wodeni (changes requested):
Thanks for doing this @liangyiliang! Very important first step towards better structure and reusability in Style. 

My main request is to take a closer look at group graph construction in the renderer and figure out if we can avoid re-computation of the graph as much as possible. 

I haven't thought deeply about the usability and performance implications of the saturation rules. At a glance it makes sense to me and the implementation seems straightforward. I'll leave it to @joshsunshine and @samestep to evaluate.

Generally, this PR has some duplicated code and some functions tend to have very long names. I'd suggest refactoring them to avoid duplication. If a function doesn't have a great name and it's called once in your code, I think you can consider inlining it to improve readability. See my code-level comments below.

liangyiliang:
> My main request is to take a closer look at group graph construction in the renderer and figure out if we can avoid re-computation of the graph as much as possible.

I agree! I will store the Group Graph in the State.

packages/core/src/utils/GroupGraph.ts
liangyiliang: If there are no roots, we want to still get a reasonable node to start rendering - that is, the overall node with the smallest layer.

packages/core/src/types/state.ts
```diff
@@ -32,7 +32,7 @@ export interface State extends OptState {
   constrFns: Fn[];
   inputs: InputMeta[]; // same length as `varyingValues`
   labelCache: LabelCache;
-  shapes: ShapeAD[];
+  shapes: ShapeAD[]; // Consistent with layer-ordering, not necessarily with group-ordering
   canvas: Canvas;
```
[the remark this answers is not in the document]
liangyiliang: Agreed

liangyiliang:
@wodeni One issue is that there are technically two different types of shapes. One is `Shape` (that stores static numbers), and one is `ShapeAD` (that stores dynamic, differentiable numbers). The style compiler only has access to `ShapeAD`, not `Shape`, but the renderer needs `Shape`.

wodeni:
> @wodeni One issue is that there are technically two different types of shapes. One is `Shape` (that stores static numbers), and one is `ShapeAD` (that stores dynamic, differentiable numbers). The style compiler only has access to `ShapeAD`, not `Shape`, but the renderer needs `Shape`.
> 
> The group graph would either store `ShapeAD` or `Shape` ... But one thing I can do is just store the names of the shape, and have a separate `nameShapeMap`.

Tagging https://github.com/penrose/penrose/issues/1282 as a reminder for ourselves. BTW is this a response to the comments about boilerplates?

liangyiliang:
> > @wodeni One issue is that there are technically two different types of shapes. One is `Shape` (that stores static numbers), and one is `ShapeAD` (that stores dynamic, differentiable numbers). The style compiler only has access to `ShapeAD`, not `Shape`, but the renderer needs `Shape`.
> > The group graph would either store `ShapeAD` or `Shape` ... But one thing I can do is just store the names of the shape, and have a separate `nameShapeMap`.
> 
> Tagging #1282 as a reminder for ourselves. BTW is this a response to the comments about boilerplates?

Nope! This is about storing the GroupGraph in the state. Right now, GroupGraph stores actual Shape instances. I will make it so that GroupGraph stores shape names, and when rendering, using a map `ShapeName --> Shape` to get the actual shape objects. Hopefully, the map `ShapeName --> Shape` should be much much easier to compute than the group shape, which would only need to be computed once.

packages/core/src/renderer/Renderer.ts
liangyiliang: Done!

packages/core/src/utils/GroupGraph.ts
liangyiliang: I originally needed it, but not anymore. Done.

packages/core/src/types/state.ts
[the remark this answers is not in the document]
liangyiliang: Done

packages/core/src/renderer/Renderer.ts
```diff
@@ -155,23 +201,83 @@ export const RenderInteractive = async (
...
+    const outSvg = await shapeMap[node.shape.shapeType]({
+      ...shapeProps,
+      shape: node.shape,
+    });
+    graph.delNode(name);
+    return outSvg;
+  }
+};
+
+const RenderGroupGraph = async (
```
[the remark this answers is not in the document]
liangyiliang: Done - I'm actually passing in an optional `interactiveProp` with everything needed for interactivity. If `interactiveProp` is undefined, we render statically. Otherwise we render interactively.

packages/core/src/renderer/Renderer.ts
```diff
@@ -155,23 +201,83 @@ export const RenderInteractive = async (
...
+  for (const childName of orderedChildren) {
+    const childSvg = await RenderGroupGraphNode(childName, graph, shapeProps);
+    elem.appendChild(childSvg);
+  }
+  const node = graph.node(name);
+  attrAutoFillSvg(node.shape, elem, [...attrTitle(node.shape, elem), "shapes"]);
+  return elem;
+};
+
+const RenderGroupGraphNode = async (
```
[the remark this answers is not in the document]
wodeni: ok seems like you need the helper for mutual recursion. in that case this is okay

wodeni (approved):
Thanks for refactoring! Code looks much better. Again, I'd wait for others to give feedback on the group layering semantics.

samestep:
I still need to review this code, but I was wondering, @wodeni why does the registry diff say that so many of the diagrams changed? It looks like this is because of the `uuid` stuff added in #1287; is that going to cause this same set of diagrams to appear as "changed" in every PR?

wodeni:
> I still need to review this code, but I was wondering, @wodeni why does the registry diff say that so many of the diagrams changed? It looks like this is because of the `uuid` stuff added in #1287; is that going to cause this same set of diagrams to appear as "changed" in every PR?

Yes. Didn't realize this was going to break the diagram diff check :(. We could use [uuid.v5](https://www.npmjs.com/package/uuid#uuidv5name-namespace-buffer-offset) to make it deterministic?

samestep:
> Yes. Didn't realize this was going to break the diagram diff check :(. We could use [uuid.v5](https://www.npmjs.com/package/uuid#uuidv5name-namespace-buffer-offset) to make it deterministic?

OK! Was just curious; but yeah, if you have a way to make it deterministic then that would be awesome! Could you open a PR?

samestep:
Still haven't read the code (hopefully can do that today!) but looking at the conceptual overview of the "saturation" approach to handle layering, does this really make the most sense as an approach? Rather than filling out a really dense layering graph anytime groups are involved, I'd think we'd want to approach this by _first_ forming the tree according to groups, _then_ doing layering by taking each `layer` directives and moving each side up the tree until both sides are siblings of each other. (For error message reporting, we can keep a mapping so we know what layering directive each thing came from initially.) So, then, instead of forming one graph that can saturate to quadratic size, we have a bunch of smaller graphs, one for each `Group` node in the diagram tree. We run topological sort separately on each of these, because that's the only context in which ordering makes sense (sorting children of a particular node in the tree).

packages/core/src/utils/Graph.ts
```diff
@@ -153,6 +153,30 @@ export default class Graph<I, L = undefined, E = undefined> {
...
+  /**
+   * throws if the graph does not contain node `i`
+   */
+  children(i: I): I[] {
+    const vert = this.g.get(i);
+    if (vert) {
+      return vert.s.map(({ i }) => i);
+    } else {
+      throw Error("Node does not exist");
+    }
```
samestep, line 165: Use `this.get` instead for better concision and error handling:

```suggestion
    return this.get(i).s.map(({ i }) => i);
```

packages/core/src/utils/Graph.ts
```diff
@@ -153,6 +153,30 @@ export default class Graph<I, L = undefined, E = undefined> {
...
+  /**
+   * throws if the graph does not contain node `i`
+   */
+  parents(i: I): I[] {
+    const vert = this.g.get(i);
+    if (vert) {
+      return vert.p.map(({ i }) => i);
+    } else {
+      throw Error("Node does not exist");
+    }
```
samestep, line 177: Same here:

```suggestion
    return this.get(i).p.map(({ i }) => i);
```

packages/core/src/shapes/Shapes.ts
```diff
@@ -185,6 +188,44 @@ const Text = ShapeDef({
...
 
+// bboxFromGroup is here, not in engine/BBox, to prevent cyclic import
+const bboxFromGroup = ({ shapes }: GroupProps): BBox.BBox => {
+  const content = shapes.contents;
+  const bboxes = content.map((gpi) => {
+    const shapeType = gpi.contents[0];
+    const rawShapeProps = gpi.contents[1];
+    if (!isShapeType(shapeType)) {
+      throw new Error("Unknown shape in Group bbox: " + shapeType);
+    } else {
```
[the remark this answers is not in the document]
samestep: Yep, agreed; #715 touches on this, but doesn't entirely cover the problem.

packages/core/src/index.ts
```diff
@@ -411,7 +410,6 @@ export {
   parseDomain,
   Synthesizer,
   showMutations,
-  RenderShape,
```
samestep, line 414: Isn't it a breaking change to remove this?

packages/core/src/utils/GroupGraph.ts
```diff
@@ -0,0 +1,62 @@
...
+import Graph from "./Graph";
+
+export type GroupGraph = Graph<
+  string, // shape name
+  number // shape layer
+>;
+
+export const makeGroupGraph = <T extends Shape | ShapeAD>(
+  shapes: T[]
+): GroupGraph => {
```
samestep, line 11: Why is this generic?

```suggestion
export const makeGroupGraph = (shapes: ShapeAD[]): GroupGraph => {
```

packages/core/src/utils/GroupGraph.ts
```diff
@@ -0,0 +1,62 @@
...
+  const graph: GroupGraph = new Graph();
+  const nameShapeMap: { [k: string]: T } = {};
```
samestep, line 13: Since you're using this as a map, can you use [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)?

```suggestion
  const nameShapeMap = new Map<string, ShapeAD>();
```

samestep (changes requested):
A couple comments on `GroupGraph`

packages/core/src/shapes/Shapes.ts
```diff
@@ -185,6 +188,44 @@ const Text = ShapeDef({
   isPolygonlike: true,
 });
 
+// bboxFromGroup is here, not in engine/BBox, to prevent cyclic import
```
samestep, line 191: I thought JavaScript was generally OK with cyclic imports; could you clarify what exactly breaks when you try to put this in `engine/BBox`?

packages/core/src/engine/EngineUtils.ts
```diff
@@ -229,7 +258,8 @@ const valueADNums = (v: Value<ad.Num>): ad.Num[] => {
       return [v.contents];
     }
     case "BoolV":
-    case "StrV": {
+    case "StrV":
+    case "ShapeListV": {
```
samestep, line 262: Could you clarify why this is the empty list? Is the idea that, in `compileCompGraph` (the only place `valueADNums` is used), we're really just thinking of a flat list of shapes, instead of a tree of `Group`s? Perhaps you could add a code comment here explaining what's going on.

packages/core/src/compiler/Style.ts
```diff
@@ -2565,7 +2590,9 @@ const evalAccess = (
     case "ColorV":
     case "FloatV":
     case "PathDataV":
-    case "StrV": {
+    case "StrV":
+    case "ShapeListV": {
+      // Not allowing indexing into a shape list for now
```
samestep, line 2595: Maybe it would be better to make a new error type for this? The error message for `NotCollError` says "Cannot index into a non-collection", which would be somewhat confusing in this case.

packages/core/src/compiler/Style.ts
```diff
@@ -3046,11 +3077,103 @@ export const translate = (
...
+  layerGraph: Graph<string>,
+  groupGraph: GroupGraph
+): void => {
+  layerGraph.setEdge({ i: below, j: above, e: undefined });
+  const belowParents = groupGraph.parents(below);
+  const aboveParents = groupGraph.parents(above);
+  aboveParents
+    .filter((p) => p !== below)
+    .forEach((p) => {
+      propagateLayeringToParents({ below, above: p }, layerGraph, groupGraph);
```
samestep, line 3120: A bit unfortunate to do this with recursion, since we pretty much avoid recursion elsewhere throughout the compiler... not the end of the world though, I'm just curious if there's a way to do it without recursion (since we are operating over an explicit graph).

samestep (commented):
Some Style compiler comments

packages/core/src/renderer/Renderer.ts
```diff
@@ -184,28 +116,190 @@ export const RenderStatic = async (
...
+    canvasSize: [number, number];
+    variation: string;
+    pathResolver: PathResolver;
+  },
+  nameShapeMap: { [k: string]: Shape },
+  interactiveProp?: InteractiveProps
+) => {
+  const visited: Set<string> = new Set();
+  while (visited.size < graph.nodes().length) {
+    const rootName = findRoot(graph, visited);
```
samestep, line 290: This feels backwards; why not instead keep track of which ones we have yet to visit, and then at each step of this loop, check if that set is empty and then pick one to do next? (Also, why are we running graph algorithms in the renderer at all? Does the compiler not return the right kind of structure for the render to just use directly?)

samestep (changes requested):
A couple `Renderer` comments

packages/core/src/renderer/Renderer.ts
liangyiliang: The shape ordering outputted by the compiler does not respect the group ordering, only the layer ordering, since we run topsort only on the layer ordering.

liangyiliang: Of course, if we manage to get group ordering and layer ordering together into one ordering, then we can do what you said.

packages/core/src/utils/GroupGraph.ts
liangyiliang: Resolved - it used to be necessary, but not anymore.

packages/core/src/utils/GroupGraph.ts
liangyiliang: Done

packages/core/src/engine/EngineUtils.ts
liangyiliang: Yup I can add a comment.

packages/core/src/renderer/Renderer.ts
```diff
@@ -68,70 +50,17 @@ const getPosition = (
...
-      document.removeEventListener("mousemove", onMouseMove);
-      onDrag((shapeProps.shape.properties.name as StrV).contents, dx, dy);
-    };
-    document.addEventListener("mouseup", onMouseUp);
-    document.addEventListener("mousemove", onMouseMove);
-  };
-  g.addEventListener("mousedown", onMouseDown);
-  return g;
+const makeNameShapeMap = (shapes: Shape[]): { [k: string]: Shape } => {
+  return Object.fromEntries(
```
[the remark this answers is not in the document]
liangyiliang: Done!

samestep:
@liangyiliang could you update the PR description to describe the new semantics instead of the old saturation semantics?

samestep:
@liangyiliang you modified some newline at the end of `yarn.lock`; could you revert that?

samestep (changes requested):
I still need to review the rest of the code, but while I'm doing that, could you add information about `Group` to the docs on the website in this PR? You'll at least need to add a `packages/docs-site/docs/ref/style/shapes/group.md` file, but you should probably also put more info in `packages/docs-site/docs/ref/style.md` too.

packages/core/src/index.ts
```diff
@@ -415,9 +415,9 @@ export {
...
   RenderInteractive,
   RenderStatic,
+  RenderShape,
```
samestep, line 420: Since this change just moves the `RenderShape` export lower down in the list, could you revert it? (It just adds unnecessary noise to the diff.)

liangyiliang: Done

liangyiliang:
> @liangyiliang you modified some newline at the end of `yarn.lock`; could you revert that?

Done! 

> could you add information about Group to the docs on the website in this PR? You'll at least need to add a packages/docs-site/docs/ref/style/shapes/group.md file, but you should probably also put more info in packages/docs-site/docs/ref/style.md too.

Yup I will do it!

liangyiliang:
> I still need to review the rest of the code, but while I'm doing that, could you add information about `Group` to the docs on the website in this PR? You'll at least need to add a `packages/docs-site/docs/ref/style/shapes/group.md` file, but you should probably also put more info in `packages/docs-site/docs/ref/style.md` too.

Done!

packages/docs-site/.vitepress/config.ts
```diff
@@ -221,6 +221,7 @@ export default defineConfig({
                       link: "/docs/ref/style/shapes/rectangle",
                     },
                     { text: "Text", link: "/docs/ref/style/shapes/text" },
+                    { text: "Group", link: "/docs/ref/style/shapes/group" },
```
samestep, line 224: Awesome docs page, great work writing this up! Super small note, could you put this menu item in alphabetical order among the others? (Maybe add a comment here saying that these should stay in alphabetical order.)

packages/core/src/utils/CollectLabels.ts
```diff
@@ -285,8 +294,42 @@ const setPendingProperty = (
...
+  }
+};
+
 export const insertPending = (state: State): State => {
   const varyingValues = [...state.varyingValues];
+  insertPendingHelper(
+    state.shapes,
+    varyingValues,
+    state
+  ); /*
```
samestep, line 332: Could you delete this code instead of commenting it out?

samestep (approved):
Looks great modulo the remaining comments I left above! Thanks for pulling this through :)

packages/core/src/utils/CollectLabels.ts
liangyiliang: Done!

packages/docs-site/.vitepress/config.ts
liangyiliang: Done

packages/core/src/compiler/Style.ts
liangyiliang: Done

PR #1302, opened 2023-02-06
fix: improve performance of pseudoTopsort

### LINKED ISSUE #1215: `pseudoTopsort` has time complexity Ω(n²)
I realized today while working on #1211 that [`pseudoTopsort` repeatedly calls `fix`](https://github.com/penrose/penrose/blob/a7aeadff585ec62fa19eab21a69a666a1d8cd9a9/packages/core/src/compiler/Style.ts#L3049), whose [time complexity is linear](https://github.com/datastructures-js/heap/blob/v3.2.0/src/heap.js#L251-L261). We added this function in #760, and back then, [the docs](https://github.com/datastructures-js/heap/tree/v3.2.0#fix) incorrectly said that the method's time complexity is logarithmic. That [has since been corrected](https://github.com/datastructures-js/heap/tree/v4.3.1#fix).

YonahGoldberg:
# Description
Improving time complexity of pseudoTopsort

Resolves #1215 

PseudoTopsort had time complexity $\Omega(n^2)$ (really something more like $O(v^2*e)$) since it made repeated calls to `fix` on the heap used in the function. I switched it to using a priority queue, which has $O(logn)$ insertion and removal, so the overall time complexity is now something like $O(v * e * log(v))$.

I removed the dependency on @datastructures-js/heap and added a new one on @datastructures-js/priority-queue.
I also tested on some examples and everything is working well.

# Checklist


packages/core/src/compiler/Style.ts
```diff
@@ -152,6 +151,8 @@ import {
 } from "utils/Util";
 import { checkTypeConstructor, isDeclaredSubtype } from "./Domain";
```
samestep, line 153: Just a minor style nitpick: blank lines introduce groupings of `import`s, so I'd remove this one to allow this new `import` to be sorted in with the existing ones:

```suggestion
```

samestep (commented):
This looks great if it works, but I'm a bit confused how this is possible... could you explain? Specifically, the reason `fix` has to be called in the current implementation is because `toVisit` heap has no way of knowing when modifications to `indegree` should cause changes in the ordering; given that you still don't do anything special with `toVisit` to notify it of changes in this new version, that suggests to me that it's broken but just happens to pass the tests we currently have (so we should add more tests), but I could be mistaken.

samestep:
> just happens to pass the tests we currently have

Ah never mind, looks like our current tests do actually catch this. By the way, if you want to run the tests locally rather than having to wait for CI, you can do so by just running the `yarn` test command, as described [here](https://github.com/penrose/penrose/blob/115a6094025347e525ebe33a2207a944df8aa837/CONTRIBUTING.md#test).

YonahGoldberg:
That makes a lot of sense, my mistake. I adapted the datastructures-ts heap implementation so that it holds a map from heap nodes to their indices, so that after making modifications to `indegree` I can just called `heapifyUp` on the updated node to make the necessary swaps to increase its priority. All the tests are passing now.

I'm not sure if this is the most efficient implementation because every time I swap I have to update the map, although at least asymptotically it is now better.

samestep (changes requested):
Sorry for taking so long to review this; it looks good to me! A couple important points:

- run `yarn format`
- include a link to the specific commit and file of https://github.com/datastructures-js/heap that you copied
- more importantly, copy/paste the entirety of the `LICENSE` file from that repo into a comment at the top of `Heap.ts`

Could you go ahead and resolve those items? Then I'm good to merge this!

YonahGoldberg:
Should be all set now.

samestep (approved):
Looks great, thanks! Will merge once CI finishes.

PR #1306, opened 2023-02-08
feat: added "ctrl+enter" binding for recompiling

### LINKED ISSUE #1272: Provide non-Vim keyboard shortcut to recompile
Compare/contrast with #1046.

Dsm0:
# Description

Resolves #1272 

"ctrl+enter" now functions identically to `:w` in VimMode. 

# Checklist

### DIFF
--- a/packages/components/src/editing/EditorPane.tsx
+++ b/packages/components/src/editing/EditorPane.tsx
@@ -42,8 +42,11 @@ export default function EditorPane({
   const editorRef = useRef<editor.IStandaloneCodeEditor | null>(null);
   const statusBarRef = useRef<HTMLDivElement>(null);
 
-  if (monaco != null && onWrite != null){
-    editorRef.current?.addCommand(monaco.KeyMod.CtrlCmd | monaco.KeyCode.Enter, onWrite)
+  if (monaco != null && onWrite != null) {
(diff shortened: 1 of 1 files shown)


packages/components/src/editing/EditorPane.tsx
```diff
@@ -41,6 +41,14 @@ export default function EditorPane({
   const monaco = useMonaco();
   const editorRef = useRef<editor.IStandaloneCodeEditor | null>(null);
   const statusBarRef = useRef<HTMLDivElement>(null);
+
+  if (monaco != null && onWrite != null) {
```
wodeni, line 45: We generally prefer strict inequality (`===` and `!==`) in all cases. See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) for details.

wodeni (changes requested):
Looks great! Just one small comment below.

packages/components/src/editing/EditorPane.tsx
Dsm0: for some reason, `onWrite !== null` throws an error 
image
but `onWrite != null` doesn't
image
and I'm not sure why

PR #1310, opened 2023-02-13
feat: matrix and vector operations in Style

### LINKED ISSUE #765: Vector DSL in Style
Copying @keenancrane's post in https://github.com/penrose/penrose/issues/715#issuecomment-1006848444:

> ... in a lot of graphics systems there are often named types for 2D, 3D, and 4D vectors, as well as generic vectors of any length. Sometimes the 2-, 3-, and 4-dimensional types are specializations of or aliases to generic types; other times they are really their own independent type. But from a readability point of view it's much better to have these types. Also, many operations are defined only for vectors of a certain dimension. For instance, the cross product is only defined for a pair of 3D vectors; the argument is defined only for a 2D vector, and so on.
> 
> Some examples:
> 
> * **[GLSL](https://www.khronos.org/opengl/wiki/OpenGL_Shading_Language)** — The OpenGL shading language (GLSL) has [vector types](https://www.khronos.org/opengl/wiki/Data_Type_(GLSL)#Vectors) `vec2`, `vec3`, and `vec4`.  The syntax for GLSL vector operations is pretty slick, and has stood the test of time with literally [zillions of shaders](https://shadertoy.com/) happily written in this language.  One particularly nice feature is [_swizzling_](https://www.khronos.org/opengl/wiki/Data_Type_(GLSL)#Vectors), where you can construct a vector of any length (2–4) from the components of a vector of any other length, e.g., if you have `vec2 u` then you can write `vec2 v = u.xxyx` which will construct `v` using three copies of the x-coordinate and two copies of the `y`-coordinate.  Also super nice is that both `xyzw` and `rgba` are aliases to the components of the vector.  These aliases provide super nice hints about whether the vector is being used to describe a position or a color—while seamlessly allowing colors to be defined in terms of positions and vice versa (<—_super_ common in graphics!).  Finally, vectors of one length can be used to initialize vectors of another length.  For instance, you can write `vec3 x = (1,2,3)` and then `vec4 w = (x,1)`, which effectively appends a "1" to `x` in order to form `w`.  This kind of construction shows up all the time when working with homogeneous coordinates—or appending an alpha value to an RGB color!  Basically, GLSL is a DSL that is _very_ thoughtfully designed for graphics coding—even if some choices might initially seem nuts from the perspective of general PL (like: no type safety between positions and colors).
> * **[Eigen](https://eigen.tuxfamily.org/dox/group__TutorialMatrixClass.html)** — Eigen is a general-purpose numerical linear algebra library, but one that is strongly influenced by (and widely used in) graphics code.  It's pretty representative of the "C++ way" of defining vector types: there is a generic `Matrix` type (of any dimensions), which is specialized to a generic `Vector` type (of any length), which is further specialized to 2D, 3D, and 4D vector types `Vector2f`, `Vector3f`, and `Vector4f`.  This is all done using templates and typedefs.  Methods that work only with vectors of a certain dimension (like cross product) are implemented via [partial template specialization](https://en.cppreference.com/w/cpp/language/partial_specialization).
> * **[GeometryCentral](https://geometry-central.net/)** — GeometryCentral is a library we wrote here at CMU for digital geometry processing, which includes some vector types.  It's pretty representative of the way a lot of graphics-specific code is written: since we know we're almost exclusively working with [2D vectors](https://geometry-central.net/utilities/vector2/) and [3D vectors](https://geometry-central.net/utilities/vector3/), just write nice versions of those classes; don't sweat the fact that there's a little bit of code duplication.  Getting deep into templates and generalization is, in this case, just over-engineering.  One important design decision here is that 2D vectors support all the same operations as complex numbers.  And why not?  A complex number is nothing more than an "ordinary" (Euclidean) 2-vectors endowed with some additional operations like multiplication.  These complex operations turn out to be _super_ useful for writing graphics code.  For instance, a 2D rotation can be expressed as multiplication with a unit complex number, etc.  In the geometry processing context, general n-dimensional vectors most often come up when you need to solve a large linear algebra problem—at which point you call out to a general purpose linear algebra library (like Eigen) and use its types.
> 
> As for which paradigm we should adopt in Penrose, my 2¢ is that **the GLSL syntax is awesome and we should try to build on its strengths as much as possible**. It's tried-and-true for writing shader code, and Style programs are gonna often end up being very similar in spirit to shaders. I also think **it's important to support 2-, 3-, and 4-dimensional vectors** as explicit types like `vec2`, `vec3`, `vec4` (no matter how they're actually implemented internally). We can also have n-dimensional vectors—a.k.a. lists!—and define some basic operations on those, like adding two lists of equal length or reducing a list via a summation. But I wouldn't try to shoehorn _everything_ into a generic n-vector type.

keenancrane:
# Description

Resolves part of #765.

This PR adds basic n-dimensional matrix and vector operations to the Style language, via modifications to the parser and compiler.  With a couple exceptions, it does not add any new high-level library functions (which will be the focus of future PRs); just low-level operators.  It also does not add advanced accessors (e.g., swizzling).  It also includes a fairly thorough collection of test programs for each operation, which have been added to the registry (but are meant for development purposes only, and should not be added to the gallery).

image

**Additions to Style language:**

* operator `'` now performs transpose
* operator `*` now performs matrix-vector multiply
* operator `*` now performs vector-matrix multiply
* operator `*` now performs scalar-matrix multiply
* operator `*` now performs matrix-scalar multiply
* operator `*` now performs matrix-matrix multiply
* operator `+` now performs matrix-matrix sum
* operator `-` now performs matrix-matrix difference
* operator `.*` now performs elementwise vector-vector multiplication
* operator `./` now performs elementwise vector-vector division
* operator `.*` now performs elementwise matrix-matrix multiplication
* operator `./` now performs elementwise matrix-matrix division

**Methods added to `Autodiff.ts`:**

* `smmul` — scalar-matrix multiplication
* `msmul` — matrix-scalar division
* `vmmul` — left vector-matrix multiplication
* `mvmul` — right matrix-vector multiplication
* `mmmul` — matrix-matrix multiplication
* `mmadd` — matrix-matrix addition
* `mmsub` — matrix-matrix subtraction
* `mtrans` — matrix transpose
* `ewvvmul` — elementwise vector-vector multiplication
* `ewvvdiv` — elementwise vector-vector division
* `ewmmmul` — elementwise matrix-matrix multiplication
* `ewmmdiv` — elementwise matrix-matrix division
* `vouter` — vector-vector outer product

**Functions added to `Functions.ts`:**

* `cross` — 3D cross product (exposes existing `cross3` autodiff method)
* `outerProduct` — outer product of two vectors of equal length

As discussed in #765, choices made about operators and conventions are meant to largely follow well-loved DSLs used for vector and matrix manipulation, as found in GLSL, MATLAB, Mathematica, etc.  Some particular choices of note:

- The set of operations is largely guided by the spec for the [OpenGL Shading Language](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.40.pdf).  However, GLSL includes far more operators, many of which have no obvious use case for diagramming (e.g., bit-shifting and so on).  We should however keep using this spec as a reference for how things might be designed as we run into future issues and feature requests.
- As in GLSL, we adopt the convention that left vector-matrix multiplication `v*A` is equivalent to multiplying the transpose of `v` with `A` (whereas `A*v` gives ordinary matrix-vector multiplication).  This convention is also adopted by Mathematica.
- In many situations, it is very useful to be able to perform elementwise operations like `u/v` (meaning that each element of `u` is divided by the corresponding element of `v`, yielding a vector).  However, we do not adopt the convention of GLSL and Mathematica, which overload ordinary `*/` for elementwise multiplication and division.  Instead, we use `./` and `.*`, which is the syntax used in MATLAB.  The reason for this choice is simply that it's otherwise too easy to make hard-to-debug errors, e.g., thinking you're dividing by a scalar when really you're dividing by a vector.  The `./` and `.*` syntax is basically saying, _"are you sure you really want to divide one vector by another?"_.  (Note that in the parser these operators appear before `.`, to help avoid parse ambiguity/errors.)
- An ASCII `'` is used for the matrix transpose, rather than a Unicode character (like `ᵀ`); there is also no Unicode sugar.  This decision came from discussions with @wodeni, that basically it's better to leave Unicode sugar for user-defined domains/substance programs.  The single quote is a standard convention in MATLAB and other languages, and is often used in online ASCII-only discussions (e.g., over email) to signify transpose.

There are a bunch of test examples in `examples/matrix-ops`.  Each example applies an operator to some fixed data, so that the output can easily be checked against diagrams generated using trusted packages like Mathematica.  See the README for that example for more info.

# Checklist

# Open questions

These operations are not specialized to matrices or vectors of any particular dimension, and effectively let Style manipulate arbitrary n x n dense matrices.  This is mostly a good thing, but especially coders new to graphics have a tendency to get themselves in trouble by manipulating _very large_ dense matrices (e.g., when they really should be using sparse matrices).  We might want to issue some kind of warning about operations that are likely to be slow, or perhaps even artificially cap matrix sizes to some large n in order to avoid absurdly slow O(n²) programs.

Future PRs may still wish to investigate swizzling.  We definitely want to add some convenience functions to the standard library—especially for constructing transformation matrices (i.e., homogeneous rotations, translations, etc.), since such matrices provide a nice solution to some long-standing questions about how to design transformations in Style.


samestep (approved):
This is awesome, thanks Keenan! No complaints on my end, just one small typo fix. As a followup it would be awesome to surface dimension-mismatch errors as Style errors rather than just throwing, but that falls under our more general problem that we don't have great typechecking for usage of builtin computation functions etc.

Issue #1307, opened 2023-02-09
References in CSS style of imported SVGs not rendered correctly


wodeni:
**Describe the bug**

In `editor`, open an example with imported SVGs that contain CSS style (often exported from Adobe Illustrator). When rendering variations of diagrams on a grid, the original diagram on the diagram panel will lose all the images. 

**To Reproduce**

![bug-image](https://user-images.githubusercontent.com/11740102/217836951-52891120-b293-4f17-845b-6ab41c836a09.gif)

1. Go to penrose.cs.cmu.edu/try
2. Click on the walk-on-spheres example
3. Select "Diagram Variations"
4. Select "Diagram"
5. Observe that the images disappear

**Additional context**

The problem here is `makeIdsUnique` doesn't transform CSS classes. Since most examples contain one imported image, this look okay in most cases. But if there are multiple images imported with different styling for the same class name, cascading rules will make every single element of the class take the same style. 

Looks like a possible solution is to inline the CSS style: https://github.com/airbnb/babel-plugin-inline-react-svg/issues/44.

Similarly, [an svgo plugin to do the same thing](https://github.com/svg/svgo/blob/main/plugins/inlineStyles.js)

PR #1316, opened 2023-02-14
fix: nondeterminism in renderer

### LINKED ISSUE #1312: Determinism is broken
#1287 broke determinism, as mentioned in https://github.com/penrose/penrose/pull/1294#issuecomment-1424859484.

wodeni:
# Description

Resolves #1312.

* In `Line.ts`, stop using `uuid.v4` for unique ID generation
* Instead, `Render*` functions take in `name`, a client-provided id for the diagram instance to be rendered. And names in the SVG are disambiguated from others instances by a unique instance id. 
* Similarly, `Simple` and `GridBox` take in `name` too.
* Remove `uuid` from `core`'s dependencies
* Note that this PR doesn't solve #1307

# Checklist

# Open questions

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -163,10 +163,10 @@ const stepUntilConvergenceOrThrow = (state: State): State => {
 /**
  * Embed a static Penrose diagram in a DOM node.
  *
- * @param domainProg a Domain program string
- * @param subProg a Substance program string
- * @param styProg a Style program string
+ * @param prog a Penrose trio and variation
  * @param node a node in the DOM tree
+ * @param pathResolver a resolver function for fetching Style imports
+ * @param name the name of the diagram
  */
 export const diagram = async (
   prog: {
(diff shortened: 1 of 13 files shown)


samestep:
Looks like `packages/core/package.json` changed but `yarn.lock` didn't; could you run `yarn`?

packages/core/src/index.ts
```diff
@@ -176,13 +176,14 @@ export const diagram = async (
     variation: string;
   },
   node: HTMLElement,
-  pathResolver: PathResolver
+  pathResolver: PathResolver,
+  name: string
```
samestep, line 180: To make this a non-breaking change, we need to make this `name` parameter optional.

packages/core/src/index.ts
```diff
@@ -207,14 +208,16 @@ export const interactiveDiagram = async (
...
```
samestep, line 212: Same here.

samestep (changes requested):
Quick note about backwards compatibility before I review the rest.

samestep:
I guess the `name` prop for the `Simple` component also needs to be optional.

samestep (approved):
Looks great, thanks!

PR #1328, opened 2023-02-28
feat: provide a `shapeDistance` function

samestep:
# Description

This PR adds a new function called `shapeDistance` which computes the signed distance from the origin on the Minkowski sum of the first shape and the negative of the second shape.

This works on all shapes. It uses the following approximations:

- when both shapes are `Circle`, we are exact
- when both shapes are `Rectlike`, we are exact
- when one shape is `Text` or `Equation` and the other is `Line`, we use the bounding circle approximation added by @wodeni in #1117
- when both shapes are `Polygonlike`, we use the intersection-of-half-planes approximation added by @jiriminarcik in #810
- when one shape is `Rectlike` and the other is `Circle`, we are exact
- when one shape is `Polygonlike` and the other is `Ellipse`, I'm not exactly sure what we do; @jiriminarcik do you happen to know if this is an approximation?
- when one shape is `Circle` and the other is `Line`, I believe we are exact
- in all other cases, we approximate both shapes by their respective axis-aligned bounding boxes

I factored out most of the existing `overlapping` constraint. Also, for better typechecking, I added a `ShapeTuple` type representing the pair format expected by objectives and constraints, and a `shapeTupleToShape` function that converts this to just a `Shape`. In the same file, I also added `Rectlike` and `Polygonlike` types along with `isRectlike` and `isPolygonlike` type predicates to go along with those.

This patch made the `overlapping` constraint more correct for the case where one shape is a circle or an ellipse and the other is a rectangle, so I corrected the tests for that case.

# Checklist

# Open questions

I didn't factor out the cases where one shape is an ellipse and the other is either a circle or an ellipse, because I don't know how to use our existing code to actually calculate the distance there.


joshsunshine:
@samestep, please describe what shapes this works on, what approximations (if any) are used, etc. in the PR description.

samestep:
@joshsunshine done!

liangyiliang (approved):

joshsunshine (approved):

PR #1332, opened 2023-03-04
feat: compute rect-line distance exactly

samestep:
# Description

This PR provides a `rectLineDist` function which differentiably computes the _exact_ signed distance from the origin to the Minkowski difference of a rectangle and a line, and modifies the implementation of the `shapeDistance` function added by #1328 to use this when one shape is a `Line` and the other is an `Equation` or `Text` (as opposed to the previous implementation, which used a bounding circle as an approximation instead of a bounding box). I also added a new example to the registry to demonstrate visually that this works.

The reason to export a new `rectLineDist` function, instead of just using `shapeDistance`, is that sometimes we want to phrase an objective or constraint as if something is a straight line segment, even if the actual shape being drawn is something else, like a quadratic Bézier curve in the case of #1317. Using `shapeDistance` requires you to actually have a genuine `Line` shape in the diagram.

Also, we didn't have an `xor` autodiff function before, so I added one here. While I was doing that, I noticed that some comments in the autodiff types file were out of date, so I updated them.

I changed the variation for `parallel-lines-euclidean` because for some reason this patch was causing it to take a long time to optimize (almost ten seconds) with the old variation.

# Checklist


joshsunshine:
> modifies the implementation of the `shapeDistance` function added by https://github.com/penrose/penrose/pull/1328 to use this when one shape is a Line and the other is an `Equation` or `Text`
After reading the code, it looks like the `shapeDistance` function calls `rectLineDist` for all `RectLike` shapes which also includes `Rectangle` and `Image`. @samestep, is that correct?

samestep:
@joshsunshine No, that is not correct:

https://github.com/penrose/penrose/blob/af2f8b567316935365d18f18b7e17c0fcea74e89/packages/core/src/contrib/Queries.ts#L275

joshsunshine:
I am glad I asked. Why didn't you add `Rectangle` and `Image` to that `else if`? Is there a reason this doesn't work for those shapes?

samestep:
Nope, no reason not to do that! Do you want me to do it in this PR, or in a followup?

joshsunshine:
> Nope, no reason not to do that! Do you want me to do it in this PR, or in a followup?

This PR.

PR #1333, opened 2023-03-07
fix: avoid `EPS_DENOM` in core autodiff

samestep:
# Description

Prior to this PR, `inverse` (both its implementation and its derivative) added `EPS_DENOM = 10e-6` before taking the reciprocal. It seems bad to have this sort of arbitrary tweak in the implementation of a basic autodiff primitive, so this PR removes that and also simplifies several of our other derivative definitions.

A couple of our test cases initially broke, so I fixed them by manually adding `EPS_DENOM` to the implementations of our `repelPt` and `repelScalar` objectives.

I also had to change the variation for `circle-example-euclidean`, because that particular example is pretty unreliable. I tried resampling that example a few times on `main` and I got similarly bad results for many variations, so the issue wasn't introduced by this PR.

# Checklist

# Open questions

- How can we remove the usage of `EPS_DENOM` from the derivative of `sqrt`? I tried in this PR and it caused a lot of failures, so I reverted that change in favor of just doing the others for now: https://github.com/penrose/penrose/actions/runs/4355028613/jobs/7611098998

### DIFF
--- a/packages/core/src/contrib/Objectives.ts
+++ b/packages/core/src/contrib/Objectives.ts
@@ -150,7 +150,12 @@ export const objDictGeneral = {
     } else {
       // Repel any two shapes with a center.
       // 1 / (d^2(cx, cy) + eps)
-      res = inverse(ops.vdistsq(shapeCenter([t1, s1]), shapeCenter([t2, s2])));
+      res = inverse(
+        add(
+          ops.vdistsq(shapeCenter([t1, s1]), shapeCenter([t2, s2])),
+          EPS_DENOM
+        )
+      );
     }
 
     return mul(res, repelWeight);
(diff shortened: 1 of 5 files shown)


wodeni (approved):
Adding epsilon is fairly common in graphics code, but I guess the question is if we want AD to do it or constraint/function implementors. I'm okay with the latter, as long as this is clearly documented somewhere.

edit: now that I think about it. Can you check if the example in the [docs page](https://penrose.cs.cmu.edu/docs/ref/constraints#objectives-example-circle-repel) still work? Also I'm surprise that `notTooClose` doesn't need the extra epsilon.

samestep:
@wodeni Yeah I'd really prefer the latter, because that way you actually get to choose based on the particular function you're writing; if epsilon is built into an autodiff primitive then there's no way to opt out.

PR #1337, opened 2023-03-09
refactor!: consolidate shape types

### LINKED ISSUE #1282: Consolidate shape types

liangyiliang:
# Description

Resolves #1282 .

Previously we had three different representations of shapes:

* `type GPI<T> = { tag: "GPI", contents: [string, Properties<T>] }`.
* `type GenericShape<T> = { shapeType: string; properties: Properties<T> }` (where `Shape = GenericShape<number>` and `ShapeAD = GenericShape<ad.Num>`).
* `type Shape = Circle | Ellipse | ... | Group`.

We keep the third representation (making it parametric with `type Shape<T> = Circle<T> | Ellipse<T> | ... | Group<T>`), removing the other two, since the third representation utilizes TypeScript's type system.

# Shape checker

We take `Circle` as an example. We now have

```typescript
type Circle<T> = ShapeCommon<T> & {
  shapeType: "Circle";
} & CircleProps<T>;

interface ShapeCommon<T> {
  shapeType: string;
  passthrough: Map<string, StrV | FloatV<T>>; // ignore this for now
}

interface CircleProps<T> extends Named<T>, Stroke<T>, Fill<T>, Center<T> {
  r: FloatV<T>;
}
```
where `Named<T>, Stroke<T>, Fill<T>, Center<T>` are the shape-hierarchy interfaces.

We construct a `Circle` object during the translation process (during compilation), and the `Circle` object is built from the symbols in the translation. Since `Circle<T>` requires parameters in `Named<T>, ..., Center<T>`, we can build a Circle as follows:
```
procedure checkCircle(path, translation) {
  named = checkNamed(path, translation);
  stroke = checkStroke(path, translation);
  fill = checkFill(path, translation);
  center = checkCenter(path, translation);

  // additionally check that the `r` property is FloatV
  r = checkProp(path + ".r", FloatV, translation); 

  if any fails, return error;
  return { ...named, ...stroke, ...fill, ...center, r, shapeType: "Circle" };
}
```
where, for example, `checkNamed` checks the follows:
```
procedure checkNamed(path, translation) {
  name = checkProp(path + ".name", StrV, translation);
  style = checkProp(path + ".style", StrV, translation);
  ensureOnCanvas = checkProp(path + ".ensureOnCanvas", BoolV, translation);

  if any fails, return error;
  return { name, style, ensureOnCanvas };
}
```

So, `checkCircle(path, translation)` both checks that every property of the circle are correctly typed, grabs those properties from the translation, and returns a `Circle` object.

The nice thing about this is that we can simply access things like
```typescript
const c: Circle<ad.Num> = ...;
c.r // TypeScript know this is FloatV and hence can access .content, etc.
c.name // TypeScript know this is StrV
c.ensureOnCanvas // TypeScript knows this is BoolV
...
```

## Implicit Casting
Some `Value<T>` types can be casted into each others.

* `ListV` and `VectorV` are implicitly casted into each other.
* `MatrixV`, `LListV`, and `PtListV` are implicitly casted into each other.
* `TupV` can be casted into `ListV` and `VectorV`, but not vice versa.

So if a shape parameter expects `PtListV`, it will also be able to take `MatrixV` and `LListV`.

# Constraints and objectives

Since we removed GPIs, constraints and objectives no longer take GPIs. They take actual shape objects. For example,

```typescript
/**
 * Require that circle `s1` overlaps ellipse `s2` with some overlap `overlap`.
 */
export const overlappingCircleEllipse = (
  s1: Circle<ad.Num>,
  s2: Ellipse<ad.Num>,
  overlap: ad.Num = 0
): ad.Num => {
  // ...
};
```

# Passthrough

Each shape has a `passthrough` field storing all the passthrough properties:
```typescript
passthrough: Map<string, CanPassthrough<T>>
```
where
```typescript
type CanPassthrough<T> = StrV | FloatV<T>;
```

The `passthrough` field is populated during compilation as the compiler iterates through each additional, previously unadded item in the translation, checks that they are correctly typed, and fills in `passthrough`.

Notice that now we require each passthrough field to have either `StrV` or `FloatV` type.

# Checklist

# Open questions

Would be nice if we have a fully-fledged type system.

The `toString` function would also help a lot.


wodeni (changes requested):
Did a rough pass and things are looking good! Some minor comments below, of which the `sampleShape` export is the most important to me.

wodeni (changes requested):
Did a rough pass and things are looking good! Some minor comments below, of which the `sampleShape` export is the most important to me.

wodeni (changes requested):
Did a rough pass and things are looking good! Some minor comments below, of which the `sampleShape` export is the most important to me.

packages/core/src/utils/shapeChecker/CheckShape.ts
```diff
@@ -0,0 +1,484 @@
...
+  checkFloatV,
+  checkPathDataV,
+  checkShapeListV,
+  checkStrV,
+  checkVectorV,
+} from "./CheckValues";
+
+const { err, ok } = Result;
+
+export const checkShape = (
```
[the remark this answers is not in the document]
liangyiliang: Sure I think I can move that into `compiler` folder.

packages/core/src/utils/shapeChecker/CheckShapeHierarchyProps.ts
```diff
@@ -0,0 +1,224 @@
...
+    style: style.value,
+    ensureOnCanvas: ensureOnCanvas.value,
+  });
+};
+
+export const checkStroke = (
+  path: string,
+  trans: Translation
+): Result<Stroke<ad.Num>, StyleError> => {
+  const strokeWidth = checkProp(path, "strokeWidth", trans, checkFloatV);
```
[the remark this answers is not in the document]
liangyiliang: I'll see what @samestep thinks of this but I agree with the general idea of removing boilerplates.

packages/automator/index.tsx
```diff
@@ -6,11 +6,12 @@ import {
   makeCanvas,
   prepareState,
   RenderStatic,
-  shapedefs,
+  shapeTypes,
   showError,
   simpleContext,
   stepUntilConvergence,
 } from "@penrose/core";
+import { shapeSampler } from "@penrose/core/dist/shapes/Shapes";
```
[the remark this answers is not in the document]
liangyiliang: Done

liangyiliang:
@wodeni Hi Nimo, everything else is ready, except the reduction of boilerplate using exceptions. @samestep said that he would still need to review, but he would "lean toward leave it as-is".

wodeni (approved):

wodeni:
> @wodeni Hi Nimo, everything else is ready, except the reduction of boilerplate using exceptions. @samestep said that he would still need to review, but he would "lean toward leave it as-is".

All other changes seem good to me. Yeah I'd agree with @samestep here. Not worth the time to do that now tbh.

packages/docs-site/docs/ref/style/shapes-overview.md
```diff
@@ -30,3 +30,38 @@ Notice that this definition, the `center:` attribute of the circle is not set. U
...
+
+### Implicit Casting
+
+For convenience, types which are similar can be casted into each other, implicitly. If type `A` implicitly casts into type `B`, then a shape field that expects value of type `A` will also accept values of type `B`. Currently, we support the following:
+
+- `ListV` and `VectorV` are implicitly casted into each other.
+- `MatrixV`, `LListV`, and `PtListV` are implicitly casted into each other.
+- `TupV` can be casted into `ListV` and `VectorV`, but _not_ vice versa.
+
+Unfortunately, conversion of types into strings (typed `StrV`) are not yet supported.
```
samestep, line 67:
```suggestion
Conversion of types into strings (typed `StrV`) is not currently supported.
```

samestep (changes requested):
A few minor doc comments before I get to the rest of the diff.

packages/docs-site/docs/ref/style/value-types.md
```diff
@@ -0,0 +1,20 @@
+# Value Types
+
+A _value_ in Penrose have the following categories:
```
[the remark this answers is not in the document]
liangyiliang: Done

packages/docs-site/docs/ref/style/value-types.md
```diff
@@ -0,0 +1,20 @@
...
+
+- Numbers (type `FloatV`), constructed as either a numerical value or `?` signifying the unknown value to be determined during optimization (see [here](usage#unknown-scalar)).
+- Booleans (type `BoolV`), constructed as either `true` or `false`.
+- Strings (type `StrV`), constructed as text written in between double-quotes (like `"hello world"`), and concatenations (like `"\text{" + x.label + "}"`)
+- Path data (type `PathDataV`), constructed as in [here](shapes/path#defining-an-svg-path-using-d).
+- List of points (type `PtListV`), constructed using implicit casting from matrices or nested lists.
```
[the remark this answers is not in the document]
liangyiliang: Instead of talking about implicit casting, I'll just say that it is constructed as a matrix or a list of lists.

packages/core/src/compiler/shapeChecker/CheckShape.ts
```diff
@@ -0,0 +1,484 @@
...
+  checkPathDataV,
+  checkShapeListV,
+  checkStrV,
+  checkVectorV,
+} from "./CheckValues";
+
+const { err, ok } = Result;
+
+export const checkShape = (
+  shapeType: string,
```
samestep, line 43: Would it make sense to use `ShapeType` here instead of `string`? (genuinely just asking)

packages/core/src/types/errors.ts
```diff
@@ -446,6 +447,14 @@ export interface UOpTypeError {
   arg: Value<ad.Num>["tag"];
 }
 
+export interface BadShapeParamTypeError {
+  tag: "BadShapeParamTypeError";
+  path: string;
+  value: Val<ad.Num> | ShapeVal<ad.Num>;
+  expectedType: string;
+  passthrough: boolean;
+}
```
samestep, line 456: If I understand correctly, this error type doesn't include source location information; would it be possible to add that?

packages/core/src/shapes/Shapes.ts
```diff
@@ -2,238 +2,118 @@ import { input } from "../engine/Autodiff";
...
-  bbox: (properties: any) => BBox.BBox;
-  isLinelike?: boolean;
-  isRectlike?: boolean;
-  isPolygonlike?: boolean;
-}): ShapeDef => {
-  const sampler = (context: Context, canvas: Canvas) =>
-    shapedef.sampler(context, canvas) as Properties;
+const shapeSampler: {
+  [k in ShapeType]: (context: Context, canvas: Canvas) => ShapeProps<ad.Num>;
+} = {
```
samestep, line 73: Would it make sense to use `Map` instead of an object type here?

packages/core/src/contrib/Utils.ts
```diff
@@ -27,31 +27,32 @@ import { Line } from "../shapes/Line";
...
 import * as ad from "../types/ad";
 
-export type ShapeTuple = [ShapeType, Omit<Shape, "shapeType">];
-
-export const shapeTupleToShape = ([t, s]: ShapeTuple): Shape =>
-  ({ shapeType: t, ...s } as Shape);
-
 // these are redundant with the `isRectlike` and `isPolygonlike` fields on
 // `ShapeDef`, which we want to remove as part of a larger consolidation effort:
 // https://github.com/penrose/penrose/issues/1282
```
samestep, line 36: I think this comment can be removed now?
```suggestion
```

packages/core/src/renderer/Equation.ts
```diff
@@ -52,20 +53,20 @@ const Equation = ({ shape, canvasSize, labels }: ShapeProps): SVGGElement => {
...
     attrToNotAutoMap.push("string");
     elem.appendChild(txt);
 
     // Map the attributes we have
     attrToNotAutoMap.push(...attrFill(shape, elem));
     attrToNotAutoMap.push(...attrWH(shape, elem));
-    attrToNotAutoMap.push(...attrStroke(shape, elem));
-    attrToNotAutoMap.push(...attrFont(shape, elem));
+    // attrToNotAutoMap.push(...attrStroke(shape, elem));
+    // attrToNotAutoMap.push(...attrFont(shape, elem));
```
samestep, line 64: Probably delete these lines instead of commenting out?
```suggestion
```

samestep (approved):
Looks great, thank you so much for doing this! Just had a few minor comments/questions.

packages/core/src/compiler/shapeChecker/CheckShape.ts
liangyiliang: Yes, will make that change.

packages/core/src/renderer/Equation.ts
liangyiliang: Yes

packages/core/src/compiler/Style.test.ts
```diff
@@ -1,5 +1,5 @@
 import setTheory from "@penrose/examples/dist/set-theory-domain";
-import im from "immutable";
+import * as im from "immutable";
```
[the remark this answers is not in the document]
liangyiliang: Fixed

packages/core/src/types/errors.ts
liangyiliang: It does not contain source location information, and I'm not sure how I can obtain them ... I don't think those are in `translation` where we get the actual values for the shape parameters.

PR #1338, opened 2023-03-12
feat!: decouple optimizer from codegen

samestep:
# Description

As of #1092, our autodiff codegen is very tightly coupled with the `@penrose/optimizer` package. This PR decouples them by removing all Wasm codegen stuff from `@penrose/optimizer`, separating the Wasm memory of the optimizer from the memory of the generated autodiff Wasm module, and switching the dispatch mechanism from a direct usage of the `__indirect_function_table` (requiring the `--keep-lld-exports` flag) to the more standard way of simply passing an arbitrary JS function into Rust and calling it.

As an aside, I updated `CONTRIBUTING.md` and `packages/optimizer/Cargo.toml` to note that the `wasm-bindgen` version must be _exactly_ 0.2.84, because any version mismatch between the library and CLI tool is automatically detected and causes an error.

# Checklist

# Open questions

This is not and should not be the final form of the `@penrose/optimizer` interface. Now that we've decoupled the optimizer from codegen, I'd like to next decouple it from the staged layout functionality we added in #1199, removing the concept of a mask from the Rust codebase and instead handling it entirely on the `@penrose/core` side now that we have the flexibility to pass a more sophisticated JavaScript function/closure to the optimizer instead of being forced to directly pass an exported WebAssembly function that must explicitly take the mask as input.


packages/core/src/engine/Autodiff.ts
```diff
@@ -1747,6 +1752,149 @@ const genBytes = (graphs: ad.Graph[]): Uint8Array => {
...
+    ),
+  },
+});
+
+const makeCompiled = (
+  graphs: ad.Graph[],
+  meta: Metadata,
+  instance: WebAssembly.Instance
+): Gradient => {
+  const f = instance.exports[exportFunctionName] as (
```
wodeni, line 1863: Can you document why `f` here takes a bunch of `number`s whereas `GradFn` (i.e. `wrapped` below) takes arrays of floats or ints? I'm assuming this has something to do with how `WebAssembly.Instance.exports` works? The [docs on exports](https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Instance/exports) are kinda sparse for me to understand.

packages/core/src/engine/Autodiff.ts
```diff
@@ -1755,21 +1903,25 @@ const genBytes = (graphs: ad.Graph[]): Uint8Array => {
  * @param graphs an array of graphs to compile
  * @returns a compiled/instantiated WebAssembly function
  */
-export const genCode = async (...graphs: ad.Graph[]): Promise<Gradient> =>
-  await Gradient.make(
+export const genCode = async (...graphs: ad.Graph[]): Promise<Gradient> => {
+  const meta = makeMeta(graphs);
```
wodeni, line 1907: My understanding is this allocates WASM memory in the autodiff module (not `@penrose/optimizer`) with built-in functions, which is then imported by the compiled big autodiff comp graph function.

packages/optimizer/src/lib.rs
```diff
@@ -1,16 +1,22 @@
...
-
 use log::Level;
 use serde::{Deserialize, Serialize};
 use ts_rs::TS;
 use wasm_bindgen::{prelude::wasm_bindgen, JsValue};
 
 type Bool = i32; // 0 or 1 (`wasm-bindgen` doesn't support `bool` well)
 
-type Vector = nalgebra::DVector<f64>;
+#[wasm_bindgen(module = "/call.js")]
```
wodeni, line 8: Can you explain what `call.js` does and what the `extern "C"` means here? I can see it generates a `call.js` file but not sure what it's for.

wodeni (commented):
Looks great. The high-level goal you described seems good to me. I left a bunch of questions about the implementation. I'm sure most of them are just because I don't understand how the interops work exactly (both wasm-wasm between modules and rust-js translation). You can either reply to me here or add inline comments if you think the question is relevant enough.

packages/core/src/utils/Wasm.ts
```diff
@@ -62,7 +62,7 @@ export const OP = {
     or: 0x72,
     xor: 0x73,
   },
-  local: { get: 0x20, set: 0x21, tee: 0x22 },
+  local: { get: 0x20, set: 0x21 },
```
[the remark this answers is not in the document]
samestep: Yeah I just removed `tee` because it happens to be unused as of this PR. It doesn't matter obviously; I could leave it and it would be fine.

packages/core/src/engine/Autodiff.ts
samestep: Correct, the `makeMeta` function creates a Wasm memory big enough to hold all the memory we use for the generated computation function; then when we instantiate our generated Wasm module, we pass it that memory object.

packages/optimizer/source.ts
```diff
@@ -105,95 +36,28 @@ export interface Outputs<T> {
...
+  mask: Int32Array,
+  gradient: Float64Array,
+  secondary: Float64Array
+) => number;
+
 export class Gradient {
-  private f: WebAssembly.ExportValue;
-  private numAddends: number;
-  private numSecondary: number;
+  f: GradFn;
```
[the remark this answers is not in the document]
samestep: In our particular case, this `GradFn` is not actually the exported function from our generated Wasm module, but a JavaScript function that wraps around it so at the boundary we can think in terms of arrays instead of directly in terms of pointers.

packages/optimizer/src/lib.rs
samestep: I added a comment attempting to explain this; hopefully it clarifies?

packages/core/src/engine/Autodiff.ts
samestep: I wrote comments explaining these two signatures; does this help?

wodeni: Gotcha. I wasn't sure they were supposed to be pointers because the variable names said `offset` a lot. Makes sense to me!

packages/optimizer/src/lib.rs
wodeni: Gotcha. Weird that `extern "C"` is the standard lol but great to know about the `call4` motivation.

packages/core/src/engine/Autodiff.ts
samestep: Yeah I guess the words `offset` and `pointer` are fairly related but not directly synonymous 🤔

packages/optimizer/source.ts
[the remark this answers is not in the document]
wodeni: Gotcha. So really `makeMeta` figures out where the pointers should point to, and `makeCompiled` translates from array args to pointers based on the offsets computed by `makeMeta`.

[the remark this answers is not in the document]
samestep: Yeah exactly, so that all happens in the autodiff module, so `@penrose/optimizer` doesn't need to think about pointers at all.

PR #1341, opened 2023-03-15
fix: unique line arrow

liangyiliang:
# Description

On Chrome, the `arrowheads-arrowheads` example breaks and shows this:
image
where every arrow head is the same.

The issue is that while rendering a `Line` shape's arrowheads, we refer to the actual arrowhead `marker` SVG using an ID. Previously, the IDs are constructed as follows:

```
  const unique = `${namespace}-${variation}`;
  const startArrowId = unique + "-startArrowId";
  const endArrowId = unique + "-endArrowId";
```

It appears that `unique` only depends on `namespace` and `variation` and nothing else. If there are multiple lines on the screen, they may be rendered with the same arrowheads. This is not desirable.

The fix essentially adds the shape name (guaranteed to be unique) to `unique`:
```
  const unique = `${namespace}-${variation}-${
    (shape.properties.name as StrV).contents
  }`;
  const startArrowId = unique + "-startArrowId";
  const endArrowId = unique + "-endArrowId";
```

### DIFF
--- a/packages/core/src/renderer/Line.ts
+++ b/packages/core/src/renderer/Line.ts
@@ -156,7 +156,9 @@ const Line = ({
   );
   const elem = document.createElementNS("http://www.w3.org/2000/svg", "g");
   // an unique id for this instance is determined by the variation and namespace
-  const unique = `${namespace}-${variation}`;
+  const unique = `${namespace}-${variation}-${
+    (shape.properties.name as StrV).contents
+  }`;
   const startArrowId = unique + "-startArrowId";
   const endArrowId = unique + "-endArrowId";
   if (startArrowhead) {


wodeni (approved):
Interestingly I’ve never encountered this on any of Chrome, Safari, nor Firefox on macOS. I wonder why. Good catch though!

liangyiliang:
It seems that this bug was introduced when we employed `uuid`. See [this](https://github.com/penrose/penrose/blame/1895d8febdf4ced698bd1a478c3f3eaf1452b511/packages/core/src/renderer/Line.ts#L154).

Before having `uuid`, it was
```
const startArrowId = shape.properties.name.contents + "-startArrowId";
const endArrowId = shape.properties.name.contents + "-endArrowId";
```

wodeni:
> It seems that this bug was introduced when we employed `uuid`. See [this](https://github.com/penrose/penrose/blame/1895d8febdf4ced698bd1a478c3f3eaf1452b511/packages/core/src/renderer/Line.ts#L154).
> 
> 
> 
> Before having `uuid`, it was
> 
> ```
> 
> const startArrowId = shape.properties.name.contents + "-startArrowId";
> 
> const endArrowId = shape.properties.name.contents + "-endArrowId";
> 
> ```

I know that. But the behavior you posted in the PR description never showed up on my environment (various browsers + Ubuntu/macOS), before or after the uuid change and even in the very diffs in this PR. I'm guessing this is somehow platform specific, but I'm glad that we addressed this now.

Let's merge it!

liangyiliang:
Interesting! Sounds good, I'll merge it.

PR #1351, opened 2023-04-03
fix: corrected `triangle` function signature

### LINKED ISSUE #1350: broken `triangle` function signature
In `Functions.ts` the `triangle` function still takes in `GPI` objects. Its function signature is
```
  triangle: (
    _context: Context,
    [t1, l1]: any,
    [t2, l2]: any,
    [t3, l3]: any
  ): PathDataV<ad.Num> => { ... }
```
which does not work with the newly consolidated shape types.

liangyiliang:
# Description

Resolves #1350.

This PR uses the actual shape types in the function signature of the library function `triangle`. Previously, each parameter is a `ShapeTuple` which was removed as a result of the consolidation shape types. For some reasons, this got left in.

### DIFF
--- a/packages/core/src/contrib/Functions.ts
+++ b/packages/core/src/contrib/Functions.ts
@@ -1054,22 +1054,17 @@ export const compDict = {
    */
   triangle: (
     _context: Context,
-    [t1, l1]: any,
-    [t2, l2]: any,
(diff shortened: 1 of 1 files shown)


samestep (approved):
LGTM, thanks for catching this!

PR #1352, opened 2023-04-19
feat: function as data

liangyiliang:
# Description

This PR extends Penrose's Style function / constraint / objective library so that each function also includes metadata information about its documentation and parameters and their types, accessible by the Style compiler and other components of Penrose. The benefit are:
* With knowledge of the types of parameters that each function takes in, we can type-check whenever we call a function to make sure that each argument is of the correct type.
* We can also automatically generate the documentation pages using the function documentations.

# Example Function

Previously, the function library is essentially a mapping from the function name to the function itself with no information about the types of the arguments. As an example, the `equal` objective would look like:
```typescript
  equal: (x: ad.Num, y: ad.Num): ad.Num => squared(sub(x, y))
```
With function-as-data, the same function is defined as follows:
```typescript
  equal: {
    name: "equal",
    description: "Encourage the inputs to have the same value: `(x - y)^2`",
    params: [
      { name: "x", description: "First value", type: realT() },
      { name: "y", description: "Second value", type: realT() },
    ],
    body: (x: ad.Num, y: ad.Num): ad.Num => squared(sub(x, y)),
  },
```
This gives us the following information:
* The function's name is `equal`.
* The function's description.
* For each parameter, the parameter name, the description, and the type. For example, parameter `x` has description of "First value" and takes in a real number.
* The function body.

# New Types and Interfaces

## Types

A type can be a value type, a shape type, or a union type.
* A value type (type `ValueT` in TypeScript) is one of: (all are fully type-checked except specially denoted)
  * `Real`: a real number
  * `Unit`: a real number within [0, 1] (we only type-check that it is a real and do not check the actual range)
  * `PosInt`: a positive integer (we only type-check that it is a real)
  * `Nat`: a natural number (we only type-check that it is a real)
  * `Real2`: a 2d vector of reals
  * `Real3`: a 3d vector of reals
  * `RealN`: a vector of reals
  * `Real2N`: a list of 2d real vectors
  * `RealNM`: a list of real vectors
  * `Color`: a color
  * `String`: a string
  * `ColorType`: either `hsv` or `rgb`
  * `PathType`: either `open` or `closed`
  * `ShapeList`: a list of shapes
  * `PathCmd`: a path command
  * `Boolean`: a boolean value of true or false
* A shape type (type `ShapeT` in TypeScript) consists of types of all shapes (say, `Circle` and `Text`), and a special `AnyShape` type that takes in any shape objects.
* A union type (type `UnionT` in TypeScript) is just the union of two types.

For example, a parameter of type
```typescript
{
  tag: "UnionT",
  types: [
    { tag: "ValueT", type: "Real" },
    { tag: "ShapeT", type: "Rectangle" }
  ]
}
```
takes either a real number or a Rectangle shape.

The actual "type" that the parameter takes in is `ValueShapeT` which is defined as `ValueT | ShapeT | UnionT`.

## Function Parameters

A function parameter is defined as
```typescript
interface FuncParam {
  name: string;
  type: ValueShapeT;
  default?: Value<ad.Num>["contents"];
  description?: string;
}
```
It has a name, a type, an optional default value, and an optional description for documentation purposes.

## Functions

Objective and constraint functions are defined as follows:
```typescript
interface ObjFunc { // or ConstrFunc
  name: string;
  params: FuncParam[];
  body: ObjFuncBody; // or ConstrFuncBody
  description?: string;
}
```
where `ObjFuncBody` and `ConstrFuncBody` are both defined as `(...args: any[]) => ad.Num`. It has a function name, a list of parameters, a function body, and an optional description.

Computation functions (`CompFunc`) are defined similarly as follows:
```typescript
export interface CompFunc {
  name: string;
  params: FuncParam[];
  body: CompFuncBody;
  returns: ValueShapeT;
  description?: string;
}
```
where `CompFuncBody` is defined as `(context: Context, ...args: any[]) => Value<ad.Num>`. In addition to elements of a objective / constraint function, it also stores types of the return value.

# Errors

We have added four new errors in the Penrose system.
* `BadArgumentTypeError` is invoked when a parameter expects a type, but is provided with an argument of incompatible type. For example, assigning a real number to a parameter that takes in a shape, or assigning a 3d vector to a parameter that takes a 2d vector.
* `MissingArgumentError` is invoked when a function call is missing an argument for a parameter that does not have a default value.
* `TooManyArgumentsError` is invoked when a function call gives a function too many arguments.
* `FunctionInternalError` is invoked when the functions fails internally. For example, some functions take in two vectors and checks whether the two vectors have the same number of elements. If so, it continues on. If not, it throws an error. We catch these error and wrap the error message in `FunctionInternalError`.

### First Example 
We know that the `circumradius` function takes in three 2d vectors. So giving it 3d vectors, as follows,
```
x = circumradius([1, 2, 3], [4, 5, 6], [7, 8, 9])
```
will yield `BadArgumentTypeError` with message ``Parameter `p` (at line 7, column 22 of Style program) of function `circumradius` expects ℝ² but is given incompatible ListV value with 3 elements``. Notice that we use symbols that denote each type.

### Second Example
As another example, function `cubicCurveFromPoints` takes two arguments:
* `pathType` is either `open` or `closed`
* `pts` is a list of 2d points.
Suppose we violate the first argument and give `pathType` an arbitrary string.
```
x = cubicCurveFromPoints("hello", [(1, 2), (3, 4), (5, 6)])
```
This will yield error message ``Parameter `pathType` (at line 7, column 30 of Style program) of function `cubicCurveFromPoints` expects "open" | "closed" but is given incompatible StrV value with content "hello".``

Suppose we violate the second argument and give a list of 3d points.
```
x = cubicCurveFromPoints("closed", [(1, 2, 3), (3, 4, 5), (5, 6, 7), (6, 7, 8)])
```
The error message will be ``Parameter `pts` (at line 7, column 40 of Style program) of function `cubicCurveFromPoints` expects (ℝ²)ⁿ but is given incompatible LListV value of shape 4-by-3.``

### Third Example
The objective `isRegular` takes in a `Polygon`, `Polyline`, or `Path` shape. If we provided it with a Circle shape, 
```
    c = Circle {}
    encourage isRegular(c)
```
The error message will be ``Parameter `s` (at line 8, column 25 of Style program) of function `isRegular` expects PolylineShape | PolygonShape | PathShape but is given incompatible CircleShape.``

# Checklist

# Open ideas
* With this implemented we can also implement IDE-related and documentation-related features.
* We can also adapt the shape parameter type checking to use these new types.


packages/core/src/compiler/Style.ts
```diff
@@ -2254,27 +2258,29 @@ const evalExprs = (
...
-    argVals.map((arg) => {
-      switch (arg.tag) {
-        case "ShapeVal": // strip the `ShapeVal` tag
-          return arg.contents;
-        case "Val": // strip both `Val` and type annotation like `FloatV`
-          return arg.contents.contents;
-      }
-    })
-  );
+// const argValues = (
```
wodeni, line 2266: Is this still useful? If not, delete these lines.

packages/core/src/compiler/styleFunctionCaller/StyleFunctionCaller.ts
```diff
@@ -0,0 +1,101 @@
...
+  try {
+    return ok(func.body(...vals));
+  } catch (e) {
+    if (e instanceof Error) {
+      return err(functionInternalError(func, range, e.message));
+    } else {
+      throw new Error("Function call resulted in exception not of Error type");
+    }
+  }
+};
```
wodeni, line 77: I'd suggest de-duplicating this code a bit. Seems like only the context-passing bit is different. You may extract the rest of the function into `checkArgs` and make this much shorter.

packages/core/src/compiler/styleTypeChecker/styleTypeChecker.ts
```diff
@@ -0,0 +1,102 @@
...
+    return checkValueAgainstValueType(expected, given.contents);
+  }
+};
+
+export const checkValueAgainstValueType = (
+  expected: ValueType,
+  given: Value<ad.Num>
+): Value<ad.Num>["contents"] | undefined => {
+  const { tag, contents } = given;
+  if (
```
wodeni, line 57: Use `switch` to just fall through if there's no checking to do.

packages/core/src/compiler/styleTypeChecker/styleTypeChecker.ts
```diff
@@ -0,0 +1,102 @@
+import { Shape, ShapeType } from "../../shapes/Shapes";
```
wodeni, line 1: I think it's okay to just put this code into `Style.ts` or at least under of `compiler` folder. We don't really do much folder nesting in the repo and I want to keep the directory structure simple.

wodeni (changes requested):
Looked through code changes. Still working through the actual functions and constraints. Here are some comments on code style things:

packages/core/src/compiler/styleTypeChecker/styleTypeChecker.ts
liangyiliang: I want this function to return `undefined` if the check fails. The idea is, `checkValueAgainstValueType(type, value)` will return `value.contents` if `value` is consistent with `type`, and `undefined` otherwise. I guess there isn't really much difference in this case?

packages/core/src/compiler/Style.ts
liangyiliang: Sure will do this

packages/core/src/compiler/styleTypeChecker/styleTypeChecker.ts
wodeni: ahh I see the two-layer `if`s. Okay with keeping this design then because I don't see a switch improving the it more.

liangyiliang: Sounds good

packages/core/src/compiler/styleTypeChecker/styleTypeChecker.ts
liangyiliang: I'll put it under the `compiler` folder.

packages/core/src/compiler/styleFunctionCaller/StyleFunctionCaller.ts
liangyiliang: Sounds good, I'll see how I can refactor this.

liangyiliang: Done

samestep:
This isn't super important, but I did want to note that this current approach introduces the possibility of the function `name` not matching its key in the dictionary where it resides. If you want, you can fix this by using some [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) magic combined with JavaScript destructuring syntax:

```typescript
const makeConstrFunc = (
  description: string,
  params: FuncParam[],
  body: ConstrFuncBody
): { [name: string]: ConstrFunc } =>
  new Proxy(
    {},
    {
      get: (_, name: string): ConstrFunc => ({
        name,
        params,
        body,
        description,
      }),
    }
  );

const { equal } = makeConstrFunc(
  "Require that the value `x` is equal to the value `y`",
  [
    { name: "x", description: "First value", type: realT() },
    { name: "y", description: "Second value", type: realT() },
  ],
  (x: ad.Num, y: ad.Num) => {
    return absVal(sub(x, y));
  }
);

const constrDictSimple: { [k: string]: ConstrFunc } = {
  equal,

  // ...
};
```

samestep:
Alternatively, we could skip the magic and just write a test that iterates through all the entries in our objective/constraint/function dictionaries and checks that the key is equal to the `name`. (Or maybe you've already written such a test? I haven't read this diff yet...)

liangyiliang:
> This isn't super important, but I did want to note that this current approach introduces the possibility of the function `name` not matching its key in the dictionary where it resides. If you want, you can fix this by using some [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) magic combined with JavaScript destructuring syntax

Thanks! Yes, I cannot be sure that I haven't made some mistake in transcribing the functions. I'll look into how Proxies works.

liangyiliang:
> Alternatively, we could skip the magic and just write a test that iterates through all the entries in our objective/constraint/function dictionaries and checks that the key is equal to the `name`. (Or maybe you've already written such a test? I haven't read this diff yet...)

Nope, I don't have such a test written yet. In fact I just realized that I haven't written test cases for errors that I added. I will add these test cases.

liangyiliang:
> Alternatively, we could skip the magic and just write a test that iterates through all the entries in our objective/constraint/function dictionaries and checks that the key is equal to the `name`. (Or maybe you've already written such a test? I haven't read this diff yet...)

Nope, I don't have such a test written yet. In fact I just realized that I haven't written test cases for errors that I added. I will add these test cases.

packages/core/src/compiler/styleTypeChecker/styleTypeChecker.ts
liangyiliang: Done

wodeni (approved):

packages/core/src/compiler/Style.ts
```diff
@@ -3116,6 +3113,15 @@ const translateExpr = (
...
       }
+      const argsWithSourceLoc: ArgValWithSourceLoc<ad.Num>[] = [];
+      for (let i = 0; i < args.value.length; i++) {
+        const { start, end } = argExprs[i];
+        argsWithSourceLoc.push({
+          ...args.value[i],
+          start,
+          end,
+        });
+      }
```
samestep, line 3124: I see this in a few places; would it make sense to extract into a helper function?

packages/core/src/contrib/Constraints.ts
```diff
@@ -28,246 +31,454 @@ import {
 } from "./ConstraintsUtils";
 import { constrDictCurves } from "./CurveConstraints";
 import { bboxFromShape, shapeDistance, shapeSize } from "./Queries";
-import { inRange, isLinelike, isRectlike, overlap1D } from "./Utils";
+import { inRange, isRectlike, overlap1D } from "./Utils";
 
 // -------- Simple constraints
 // Do not require shape queries, operate directly with `ad.Num` parameters.
-const constrDictSimple = {
+const constrDictSimple: { [k: string]: ConstrFunc } = {
```
samestep, line 38: This seems wrong; doesn't this erase all the individual function types when you try to access them from the dictionary? Is this why you need to do a bunch of casting in the tests now? I'd prefer to leave this off so that the individual functions each get their own precise types, and then use something like the `constrDictVals` we had before to make sure they each fit into the `ConstrFunc` type.

packages/core/src/contrib/Functions.ts
```diff
@@ -2026,10 +3042,10 @@ const closestPointEllipseCoords = (
...
 // `_compDictVals` causes TypeScript to enforce that every function in
 // `compDict` takes a `Context` as its first parameter and returns a `Value`
-const _compDictVals: ((
-  context: Context,
-  ...rest: never[]
-) => Value<ad.Num>)[] = Object.values(compDict);
+// const _compDictVals: ((
+//   context: Context,
+//   ...rest: never[]
+// ) => Value<ad.Num>)[] = Object.values(compDict);
```
samestep, line 3048: I'd delete this code instead of commenting it out, but I actually think you should uncomment it instead (see the comment I left above about precise types).

samestep (commented):
Looks great overall, thanks for doing this! Had one minor comment and one medium-priority comment.

packages/core/src/contrib/Constraints.ts
liangyiliang: Yep I can do that

liangyiliang: But then `constrDict[someFunc]` has type `any` ...?

packages/core/src/contrib/Constraints.ts
```diff
@@ -278,8 +489,4 @@ export const constrDict = {
   ...constrDictCurves, // Curve-specific constraints.
 };
 
-// `_constrDictVals` causes TypeScript to enforce that every function in
-// `constrDict` returns an `ad.Num`
-const _constrDictVals: ((...rest: never[]) => ad.Num)[] = Object.values(
-  constrDict
-);
+const _constrDictVals: ConstrFunc[] = Object.values(constrDict);
```
samestep, line 492: Can you add back a comment similar to the one that was here before?

packages/core/src/contrib/Functions.ts
```diff
@@ -1543,254 +2401,414 @@ export const compDict = {
...
+      points: [ad.Num, ad.Num][],
+      closed: boolean
+    ): FloatV<ad.Num> => {
+      return { tag: "FloatV", contents: totalCurvature(points, closed) };
+    },
+    returns: valueT("Real"),
   },
 };
 
+const _compDictVals: CompFunc[] = Object.values(compDict);
```
samestep, line 2810: Same here

packages/core/src/contrib/Constraints.ts
liangyiliang: yes

Issue #776, opened 2022-01-07
Implement Style modules (imports and namespaces)


keenancrane:
A feature we had targeted for Style 1.1, but didn't make it into the release, was wrapping a Style "module" in a namespace, and providing the ability to import several modules into the same Style program.

The design for this feature is detailed in [this design doc](https://docs.google.com/document/d/1mBSs0IvJpwo29iVAlw0EHW_QU_PLWczmL_GzZB-zDJw/edit#heading=h.4qncuow56vv0).

PR #1354, opened 2023-05-01
refactor: restore the 2D triangle mesh domain

wodeni:
# Description

#770 and #808 introduced the 2D triangle mesh domain but never tracked them in the registry. This PR restores this domain by fixing a few minor issues and add diagrams to the registry. 

* `attrRotation` uses the top-left corner as the default rotation origin, which seems wrong. Changed so that the default is the shape's bbox center. Due to this fix, all diagrams that contain `Text` are impacted, since we include the `rotation` transformation even if the rotation is zero.
* The font size settings in the original Style program seems a little off from the targeted diagrams. Not really sure if anything changed between #770 and now, as the system changed quite a bit, but I updated some stylings so they are a little closer.
* Added mesh disjoint styling for `angle-equivalence`. The code remains commented out because it's blocked by #776, but here's a diagram with the block uncommented.
image

# Checklist

# Open questions


samestep (approved):
Looks great, thanks so much for doing this! A lot of the diagrams look really nice :smiley:

PR #1355, opened 2023-05-02
refactor: put L-BFGS into its own module

samestep:
# Description

This PR refactors our implementation of [L-BFGS](https://en.wikipedia.org/wiki/Limited-memory_BFGS) into its own Rust module, while also refactoring the code to be cleaner, and adding comments indicating textbooks from which the implementations are drawn.

This new implementation also happens to get rid of all memory allocation in the L-BFGS loop (at least, once the list of vector pairs has reached its maximum length); I was hoping that this would give a performance increase, but it does not seem to. Still, the resulting code is nice and clean, in my opinion.

# Checklist


wodeni:
Looks like some of the registry diagrams changed quite a bit. Was there a change of optimizer settings? e.g. https://github.com/penrose/penrose/blob/6a68a7fa227c383fc2f1e1e9ce94cb501939654d/lagrange-bases/lagrange-bases.svg

packages/optimizer/src/lib.rs
```diff
@@ -656,97 +309,116 @@ fn minimize(
...
-    let mut i = 0;
-    let mut t = 0.0001; // NOTE: This const setting will not necessarily work well for a given opt problem.
-    let mut failed = false;
 
-    let mut new_lbfgs_info = lbfgs_info;
+    let cfg = lbfgs::Config {
+        m: 17,
+        armijo: 0.001,
+        wolfe: 0.9,
+        min_interval: 1e-9,
```
wodeni, line 322: This was changed from `1e-10` to `1e-9`, for example. Why?

samestep: Previously it was `10e-10`, which is the same as `1e-9`.

PR #1357, opened 2023-05-04
feat: line wrap in error pane

### LINKED ISSUE #1356: Line-wrapping in error panes
Here is an example error message.
image

The message is too long that we need to scroll horizontally.

We should have line-wrapping.

liangyiliang:
# Description

Resolves #1356.

Now the error pane looks like this:
image


samestep (approved):
Looks good to me 👍 @wodeni?

wodeni (approved):
LGTM!

PR #1361, opened 2023-05-08
feat: abstract shape parameters

liangyiliang:
# Description

For many shape-related Style functions, we expose their lower-level equivalents (that take _parameters_ of shapes, not actual shape objects) to the Style writers.

# Motivation
## Concrete Shape Functions
Penrose reasons about shapes via Style functions. Function call `shapeDistance()` takes in two shape objects and returns the Minkowski distance between the two shapes. These functions are invoked by providing the actual shape objects as arguments to the function, like `shapeDistance(rectangle, circle)`. This invocation requires that the shapes `circle` and `rectangle` are previously defined, like
```
circle = Circle {
  ...
}
rectangle = Rectangle {
  ...
}
```
which necessarily renders the two shapes onto the canvas. As a result of their rendering, these shapes are _concrete_.

## Abstract Shape Functions

But sometimes, we would like to reason about _abstract_ shapes, without the shapes being rendered on the canvas. Exposing lower-level functions that take in shape parameters instead of actual shape objects allow us to do these. Consider the abstract variant of the `shapeDistance` function on `Circle` shape and `Rectangle` shape:
```
shapeDistanceRectCircle(
  four_points_of_rect,
  circle_center,
  circle_radius,
)
```
Here, the function no longer takes in shape objects, but their underlying values: a rectangle shape can be determined by its four points (top-right, top-left, bottom-left, bottom-right) which constitute a 4x2 matrix, and a circle shape is determined by its center (a vector) and a radius (a scalar), so the abstract variant takes in the matrix of rectangle points, the circle center vector, and the circle radius scalar.

## Abstract Functions Is Almost More Powerful

These abstract versions can do **almost** everything that the concrete versions can do. For example, the concrete function call `shapeDistance(rectangle, circle)` is conceptually equivalent to the abstract version `shapeDistanceRectCircle(bboxPts(rectangle), circle.center, circle.r)`, where `bboxPts` takes in a concrete shape and returns the four points (top-right, top-left, bottom-left, bottom-right) (in that order) of its bounding box.

The abstract versions of the functions can be more powerful. The abstract parameters don't need to be related to any concrete shapes - in fact, we can provide _any_ number (or combinations of numbers) as long as the types match according to the functions' signatures.

## Almost ...

We emphasize the word **almost** because there are some things that cannot be done with the abstract versions. The concrete functions can have branches that trigger depending on the types of the shape given. This gives us some polymorphism which can come in handy if we want to write generic code:
```
-- dsl
type Shape
type Circle <: Shape
type Line <: Shape
type Rectangle <: Shape

-- sty
forall Circle C {
   C.icon = Circle{}
}
forall Line L {
   L.icon = Line{}
}
forall Rectangle R {
   R.icon = Rectangle{}
}
forall Shape S1; Shape S2 {
   ensure shapeDistance(S1.icon, S2.icon) > 50
}
```
The concrete function `shapeDistance` works for all shape types that `S1.icon` and `S2.icon` have. Using only abstract functions we wouldn't be able to write these, since different variants of the abstract `shapeDistance` take in different parameters whose types are incompatible.

## Goal

Since the concrete shape functions still play an important role, the goal is an exposure of the underlying functions, not the removal of any present functions (except some badly-behaving ones which we will talk about below). That is, we keep `shapeDistance` that take in two actual shapes, while adding smaller functions like `shapeDistanceRectCircle`.

Another goal is to `export` as many library functions as possible to outside of the function dictionary. This is mostly to benefit future Penrose API development.

# Design Details

## Functions Refactored Into Abstract Branches

Here is a list of functions whose subbranches we refactored into abstract functions.

* `rayIntersect`
* `rayIntersectNormal`
* `closestPoint`
* `closestSilhouettePoint`
* `shapeDistance` (we add `shapeDistanceLines` that computes the Minkowski distance between two `Line` shapes)

Here is a list of constraints whose subbranches we refactored into abstract functions.

* `overlapping`
* (`disjoint` and `touching` are just wrappers of `overlapping` so we don't do anything about them)
* `contains`

No objective functions were changed because no objective functions actually take in concrete shapes.

## Functions Added

We add in the function `bboxPts` that returns the top-right, top-left, bottom-left, and bottom-right points (in that order) of the bounding box of a shape. This serves two purposes.

* Since the abstract functions represent rectangular shapes as its four points, `bboxPts` allows us to extract the four points of a rectangular shape.
* The points are ordered in a way that makes a polygon. These points can be assigned to the `points` field of a `Polygon` or `Polyline` shape.

We also allow the direct extraction of points from rectangle-like shapes using function `rectPts`. This function skips the computation of bounding boxes and takes into account the `rotation` field, making it less expensive.

Finally, we expose the `inDirection` function in its abstract form.

## Functions Removed

We removed function `shapeSize` and its related functions `smallerThan` and `largerThan`. These functions are very fragile and can break easily. Through discussions, we concluded that it is better to have the user specify an exact measure of shape size.

If the user wants to ensure that a circle `a` is larger than another circle `b`, the user can just write `ensure a.r > b.r` or similar expressions with more fine-grained control.

# Checklist


packages/docs-site/docs/ref/constraints.md
```diff
@@ -128,43 +128,6 @@ Previously, we've talked about how we convert everything to zero-based inequalit
 
 One common operation is to access the parameter of a shape via `shapeName.propertyName.contents`, which will return a `ad.Num`. For example, if you have a circle `c` as input, and you want its radius, `c.r.contents` will give you something like `5.0` (of type `ad.Num`).
 
-## Constraints Example: minSize & maxSize
```
wodeni, line 131: Thanks for updating the docs too! In this case it's probably better to change this example to another one. Can you open an issue so we remember to do this later?

wodeni (changes requested):
Looks great in general! Thanks so much for the hard work on this. Exposing lower-level functions would be super useful for the API for sure. My comments are mostly about documentation.

packages/core/src/contrib/Functions.ts
```diff
@@ -272,9 +289,9 @@ export const compDict = {
     description:
       "Return a paint color of elements `h`, `s`, `v`, `a` (hue, saturation, value, opacity).",
     params: [
-      { name: "h", description: "Hue in [0, 360)", type: realT() },
-      { name: "s", description: "Saturation in [0, 100]", type: realT() },
-      { name: "v", description: "Value in [0, 100]", type: realT() },
+      { name: "h", description: "Hue", type: unitT() },
```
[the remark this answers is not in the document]
liangyiliang: Apparently,
* Originally, `h, s, v` were all `unitT()` which is incorrect. This incorrect documentation was in the main branch which made it into the `absshapes` branch.
* @joshsunshine 's timeline PR fixed the documentations to use `realT()`.
* This change was accidentally ignored when I merged `main` into the `absshapes` branch.

Fixed.

packages/core/src/contrib/Functions.ts
```diff
@@ -1836,7 +1852,7 @@ export const compDict = {
    */
   unitRandom: {
     name: "unitRandom",
-    description: "Uniformly sample a random value in the range [0,1).",
+    description: "Uniformly sample a random value in the range [0,1].",
```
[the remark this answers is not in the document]
liangyiliang: Yep this was copied from an older version. Fixed.

packages/docs-site/docs/ref/constraints.md
liangyiliang: This whole page was outdated since it still used the older shape-tuples (`[shapeType, props]`) which doesn't exist anymore since the shape-type refactoring. Since one of our goals is to make it easier for outsiders to contribute library functions (I think), I will write up an issue to update this page.

liangyiliang: See #1488

wodeni: Looks good. Thanks!

wodeni (approved):
Looks great. Thanks for addressing the comments!

keenancrane:
@liangyiliang **Question:** why are rectangles specified abstractly as four points, rather than bottom left/top right corners, or center/width/height?  It would seem that specifying them as four points could cause some unusual/unexpected behavior if these points do not actually describe a rectangle.

I suspect this choice was made to facilitate the internal implementation.  For the external interface, perhaps we can provide convenience functions that allow one to specify either (i) lower-left/upper-right corner and (ii) center/width/height, depending on what data is available.

keenancrane:
This PR caused regressions in the test for geometric queries.  For instance, we get these random-looking lines for ray intersections that go outside the individual boxes:

image

@liangyiliang can you take a look and try to figure out what broke here?

PR #1369, opened 2023-05-09
feat: timeline

### LINKED ISSUE #1263: Style program for Timelines
I think we should build `Domain` and `Style` programs for timelines. These diagrams appear in almost all grant proposals, which allows us to make further inroads into our core academic audience. In addition, these proposals are common in business presentations, white papers, etc. 
Writing a Substance should be as simple as specifying tasks, giving tasks tags, and specifying the begin and end time of each task. 

An example timeline from one of our recent proposals looks like this:
PenroseSloanTimeline

I think this could be specified in `Substance` as so:
```
TODO: Will fill in later.
```

TODO: Compare to [Mermaid Gantt diagrams](https://mermaid.js.org/syntax/gantt.html)

joshsunshine:
### To-do list
- [x] Add diagram to registry
- [ ] Describe user-parameters in Style program

# Description
This pull request resolves #1263. It introduces a new domain and style programs for creating timeline diagrams like those found in grant proposals and thesis proposals. An example of the output is shown here:
Example timeline diagram.

The domain program has four Types: `Task`, `Category`, `Year`, and `Quarter`.  Tasks are the main element of the timeline and are shown visually in the examples as rounded rectangles. Each task is labeled. For example, the top left task is labeled "Web IDE." Each tasks has a `Category` and all events in the same category are displayed with the same color. For example, the outcome category is shown in purple in the diagram above while the accessibility category is shown in green. The`Year` and `Quarter` types define the beginning and end periods for each task. There are three years (2023, 2024, and 2025) defined for the diagram above. Quarters belong to years. There are 8 quarters (Q1-Q4 in 2023 and 2024, and Q1&Q2 for 2025). Gridlines show where each quarter begins and ends and the labels at the top show which belongs to each quarter. 

The remainder of the Domain program allows users to specify the relationships:
```
constructor MkTask (Quarter start, Quarter end) -> Task
predicate In(Task t, Category c)
constructor MkQuarter(Year y) -> Quarter
```
`MkTask` specifies relationships between tasks and time periods. For example, `MkTask(q1_2023, q2_2023)` creates a task between the first quarter and second quarters of 2023 like "Web IDE" above. `In` specifies relationships between tasks and categories.`MkQuarter` specifies relationship between quarters and years. 

This style of diagram is relatively concrete so most shape properties are computed, not optimized. However, optimization is used to ensure that the tasks are disjoint and the events in each category are grouped vertically. 

**Limitations:**
- The diagram looks best when the height of the canvas is tweaked manually based on the number of tasks displayed. Similarly, the width of the canvas should be tweaked to based on the number of quarters and the amount of text labeling each event. We should consider allowing Style programs to optimize or compute the canvas size.
- The `Before`, `First`, and `Last` predicates are necessary because Style does not support aggregation. They should be removed once #1378 is resolved. The predicates enable the Style program to do the following without aggregation:
  - automatically center the year label above its component quarters
  - correctly position gridlines around each year
  - layer the gridlines below the events
- All diagrams created with this style must have both years and quarters. It would be nice to generalize the style program so it can support more generic time periods (e.g. days instead of quarters) and time period hierarchies (e.g. days in months in years). 
- In the original diagram that inspired this PR, events were represented by pointed ribbons (wide hexagons) with gradient color fills. This style is visually simpler.
- In the original diagram that inspired this PR, there was a "And more ..." time bucket that spanned two columns in the header. This style does not support that time of heading.
- Style establishes vertical grouping of events by category and category coloring by specific identifier matching (e.g. `` Category  `outcome` `` not `Category c`). The style program is therefore not robust to different categories. This limitation should be fixable once #1378 is resolved.
- There is a lot of cruft in the Substance program. The program for the diagram above is 95 lines long. 15 lines of that will be cut once #1378 is resolved but some of the cruft is fundamental to our design of the Substance language.


wodeni:
Made some updates to the Style program:
* Use `collect` to resolve the extents of grid lines
* Remove `Before(Quarter, Quarter)` after refactoring layering to work without requiring them
* Add `Before(Category, Category)` so the `sort` layout stage is not specific to one Substance program
* Compute visually distinct and pleasing colors by incrementing hue values by `match_id`. This also makes the Style program more generic.

wodeni (approved):
LGTM! While working on the Style program I also found that the docs for `hsva` are wrong. Fixed it in this PR too.

PR #1374, opened 2023-05-10
feat: `appendShape` function

### LINKED ISSUE #1372: `appendShape` function

If we have a `Group` shape, it is currently not possible to add a shape into the group after the creation of the group.

We should have a function (with side-effects) called `appendShapes`, used as follows:

```
appendShape(shapeList, shape)
```

and that would imperatively add the shape into `shapeList`.

Another solution we considered is

```
shapeList = add(shapeList, shape)
```

but that would cause a cyclic dependency graph issue.

liangyiliang:
# Description

Resolves #1372 .

We implement a function called `appendShape` that takes a list of shapes, and appends a shape into the list of shapes imperatively (in-place).

Given that there now exists Style computation functions that have side-effects, we add another dictionary full of imperative computational functions, `impDict`.

We also allow top-level function calls for functions in `impDict`, since, many times, we only want the imperative behavior. If a function is not in `impDict`, we do not allow top-level function call.

https://appendshape.penrose-72l.pages.dev/try/?gist=2f26b6d92c64332ca92a80afcb40d54b

# Checklist

### DIFF
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -2884,7 +2884,9 @@ const evalExpr = (
           oneErr({ tag: "InvalidFunctionNameError", givenName: name })
         );
       }
-      const f = compDict[name.value] ? compDict[name.value] : impDict[name.value];
+      const f = compDict[name.value]
+        ? compDict[name.value]
+        : impDict[name.value];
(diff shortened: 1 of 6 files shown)


wodeni (approved):

liangyiliang:
This PR has a pretty fatal issue with dependency graph which I don't know how to solve. Leaving it for now ...

PR #1391, opened 2023-05-15
feat: export `core` API functions for optimization specification and solving

wodeni:
# Description

Following up on #1368, #1366, and #1418, this PR creates an `api.ts` file in `@penrose/core` to hold types and functions meant for programmatic and interactive usage of Penrose core, including a new API for optimization which separates the concept of a `Problem` (with an `objective` and some `constraints`) from that of a `Run` (in which you can `freeze` some `vals`). 

This API provides a nicer abstraction over `@penrose/optimizer`, which we run in a loop until the `until` callback returns `true`, instead of possibly returning early before convergence or `until`. We also rename `ad.Input` to `ad.Var`, and the `input` function to `variable`, and we change all the output types in `AutodiffFunctions` from more precise subtypes to just `ad.Bool` and `ad.Num`.

In addition to the optimization API, we also provide utilities for building interactive diagrams with optimization. The main concern is the interplay among the computation graph, optimization, and reactivity. We chose SolidJS to demonstrate how our API can work with a reactive framework. SolidJS is chosen mainly because it provides fine-grained control over reactivity and is syntactically nicer to work with. In our examples, we wrap [mutables](https://www.solidjs.com/docs/latest#createmutable), [stores](https://www.solidjs.com/docs/latest#stores), and [signals](https://www.solidjs.com/docs/latest#createsignal) around computation graph values (both inputs and intermediate nodes) to achieve reactivity. Optimization is handled using [effects](https://www.solidjs.com/docs/latest#createeffect). 

To demonstrate this new API, we add a couple new interactive examples in `@penrose/solids`:

- rotating 3D triangles with real-time label optimization and dynamic layering

https://github.com/penrose/penrose/assets/11740102/1976ff77-0776-4734-8c50-39a836813e47

- a recreation of https://setosa.io/ev/eigenvectors-and-eigenvalues/

https://github.com/penrose/penrose/assets/11740102/88c74bef-ff72-4e93-8672-aa0f5e83c5d5

- an implementation of @jiriminarcik's elastic curve demo with configurable number of points

Recording 2023-05-18 at 23 53 14

We import the rotating-triangles and eigenvalues examples into `@penrose/examples` to check it in the registry.

While building these examples, we developed some more general utilities for using Penrose with SolidJS, which we've gathered into `packages/solids/src/util.ts`. As the API stabilizes a bit more, we will put these more general utilities into a `@penrose/solid` package.

```sh
npx nx run solids:dev
```

# Checklist


wodeni:
Added demo videos in the PR description and more description about reactivity. Merging now!

PR #1396, opened 2023-05-16
fix: passthrough auto fill

liangyiliang:
# Description

In #1337, while adapting the passthrough auto-fill feature to work with the consolidated shape types, I introduced a bug where the `style` would not correctly map.

This PR fixes it by reverting the logic of the `attrAutoFilSvg` to the version before #1337.


samestep (approved):
Awesome, thanks!

PR #1388, opened 2023-05-15
feat: Blobs

jiriminarcik:
# Description

More curve examples - see related PRs: #1384, #1386

1. New example in the registry:
- `Blobs`: blobs representing manifolds and arrows between them representing maps (see figure [Blobs](https://raw.githubusercontent.com/penrose/penrose/8ef4e28dceee792a69188cde9dc20e2f7c9ea696/blobs-blobs.svg))

2. Added new functions to enable the blobs and arrow between them:
- `inflectionEnergy`: integral of derivate of curvature raised to a given power 
- `centerOfMass`: center of mass for a point clouds

3. Generalised `consecutiveTuples` and `consecutiveTriples` in `core/src/contrub/Utils.ts`.

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have reviewed any generated registry diagram changes

# Open questions

The style file uses many hacks, mostly created by a python script that prints out list of variables, shapes etc.

PR #1386, opened 2023-05-12
feat: Another curved graph style

jiriminarcik:
# Description

Use elastic energy from #1384 to create another curved graph style for `graph-domain`.
- Added 2 new styles (all in example folder `graph-domain`):

  - `simple-curved-graph-dots.style`: The labels are in the middle of the nodes, instead of next to them.

  - `simple-curved-graph-boxes.style`: Nodes are boxes instead of circles to better accommodate large text labels.

- Added 5 new graph substance files to demonstrate the new styles (all in example folder `graph-domain`):
I see, in that case the revised descriptions would be:

  - `other-examples/arpanet.substance`: This file contains a representation of the early ARPANET with more nodes that were added later in its development. The vertices represent institutions such as UCLA, SRI, UCSB, Utah, BBN, MIT, RAND, SDC, Harvard, LL, CWRU, CMU, and NASA. Edges represent direct network links between these institutions, demonstrating the expanding connectivity of this precursor to the modern internet.

  - `other-examples/evolutionary-tree.substance`: This file represents a simplified model of the tree of life, with vertices representing major domains of life (Life, Eucaryote, Bacteria) and edges indicating evolutionary relationships. The structure further branches into subdomains like Animals, Plants, Fungi under Eucaryotes and Proteobacteria, Cyanobacteria under Bacteria, illustrating the hierarchical nature of evolutionary relationships.

  - `other-examples/glycolysis-pathway.substance`: This file represents a simplified version of the glycolysis metabolic pathway, where glucose is converted into pyruvate. Each vertex represents a different metabolite, and each edge represents a biochemical reaction where one metabolite is converted into another. The metabolites are represented using abbreviated names for simplicity.

  - `other-examples/krebs-cycle.substance`: This file represents a simplified version of the citric acid (Krebs) cycle, a key metabolic pathway that provides high-energy molecules through the oxidation of acetyl-CoA derived from carbohydrates, fats, and proteins. The vertices represent different molecules involved in the cycle, and edges represent the biochemical reactions transforming one molecule into the next.

  - `other-examples/nyc-subway.substance`: This file contains a simplified representation of the New York Subway System. The vertices represent specific subway stations such as Times Square, Grand Central, Wall Street, etc., and the edges represent direct subway routes connecting these stations. This graph provides an abstracted view of the transit connections in New York City.

- Added 2 new examples to `registry.json`:
[...]

PR #1384, opened 2023-05-11
feat: Improve elastic energy

jiriminarcik:
# Description

- Added new curvature approximation modes:
```
enum CurvatureApproximationMode {
  FiniteDifferences = "FiniteDifferences",
  Angle = "Angle",
  SteinerLineSegment = "SteinerLineSegment",
  SteinerCorner = "SteinerCorner",
  OsculatingCircle = "OsculatingCircle",
}
```
- Changed the curvature approximation used in the `elasticEnergy` function
- Added simple examples illustrating the usage of simple curve energies for open and closed curves
- Exposed `isEquilateral` and `isEquiangular` in objectives
- Added `signed` parameter to `totalCurvature`, the default is still signed
- Added new functions:
  - `maxCurvature`: maximum of curvature along open/closed curve
  - `pElasticEnergy`: generalisation of elastic energy
  - `lengthK`: generalised length (encourages equidistant discretisation)
- Added 3 new examples to the registry:
  - Open elastic curve example (only for demonstration, not a useful diagram)
  - Closed elastic curve example (only for demonstration, not a useful diagram)
  - Curved graph example

# Examples with steps to reproduce them

- [Curved graph example](https://raw.githubusercontent.com/penrose/penrose/0bfcbc370afad5169f35ddb3755cd72d42cb9838/sec5-ex46-simple-curved-graph.svg)
- [Open elastic curve example](https://raw.githubusercontent.com/penrose/penrose/0bfcbc370afad5169f35ddb3755cd72d42cb9838/cubic-bezier-open-elastic-curve.svg)
- [Closed elastic curve example](https://raw.githubusercontent.com/penrose/penrose/0bfcbc370afad5169f35ddb3755cd72d42cb9838/cubic-bezier-closed-elastic-curve.svg)

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have reviewed any generated registry diagram changes

# Observations

- Repeating stage sequences in layering (e.g. `layout = [dots, text, adjust, dots, text, adjust, text, adjust]`) can be helpful to avoid local minima. The optimiser can maybe do this automatically when it detects being stuck.
- `isEquilateral` objective or constraint is needed to ensure curve stability. In some cases `lengthK` can be used instead.
[...]

PR #1399, opened 2023-05-17
feat: space curves

jiriminarcik:
# Description

This PR generalises constraints and objectives for curves from #1388, #1386 and #1384 to work for space curves.
The new constraints and objectives are demostrated in a new [example](https://raw.githubusercontent.com/penrose/penrose/b61142075e64512b348e6f51b2aa2738d613578d/curve-examples/space-curves.svg) that has been added to the registry.

# Checklist

# Open questions

- Can we do surfaces?


packages/core/src/contrib/CurveConstraints.ts
```diff
@@ -74,7 +67,8 @@ export const curvature = (
   //     DOI: 10.1090/noti1578
   const v1 = ops.vsub(p2, p1);
   const v2 = ops.vsub(p3, p2);
-  const angle = ops.angleFrom(v1, v2);
+  const angle =
+    p1.length === 2 ? ops.angleFrom(v1, v2) : ops.angleBetween(v1, v2);
```
samestep, line 71: Why don't you just always use `angleBetween` here?
```suggestion
  const angle = ops.angleBetween(v1, v2);
```

samestep (commented):
Amazing! I just had one question, other than that everything looks great.

packages/core/src/contrib/CurveConstraints.ts
jiriminarcik: The signed curvature is useful for computing the turning number of planar curves. I added comments to clarify the behaviour of this functions.

samestep (approved):

Issue #1020, opened 2022-05-25
Provide explicit random sampling in Style


keenancrane:
It would be really helpful in Style to be able to say something like

```
scalar x.val = random( minVal, maxVal )
```

and get a random value that stays fixed throughout optimization.  This value would change only if the `variation` changes.  In fact, we already get this behavior if we write

```
scalar x.val = ?
```

and `x.val` is not used in any constraints/objectives: it gets set to a random value that is unaffected by optimization, and does not change unless the `variation` changes.  Basically we’re just lacking the ability to sample values from a user-specified range rather than whatever heuristic the sampler uses by default.

#### Bonus points

In a graphical context, it's often useful to have other random samplers.  For instance, a method

`vec2 p = uniformRandomDisk()`

might be provided to sample a point in the unit disk D² := { x ∈ ℝ² : |x| ≤ 1 }.  This method can be implemented via the formula

> √u ( cos(v), sin(v) )

where `u` and `v` are both uniform random values from [0,1].  Note that you have to be careful when implementing these formulas—it's easy to write something that sounds good, but gives the wrong distribution.  E.g., for the disk it's tempting to just use the first number as the radius (without the square root) and the second number for the angle:

![image](https://user-images.githubusercontent.com/3604525/170298765-96885472-0b8d-4d6e-b044-44d017c83052.png)

Other formulas, as well as a recipe for deriving uniform sampling formulas for other shapes, can be found in the article,

Jim Arvo, "[Stratified Sampling of 2-Manifolds](https://cseweb.ucsd.edu/~viscomp/classes/cse274/wi18/readings/course29sig01.pdf)", SIGGRAPH Course Notes (2001)

samestep:
#1199 refactors our `InputMeta` type to support this; then all that would remain is to expose whatever particular `random` sampler functions we want in `packages/core/src/contrib/Functions.ts`.

keenancrane:
Thanks to the work of @wodeni and others on #1199, this issue has now been addressed in #1404.

PR #1404, opened 2023-05-17
feat: Expose random sampling in Style

keenancrane:
# Description

image

This PR exposes generic random sampling in Style, adding several sampling functions to the function library that commonly arise in diagrams.  The ability to set values to random constants is especially helpful for simplifying programs that would otherwise need to enforce containment in a set via constraints.  The point here is not to reduce problem size, but rather to reduce "fighting" between constraints/objectives, and in turn minimize the amount of problem-tweaking needed to make a good-looking diagram.

A new example in `examples/src/random-sampling` provides tests for the sampling method.  This example looks nice and steers users toward potentially useful functionality, so it is also now included in the gallery.  `CONTRIBUTING.md` has also been updated to explain that tests should be added to the set of examples (as done here).

The PR also adds a `Random Sampling` item to the (Vite) Style documentation, explaining that randomly-sampled values depend on the `variation`, and otherwise basically behave like constants (e.g., they cannot be optimized or changed, just like any other numerical constant).  Hopefully this helps avoid common misunderstandings of how sampling works / what it means.

The PR introduces the following functions:

- `random( minVal, maxVal )` — samples a random constant uniformly from the range `[minVal,maxVal]`
- `unitRandom()` — samples a random constant uniformly from the range `[0,1]`.
- `normalRandom()` — samples a random constant from a normal distribution with mean 0 and standard deviation 1.
- `diskRandom()` — samples a random point uniformly from the unit disk.
- `sphereRandom()` — samples a random point uniformly from the unit sphere.
- `triangleRandom(a,b,c)` — samples a random point uniformly from a triangle with vertices `a`, `b`, `c`.

# Checklist

# Open questions

There may be other common samplers not implemented here, but seems fine to add them in a future PR.

Another very reasonable thing to do in the future is to create samplers associated with shapes, i.e., provide a method like `uniformSample( S )` where `S` has any of our shape types (`Circle`, `Rectangle`, `Ellipse`, `Polygon`, etc.).  This then becomes one of our basic geometric queries, similar to (signed) distances and closest points.


samestep (changes requested):
Looks great, thanks for implementing and documenting these! Could you add some tests for them so they don't break in the future? I see that you have a teaser diagram in your PR description; if that was generated via Penrose then you could just add that trio to the registry.

samestep:
Also, could you be more precise about inclusive/exclusive ranges in your documentation? The `uniform` sampler is inclusive of the lower bound and exclusive of the upper bound; it would be good to be accurate about that in your `random` and `unitRandom` function documentation, for instance.

keenancrane:
> The `uniform` sampler is inclusive of the lower bound and exclusive of the upper bound; it would be good to be accurate about that in your `random` and `unitRandom` function documentation, for instance.

I guess you're referring to [this line](https://github.com/davidbau/seedrandom/blob/released/seedrandom.js#LL72C1-L72C1)?

> Also, could you be more precise about inclusive/exclusive ranges in your documentation?

I hesitate to say this in the docs, because it may lead diagrammers to worry about things they shouldn't (e.g., _"ok, so then how do I get values in [0,1] rather than [0,1)_? Now I'm worried!").

In the continuous setting, there is no difference between saying "draw a sample uniformly at random from [0,1]" and "draw a sample uniformly at random from "[0,1]" (unless you really want to split hairs and consider [nonstandard assumptions](https://stats.stackexchange.com/questions/140805/in-the-definition-of-probability-density-function-does-it-matter-if-the-interva)).

In the implementation, there is indeed a formal difference in the sense that (due to the line cited above) the value 1.0 will never be sampled by the PRNG.  But also note the comment

```
// This function returns a random double in [0, 1) that contains
// randomness in every bit of the mantissa of the IEEE 754 value.
```

Such a value has 24 bits of mantissa, meaning that the interval of the real line closer to the value `1.0` than any other floating-point value has size (approximately...) 1/2^24 ≈ 0.000000059604645.  Especially in the diagramming context, this is not something that is going to have any discernible impact.  (Even for more demanding applications like numerical integration and cryptography, it would be quite hard to detect.)

wodeni (changes requested):
Looks good! Thanks for the detailed docs and PR description. I think this PR fixes https://github.com/penrose/penrose/issues/1020?

Like @samestep said, it'd be great to add the trios in the teaser figure in the registry as well!

keenancrane:
Thanks for the feedback, @samestep and @wodeni!

I've also now added an example, and updated the registry accordingly.

Let me know if you spot anything else…

samestep (approved):

wodeni (approved):
Thanks for the examples! They look great

wodeni (approved):
Thanks for the examples! They look great

PR #1406, opened 2023-05-18
test: switch from jest to vitest

wodeni:
# Description

This PR switches our testing framework from jest to [vitest](https://vitest.dev/), which is ESM-first and much faster. 

* Vitest doesn't support imports without file extensions and only seems to work well with [Node's ECMAScript module resolution](https://www.typescriptlang.org/docs/handbook/esm-node.html), which we already use. Therefore, instead of relying on the bundler and testing framework to resolve the extensionless imports in `core`, we switched all imports in `core` to include the `.js` file extension. 
* Remove a lot of configuration since vitest requires a lot less of them compared to jest.
* Ported tests in both `core` and `edgeworth` to vitest and they worked well locally. However, `edgeworth`'s tests seemed to hang the CI. Since `main` only runs coverage on `core` anyway, this PR won't change the behavior of the test script.

# Checklist

# Open questions

* Why do `edgeworth` tests hang CI? Seems like a known issue: https://github.com/vitest-dev/vitest/issues/2008
* I get a `canvas` related error whenever I run tests. This is reported in https://github.com/vitest-dev/vitest/issues/274 and a plugin will be needed for this. We don't actually produce diagrams in the tests, so this is really just an artifact of us doing mathjax initialization at the top level in `CollectLabels`. 
```
Error: Not implemented: HTMLCanvasElement.prototype.getContext (without installing the canvas npm package)
```


samestep:
I'm probably misremembering; did you say there's a lint that allows us to enforce the `.js` prefix in imports?

wodeni:
> I'm probably misremembering; did you say there's a lint that allows us to enforce the `.js` prefix in imports?

Yeah https://github.com/mysticatea/eslint-plugin-node/blob/HEAD/docs/rules/file-extension-in-import.md

.github/workflows/build.yml
```diff
@@ -214,28 +214,8 @@ jobs:
...
-      - name: Make test report dir
-        run: mkdir -p reports/junit/
-        working-directory: packages/core/
-      - name: Run core tests and generate test data
-        run: npx nx run core:coverage
-      - name: Move test report
-        run: mv junit.xml reports/junit/
-        working-directory: packages/core/
-      - name: Upload coverage report to codecov
-        uses: codecov/codecov-action@v3
```
samestep, line 228: Given that we're deleting these lines, should we also delete `codecov.yml` from the repo root?

samestep (approved):
Thanks for doing this Nimo! I'm cool with merging this PR under the following caveats:

- see the minor comment I left above about `codecov.yml`
- we should open an issue about actually running Edgeworth tests in CI
- we should open an issue about the canvas warnings in Edgeworth tests
- we should open an issue about switching from `"moduleResolution": "node"` to `"moduleResolution": "node16"` in `optimizer`, `components`, `edgeworth`, and `editor`

samestep:
> > I'm probably misremembering; did you say there's a lint that allows us to enforce the `.js` prefix in imports?
> 
> Yeah https://github.com/mysticatea/eslint-plugin-node/blob/HEAD/docs/rules/file-extension-in-import.md

I no longer think we need to bother adding such a lint, because this is automatically enforced by TypeScript when we set `"moduleResolution": "node16"`.

.github/workflows/build.yml
wodeni: yeah, I don't think historical coverage data help us much so it's not worth the maintenance

wodeni: done

wodeni:
> * we should open an issue about actually running Edgeworth tests in CI

#1415 

> * we should open an issue about the canvas warnings in Edgeworth tests

#1416

> * we should open an issue about switching from `"moduleResolution": "node"` to `"moduleResolution": "node16"` in `optimizer`, `components`, `edgeworth`, and `editor`

#1417

PR #1411, opened 2023-05-19
feat: shape clipping

### LINKED ISSUE #1405: Expose clip paths in Style
Thanks to @liangyiliang, groups are supported as of #1294. This addition paves the way to support clip paths.  As the name suggests, a clip path will clip a shape or group of shapes by another shape:

image

The proposal is to add a field `clipPath:` to the `Group` shape in Style, a la
```
shape G = {
   shapes: [ S0, S1, S2, S3 ]
   clipPath: P
}
```
where `P` is another shape.  The generated SVG code will then follow this pattern:
```
<g>
   <defs>
      <shape id="UNIQUE_SHAPE_ID" ... /> <!-- defines the shape of the clip path -->
   </defs>
   <clipPath id="UNIQUE_CLIPPATH_ID"> <!-- defines the clip path itself -->
      <use xlink:href="#UNIQUE_SHAPE_ID" style="overflow:visible;"/>
   </clipPath>

   <!-- each shape then references this clip path -->
   <shape style="clip-path:url(#UNIQUE_CLIPPATH_ID)" ...>
   <shape style="clip-path:url(#UNIQUE_CLIPPATH_ID)" ...>
   <shape style="clip-path:url(#UNIQUE_CLIPPATH_ID)" ...>
   ...
</g>
```

Since a clip path may get used by multiple groups, it should be defined outside the group.  Also, in some cases one might wish to render the clip path as a visible shape; in other cases, they don't.  So, we can also add another field to all shape types, called `rendered:`.  E.g.,

```
shape S = Shape {
   rendered: false
}
```

Finally, for the purpose of constraints and objectives, the shape associated with a clipped group will simply be the shape of the clip path.  This convention makes sense, since a clipped group basically describes a "textured region," where the shape of the region just depends on the clip path itself.

liangyiliang:
# Description

Resolves #1405.

We implement shape-clipping and improve a couple of library computation and constraint functions to work with clipped shapes.

# `Group` Shape `clipPath` Property

A `Group` shape now has an additional `clipPath` property that takes a `ClipDataV` value. A `ClipDataV` value has two variants: `NoClip` (which takes no additional data) and `Clip` (which takes in a `Shape` object). A `ClipDataV` can be constructed using one of two computation functions:
* Function `noClip` takes no parameters and returns the `NoClip` variant.
* Function `clip` takes in a shape and returns the `Clip` variant with the given shape.

The default value of the `Group` shape's `clipPath` property is the `NoClip` variant.

As an example, suppose we have shapes `s1` and `s2` in a group, and we want to clip the group by shape `s3`. We can write,
```
g = Group {
  shapes: [s1, s2]
  clipPath: clip(s3)
}
```

A shape involved in the `clipPath` field of a `Group` is considered a part of the group through all stages of the compilation, including the construction of the `GroupGraph`, so all the usual rules of shapes not belonging to multiple groups still apply. That is,
* A shape used in `clipPath` of a group cannot appear in any other group, and cannot be used to clip another group; and
* A shape in a group cannot be used to clip another group.

# Corresponding SVG

Since a shape in `clipPath` is considered a group member, we render them with all other group members. For a group with `clipPath`, the generated SVG has structure,
```
<g>
  <clipPath id="some-id">
    ... (clipPath shape rendered here)
  </clipPath>

  <shape id="group-member-1" clip-path="url(#some-id)"> ... </shape>
  <shape id="group-member-2" clip-path="url(#some-id)"> ... </shape>
  <shape id="group-member-3" clip-path="url(#some-id)"> ... </shape>
  ... (other group members)
</g>
```

The `clipPath` SVG object id `some-id` is the name of the clipping shape, with suffix `-clip`. For example, if the clipping shape name is `myShape`, then the `clipPath` SVG object id is `myShape-clip`.

# Bounding Box

Previously, the bounding box of a `Group` shape is simply the min-max-bbox. That is, take the bounding boxes of all group members, and then find the min-x, min-y, max-x, max-y, and use those to construct a new bounding box.

If a `Group` shape has a clipping shape, we compute the new bounding box as the previous bounding box, intersecting the bounding box of the clipping shape.

image

In the picture above, the bounding box of the yellow region (the larger ellipse clipped by the smaller ellipse) is computed as the bolded blue rectangle. One may observe that this bounding box is not optimum. Indeed, this box is not a "minimum bounding box" as it can overestimate the object. But this is relatively simple to compute.

# Constraint `contains`

We add a branch to the `contains` constraint to deal with groups.

If a group does not have a clipping shape, checking whether a group contains a shape is equivalent to checking whether _some_ group member contains the shape. We simply compute the `contains` constraint penalty between all of the group members and the shape, and take the minimum. If the minimum of the penalties is small, then one group member must contain the shape.

If a group has a clipping shape, then the `contains` constraint between the group and another shape just requires:
* the group members (excluding the clipping shape) contains the other shape, and
* the clipping shape contains the other shape.

We compute both constraints, and require that both constraints hold.

https://clip.penrose-72l.pages.dev/try/?gist=11c14ac7a7e0cf45a13ec68066925b18

This example shows two sets, with their intersection between marked out using clipping. We also demonstrate the bounding box of the intersection (see the rectangle) and the `contains` constraint (the small, red circle is within the intersection).

# Checklist


keenancrane:
This looks good from the language design end.  It's a bit unfortunate we can't just do `clipPath: shape`, but I understand the reasons from an implementation perspective.  Thank you for updating the documentation!

wodeni:
Merged `main` into this branch and tested clip paths with the diagram variation grid. Turns out that we need to namespace the ids of clip paths. I chose to use the namespace instead of `makeIdsUnique` because the latter doesn't behave well on a single SVG element. We use `makeIdsUnique` for opaque imports and generally favor namespaces for id-uniqueness.

liangyiliang:
Thanks for catching this!

wodeni (approved):
Looks great! Thanks for adding this feature. As mentioned, I fixed one issue with id-uniqueness. I also added inline comments for `makeIdsUnique` to clarify its purpose (which is not used in this case so I explained why there) and for the implicit group membership of clipping shapes.

PR #1423, opened 2023-05-24
fix: fill and stroke settings for `Path`, `Line`, and `Polyline`

### LINKED ISSUE #1421: SVG renderer should explicitly write "none" fields
For paths where `fillStyle` is set to `none()` in Style (which is the default for several shapes), the renderer currently omits an explicit `fill` tag.  This leaves the default fill open to interpretation by other programs.  For instance, a diagram that shows up like this in Chrome:

image

Shows up instead like this in Illustrator, which assumes a default fill color of black (and draws a hairline fill for any stroke):

image

The solution is simply to render such shapes with an explicit `fill="none"` attribute.

### LINKED ISSUE #1436: Bug: Polyline ignores `strokeLinecap`
image

The `Polyline` shape seems to ignore the `strokeLinecap` field, even though (i) this field is properly handled by `Line`, and (ii) even if if the linecap is not an explicit field of `Polyline`, it should be handled by SVG passthrough.

Here's a complete example comparing the behavior of `Line` and `Polyline` for a round linecap.  Inspecting the generated SVG confirms that the `<polyline>` element has an attribute `stroke-linecap="butt"` (which is the default value), rather than `"round"`.

**.domain**
```
type Line
type Polyline
```

**.substance**
```
Line L
Polyline M
```

**.style**
```
canvas {
    width = 100
    height = 100
}

forall Line L {

    vec2 a = (?,?)
    vec2 b = (?,?)

    shape L.icon = Line {
        start: a
        end: b
        strokeLinecap: "round"
        strokeColor: #0a0
        strokeWidth: 5
    }

    shape L.title = Text {
        string: "Line"
        center: (a+b)/2
        fillColor: #7c78
    }
}

forall Polyline M {

    vec2 a = (?,?)
    vec2 b = (?,?)

    shape M.icon = Polyline {
        points: [ a, b ]
        strokeLinecap: "round"
        strokeColor: #a00
        strokeWidth: 5
    }

    shape M.title = Text {
        string: "Polyline"
        center: (a+b)/2
        fillColor: #c998
    }
}
```

wodeni:
# Description

Resolves #1421. Resolves #1436

This PR fixes a few small issues in the renderer:
* `Line` and `Path` should reuse `attr*` code for fill and stroke
* `Line` should be rendered as `<line>` (@keenancrane)
* `Line` should support both fill and stroke properties
* #1421
* #1436: now `Line`, `PolyLine`, and `Path` support `strokeLinecap`, according to the [w3d docs](https://www.w3docs.com/learn-css/stroke-linecap.html). 

There are lots of registry diagram changes because of the renderer changes, but most of them should look the same visually.

https://fix-fill-none.penrose-72l.pages.dev/try/?gist=998cf0181e4a6a1dc2418e435c759ae5
https://fix-fill-none.penrose-72l.pages.dev/try/?gist=f23fb193260fce8e54a82dfcad89ca30

# Checklist


packages/core/src/shapes/Polyline.ts
```diff
@@ -15,7 +16,9 @@ export interface PolylineProps<T>
     Stroke<T>,
     Fill<T>,
     Scale<T>,
-    Poly<T> {}
+    Poly<T> {
+  strokeLinecap: StrV;
```
samestep, line 20: Could you explain why this is only in `Polyline` and not in `Stroke`?

samestep (approved):
Awesome, thanks! So weird that we weren't rendering `Line` as `<line>` before :joy:

packages/core/src/shapes/Polyline.ts
wodeni: Wasn't too sure about adding them to other `Stroke` shapes like `Path` and `Polygon` because they can be closed paths. From the [docs](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linecap) this seems to be fine. Do you want me to add it to `Stroke` instead?

samestep: Either way is fine with me! If you don't add it to `Stroke` then could you write your reasoning (as you've written here) as a code comment on the `strokeLinecap` attribute?

Also, unrelated but now that I think about it, don't we also need to remove this from our passthrough JSON file? (I forget where that is...)

wodeni: Don't think I'm aware of this file?

wodeni: Found some docs on this property. Looks like only `Line`, `Polyline` and `Path` can be affected by this prop. So I'm not adding this to all stroke-able shapes. Added some comments in the PR desc and [08c284d](https://github.com/penrose/penrose/pull/1423/commits/08c284d90c877af29d5948e0f38f93db17720e5a)

PR #1205, opened 2023-01-09
feat: initial type definitions for function interface

maxkrieger:

PR #897, opened 2022-02-28
chore: Add initial documentation to Docusaurus site

cmumatt:
# Description

This PR updates the TypeDoc dependency, adds the TypeDoc Markdown plugin, and begins to fill out the documentation on the Docusaurus site.  It also removes the empty Blog and adds a team page, currently without headshots.  

Due to poor results auto-generating Markdown docs from TypeDoc function dictionaries, this PR does not include an automated pipeline for function doc generation.  This and language documentation will be addressed in subsequent PRs.  

# Implementation strategy and design decisions

Load initial documentation into Docusaurus site.

# Checklist

- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new ESLint warnings
- [X] I have reviewed any generated changes to the `diagrams/` folder

# Open questions

- Auto-generate function documentation, possibly using the JSON output from TypeDoc
- Add language documentation

PR #1427, opened 2023-05-24
docs: auto-generate documentation for Style functions

wodeni:
# Description

Following up on #897, #1205, and #1352, this PR generates documentation for Style functions: constraints, objectives, and computation functions. This PR only aims to make sure the docs are up to date, and style/organization issues should be resolved in future PRs. 

* `core` is now a `depDependency` of `docs-site` because the site directly reads from `core` to retrieve the functions
* Vitepress require components, and by extension all imports in components, to be SSR-friendly. Therefore, the browser-specific code in `core` initially causes errors in the production build but not the dev build since they build things differently.
  * To workaround this, I moved the top-level browser-specific code in `core` into functions. There's a potential performance cost of re-running some initialization code repeatedly, but none is observed so far.

https://autodocs.penrose-72l.pages.dev/docs/ref/style/functions

# Checklist

# Open questions

There are some stylistic issues to address:
* Should functions be on separate pages or stay in one single page?
* How can we add handwritten docs to individual functions?


packages/docs-site/package.json
```diff
@@ -42,10 +42,13 @@
...
-    "@penrose/roger": "^3.0.0-alpha.1",
+    "@penrose/core": "^3.0.0-alpha.0",
     "@penrose/editor": "^3.0.0-alpha.0",
+    "@penrose/roger": "^3.0.0-alpha.1",
+    "@types/markdown-it": "^12.2.3",
+    "markdown-it": "^13.0.1",
     "markdown-it-katex": "^2.0.3",
     "shx": "^0.3.3",
-    "vitepress": "^1.0.0-alpha.35"
+    "vitepress": "^1.0.0-beta.1"
```
samestep, line 52: Ah cool! Did this version bump change anything?

packages/docs-site/docs/ref/style/functions.md
```diff
@@ -1,3265 +1,36 @@
...
-| `x`  | `IVarAD` |
-| `x0` | `IVarAD` |
-| `x1` | `IVarAD` |
-
-#### Returns
-
-`IVarAD`
-
-#### Defined in
+<div v-for="(f, index) in constrDict" :key="index">
```
samestep, line 10: I'm curious, why do you set `:key` here but not in the other two sections?

packages/core/src/utils/CollectLabels.ts
```diff
@@ -257,9 +262,13 @@ const measureTextContext = measureTextElement.getContext("2d")!;
  * @returns `TextMeasurement` object and includes data such as `width` and `height` of the text.
  */
 export function measureText(text: string, font: string): TextMeasurement {
+  const measureTextElement = document.createElement("canvas");
+  const measureTextContext = measureTextElement.getContext("2d")!;
   measureTextContext.textBaseline = "alphabetic";
   measureTextContext.font = font;
   const measurements = measureTextContext.measureText(text);
+  // TODO: removing the element causes issues of "NotFoundError: The node to be removed is not a child of this node."
+  // document.removeChild(measureTextElement);
```
samestep, line 271: This makes sense, right? Calling [`document.createElement`](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement) just creates a DOM node, it doesn't actually attach it as a child of `document` as far as I know.

samestep (approved):
Awesome, thanks for finally doing this! :smiley:

packages/docs-site/package.json
wodeni: Not really. Was attempting this to fix the SSR error in production build. Didn't end up helping but I figured it's better to use the beta version instead of alpha anyway.

packages/docs-site/docs/ref/style/functions.md
wodeni: No particular reason. Removing now.

packages/core/src/utils/CollectLabels.ts
wodeni: Right. I meant to call `remove()` instead. Seems to work just fine.

PR #1429, opened 2023-05-25
fix: dependency graph for paths with indices

liangyiliang:
# Description

This fixes a bug in the generation of the dependency graph when it sees `Path` expressions with indices.

# What broke?

Conceptually, the expression `arr[index]` depends on both `arr` and `index`. When handling the dependency graph for this expression, it only adds `arr`, not `index`. This sometimes result in `arr[index]` being evaluated before `index` even exists in the translation.

# The fix

In `findPathsExpr` we modify the case for `Path` to return the `arr` part and the `index` part:
```
      return [
        {
          ...expr,
          indices: [],
        }, // this is the `arr` part
        ...expr.indices.flatMap((index) => findPathsExpr(index, context)), // this is the `index` part
      ];
```

# Checklist

### DIFF
--- a/packages/core/src/compiler/Style.test.ts
+++ b/packages/core/src/compiler/Style.test.ts
@@ -1336,16 +1336,21 @@ delete x.z.p }`,
       const sty =
         canvasPreamble +
         `
-forall T t {
-  t.xs = [1, 2, 3, 4, 5, 6]
-  t.index = 3
-  t.v = t.xs[t.index]
-}
+        forall T t {
+          t.vals = [1, 2, 3, 4, 5, 6]
+          t.val = t.vals[match_id]
+        
+          Circle {
+            r: t.val
+          }
+        }
       `;
 
+      // This problem would have failed compilation when indexing is not handled correctly
+      // And this would fail:
       const { graph } = await loadProgs({ dsl, sub, sty });
-      expect(graph.parents("`t`.v").sort()).toEqual(
-        ["`t`.xs", "`t`.index"].sort()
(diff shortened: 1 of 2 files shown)


samestep (changes requested):
Can you add some sort of test which demonstrates this?

packages/core/src/compiler/Style.test.ts
```diff
@@ -1328,4 +1328,25 @@ delete x.z.p }`,
...
+  t.xs = [1, 2, 3, 4, 5, 6]
+  t.index = 3
+  t.v = t.xs[t.index]
+}
+      `;
+
+      const { graph } = await loadProgs({ dsl, sub, sty });
+      expect(graph.parents("`t`.v").sort()).toEqual(
+        ["`t`.xs", "`t`.index"].sort()
+      );
```
samestep, line 1349: Thanks for adding this test! This isn't really what I meant, though; could you add a test that demonstrates a Style program that should succeed, but failed before this PR? You can use this same Style program you've already written here if you want, but test it in such a way that you're actually checking that running it works, rather than just looking at some intermediate state of the compiler.

samestep (changes requested):

packages/core/src/compiler/Style.test.ts
liangyiliang: Sounds good - I'll see if I can add this test tonight.

samestep (approved):
Looks good to me!

wodeni (approved):
LGTM!

PR #1433, opened 2023-05-30
feat: export SVGs with plain TeX labels

### LINKED ISSUE #303: Support export to TeX with labels
Currently we dump labels as part of the output image itself, which is useful in contexts where the image will be used directly.  When the image is meant to be used as a figure in a TeX document, it would be nice to actually typeset the labels via TeX itself.  This can be done via the [`overpic` package](https://ctan.org/tex-archive/macros/latex/contrib/overpic?lang=en), which superimposes TeX math over a PDF image.

Here's an example:

```
\begin{overpic}{images/ddg_surface_patch.pdf}
   \put(210.6387,134.2822){\(f\)}
   \put(64.3818,33.7637){\(X\)}
   \put(428.127,64.626){\(f(M)\)}
   \put(116.3086,27.2295){\(M\)}
   \put(344.8896,62.5127){\(\dfX\)}
   \put(377.1357,99.0137){\(N\)}
\end{overpic}
```

This TeX code will display the image `ddg_surface_patch.pdf`, and then place the given labels on top of it at the specified (x,y) coordinates (see `overpic` docs for details).

To incorporate this into Penrose we need to make a few small changes:

- Don't output label images in the final SVG
- Output a TeX file with the (x,y) coordinates and label strings, following the `overpic` example above
- Make sure to perform the correct conversion from the Penrose coordinate system to the TeX/PDF coordinate system, namely: (i) all coordinates should be relative to the corner of the document, and (ii) the scale between Penrose units and PDF units should be properly accounted for

(It's really this last bullet point that will likely be most painful to get right…)

wodeni:
# Description

Resolves #303.

This PR implements a version of option (1) in #303 :

> Automatically convert all Equation shapes to Text shapes, replacing the string S with the string $S$ so that it gets typeset as math when imported into TeX. (Note that with this conversion, existing Text shapes are unaffected will still get rendered properly within TeX.)

The implementation strategy is simpler than the proposal: we allow an additional flag (`texLabels`) to `RenderStatic` to indicate if we want to render TeX strings. This flag is then passed to the `Equation` renderer, which will render the equation as a plain `<text>` element with surrounding `$`s so that they can be rendered by an external TeX engine like the `svg` package mentioned in #303. We explicitly forward the following properties to this `<text>` object:

* fillColor
* fillOpacity
* width
* height

Try to export SVGs with plain TeX labels either in the IDE or `roger`. In the IDE, click "SVG (TeX)" button. In `roger`, run the following in `packages/examples`:

```
roger trio src/set-theory-domain/tree-venn-3d.trio.json --tex-labels > out.svg
```
Either outputs should have plain TeX strings in the raw SVG, which will be rendered in TeX if imported through `\includesvg{}`.

# Checklist

# Open questions

* TODO: docs in `docs-site` about TeX import support

### DIFF
--- a/packages/core/src/renderer/Equation.ts
+++ b/packages/core/src/renderer/Equation.ts
@@ -1,4 +1,8 @@
+import seedrandom from "seedrandom";
+import { input } from "../engine/Autodiff.js";
+import { makeCanvas } from "../index.js";
 import { Equation } from "../shapes/Equation.js";
+import { sampleText } from "../shapes/Text.js";
 import { getAdValueAsString } from "../utils/Util.js";
 import {
   attrAutoFillSvg,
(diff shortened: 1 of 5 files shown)


wodeni:
I'm merging this PR, noting two remaining issues, both of which will be addressed by #measure:
1. #1448: this will be addressed by #1445

Issue #1261, opened 2023-01-24
Crop gallery SVGs for preview in `editor`


wodeni:
#1212 experiments with preview thumbnails, but the result doesn't look good enough for production use. Notably, the diagrams sometimes contain too much whitespace. Ideally there should be a system utility for cropping a Penrose-generated diagram based on shape dimensions (e.g. bounding boxes). 

![image](https://user-images.githubusercontent.com/11740102/214394896-2dd7c1b8-536d-498d-b70e-90fde03df2cd.png)

Issue #1288, opened 2023-02-02
Name collisions in SVG `Image` import


wodeni:
**Describe the bug**

When there are more than one diagram on the page, imported images might not show up. This is usually caused by name collisions among the imported SVGs. If an imported SVG defines a `<linearGradient id="foo">` and refer to `foo` by `url(#foo)` somewhere else, there must be only one such reference in the entire DOM tree. 

**To Reproduce**

1. Go to https://361060b0.penrose-72l.pages.dev/try/
2. Click on the "Walk on Spheres" example
3. Select `Grid`
4. See the following output

![image](https://user-images.githubusercontent.com/11740102/216257784-d03da57f-9675-4374-9287-a408a0677d38.png)

**Expected behavior**

Spheres should be shaded by the imported image:

![Walk on Spheres - Laplace Estimator](https://user-images.githubusercontent.com/11740102/216257960-910934b9-8d6c-4093-bb62-5ae578389937.png)

**Additional context**
This is a known problem with SVG (e.g. https://github.com/gregberge/svgr/issues/150) and there are some SVG cleaning tools like [svgr](https://github.com/gregberge/svgr). It would be great to experiment with some of them.

wodeni:
#1444 revealed another likely case of this issue, which I bumped into again in #1464. Reopen to investigate.

keenancrane:
> It would be great to experiment with some of them.

Beyond being "great" to experiment with, I've flagged this as a high-priority issue.  Let's fix it now.

In general our SVG support has really suffered regression with some recent PRs.  Correct SVG import is a central feature of our system.  Let's please not sacrifice it merely for the sake of "nice to have" features like the gallery, preview icons, build system niceties, etc.

PR #1444, opened 2023-06-01
feat: `renderer` adds tightest viewbox metadata to SVG

rjainrjain:
# Description

Resolves [#1261.](https://github.com/penrose/penrose/issues/1261)

This PR takes the approach described in [#1261](https://github.com/penrose/penrose/issues/1261) of providing a way of 

> cropping a Penrose-generated diagram based on shape dimensions (e.g. bounding boxes)

by calculating the tightest cropped viewbox based on all shapes in the diagram and adding the viewbox values as metadata in the `<penrose>` tag in the exported SVG. Replacing the viewbox values with these cropped values will solve the problem of sub-production level diagram previews. 

Rather than modifying the generated SVG after the `renderer` is through with it, as initially planned, we calculate the cropped viewbox values using the bounding boxes of the `shapes: Shape<Num>` generated in `RenderStatic`, adding the metadata then as well. In [DiagramPanel.tsx](https://github.com/penrose/penrose/blob/9bca496d09fa2e6624dd87991f3f8f411c94b1f4/packages/editor/src/components/DiagramPanel.tsx#L4) we add a check to create a new metadata tag if there is none, or simply to add children to the one ostensibly created in the `RenderStatic` step. In [ExamplesBrowser.tsx](https://github.com/penrose/penrose/blob/dc7d41f031cb7a8aad65ab09d139c7ac536f6f75/packages/editor/src/components/ExamplesBrowser.tsx#L4) we set the preview SVG's viewbox to the cropped values if the metadata is present, and display the SVG previews.

In the Penrose editor, click on the first example and inspect the SVG. It should have the following metadata.

`
<penrose><croppedViewBox>32.315104085276914 0.7731797378698957 601.3836522786437 601.3836522786437</croppedViewBox></penrose>
`

After clicking the SVG download button in the editor, inspecting once more will show the following metadata.

```
<penrose><croppedViewBox>32.315104085276914 0.7731797378698957 601.3836522786437 601.3836522786437</croppedViewBox><version>0.1</version><variation>PlumvilleCapybara104</variation><sub>Set A, B, C, D, E, F, G

IsSubset(B, A)
IsSubset(C, A)
IsSubset(D, B)
IsSubset(E, B)
IsSubset(F, C)
IsSubset(G, C)

Not(Intersecting(E, D))
Not(Intersecting(F, G))
Not(Intersecting(B, C))

AutoLabel All
</sub><sty>canvas {
  width = 800
  height = 700
}

forall Set x {
  x.icon = Circle {
    strokeWidth : 0
  }

  x.text = Equation {
    string : x.label
    fontSize : "25px"
  }

  ensure contains(x.icon, x.text)
  encourage sameCenter(x.text, x.icon)
  x.textLayering = x.text above x.icon
}

forall Set x; Set y
where IsSubset(x, y) {
  ensure smallerThan(x.icon, y.icon)
  ensure disjoint(y.text, x.icon, 10)
  ensure contains(y.icon, x.icon, 5)
  x.icon above y.icon
}

forall Set x; Set y
where Not(Intersecting(x, y)) {
  ensure disjoint(x.icon, y.icon)
}

forall Set x; Set y
where Intersecting(x, y) {
  ensure overlapping(x.icon, y.icon)
  ensure disjoint(y.text, x.icon)
  ensure disjoint(x.text, y.icon)
}
</sty><dsl>type Set

predicate Not(Prop p1)
predicate Intersecting(Set s1, Set s2)
predicate IsSubset(Set s1, Set s2)
</dsl></penrose>
```
# Checklist

# Open questions

Whether to provide an option in the editor to switch to cropped view.


packages/editor/src/components/ExamplesBrowser.tsx
```diff
@@ -9,25 +9,44 @@ export default function ExamplesBrowser() {
...
-        </div>
-      ))}
+      {examples.contents.map((example, k) => {
+        const svgDoc = parser.parseFromString(
+          example.preview!,
+          "image/svg+xml"
+        );
+        const cropped = svgDoc.querySelector("croppedViewBox")?.innerHTML;
+        const svgNode = svgDoc.querySelector("svg")!;
+        if (!(cropped === undefined)) {
```
wodeni, line 25:
```suggestion
        if (cropped !== undefined) {
```

Can also use [strict inequality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality) here to get rid of the parens.

packages/editor/src/components/ExamplesBrowser.tsx
```diff
@@ -9,25 +9,44 @@ export default function ExamplesBrowser() {
...
+          svgNode.setAttribute("viewBox", cropped!);
```
wodeni, line 26:
```suggestion
          svgNode.setAttribute("viewBox", cropped.innerHTML);
```

This should typecheck without the assertions?

packages/editor/src/components/ExamplesBrowser.tsx
```diff
@@ -9,25 +9,44 @@ export default function ExamplesBrowser() {
...
-            </div>
-          </FileButton>
-        </div>
-      ))}
+      {examples.contents.map((example, k) => {
+        const svgDoc = parser.parseFromString(
+          example.preview!,
+          "image/svg+xml"
+        );
+        const cropped = svgDoc.querySelector("croppedViewBox")?.innerHTML;
```
wodeni, line 23:
```suggestion
        const cropped = svgDoc.querySelector("croppedViewBox");
```

See below for doing stuff after `undefined` check

wodeni (approved):
Great job on this PR @rjainrjain 💯! Very nice PR description and clean implementation. Looking forward to your future PRs :D.

Also noting a weird thing about previews with imported SVGs. Don't have to solve this in this PR, but perhaps this has something to do with https://github.com/penrose/penrose/issues/1288.
image

wodeni:
Should've suggested just one change instead of 3 :(. My bad! Hopefully you can use the ["Add suggestion to batch"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request) feature, which is kinda new?

packages/editor/src/components/ExamplesBrowser.tsx
rjainrjain: hmm, I tried this out, along with the suggestion below — it still requires the assertion at some point because `cropped` is still possibly `null` when passing `cropped.innerHTML` to `setAttribute`.

Issue #1438, opened 2023-05-31
`Equation` font size is wrong / canvas size dependent


keenancrane:
![image](https://github.com/penrose/penrose/assets/3604525/00b2f4e3-165b-4733-a584-5f42b0a5a079)

`Equation` shapes (rendered via MathJax) do not yield the size requested in the `fontSize` field—the rendered size depends on the `canvas` size in the Style program.  In contrast, `Text` shapes always get rendered at the proper size, independent of canvas dimensions.

Here's a complete example showing the issue (the Domain schema and Substance program are empty):
```
canvas {
    width = 10
    height = 8
}

global {
    shape plainText = Text {
        string: "A"
        fillColor: #0a0
        fontSize: "3px"
        fontFamily: "MathJax_Math"
    }
    shape plainTextLabel = Text {
        string: "Text"
        center: plainText.center - (0,1.75)
        fontSize: "1px"
        fillColor: plainText.fillColor
        fontFamily: "monospace"
    }

    shape mathText = Equation {
        string: "A"
        fillColor: #00a
        fontSize: "3px"
    }
    shape mathTextLabel = Text {
        string: "Equation"
        center: mathText.center - (0,3)
        fontSize: "1px"
        fillColor: mathText.fillColor
        fontFamily: "monospace"
    }

    ensure disjoint( mathText, plainText )
}
```

wodeni:
To repro this, I also need to download the `MathJax_Math` font and install it locally:
https://imathas.valenciacollege.edu/mathjax/fonts/HTML-CSS/TeX/otf/

Issue #340, opened 2020-05-13
Label point sizes appear visually different despite being set to the same size in Style


hypotext:
I made some simple examples that Keenan requested for the Penrose blog post / Twitter thread, but am running into a weird font size issue. 

@maxkrieger @wodeni can you take a quick look and let me know how to address this ASAP, or if there's a (reasonably principled) workaround? It's blocking us from posting examples :)

Repro with minimal example: on branch `set-example`, do `runpenrose set-theory-domain/intersection-venn-diagram.sub  set-theory-domain/venn-simple.sty  set-theory-domain/setTheory.dsl`

Expected result: all labels have the same size, 15 pt

Actual result: "Circle" label looks smaller than the other two

![image](https://user-images.githubusercontent.com/2762356/81773904-60ab8180-94b7-11ea-8012-6c1168c82be8.png)

Tried: 
* changing the label size in Style (seems to actually set the label size)
* running examples with other Styles ("Circle" seems to be consistently mis-sized across Style programs)
* looking at DOM (svg indeed has fontSize set to the style size, but changing it in chrome does not change the font size live, as noted in https://github.com/penrose/penrose/issues/301#issue-529443958)
* changing the label text (I can't consistently repro the mis-sizing, e.g. if I change the labels to strings of different sizes, like "Z", "ZZZZ", and "ZZZ ZZZ ZZZZ", the "Z"s all appear the same size, so I'm not sure if it's an aspect ratio problem)

I also don't know if all three labels are actually mis-sized, but only "Circle" is the visually noticeable one.

Related issues: https://github.com/penrose/penrose/issues/301#issue-529443958

hypotext:
Actually, if I set the height of the label to be constant, it seems to look right:

![image](https://user-images.githubusercontent.com/2762356/81774552-d237ff80-94b8-11ea-8707-925ee6c78d97.png)

Resolved for my immediate purposes, but definitely unintuitive to a user

maxkrieger:
this is so weird. Do you think this is a regression from prior behavior or a latent bug? It's possible that it's MathJax3 behaving differently than 2. One way to test this is comparing to this commit, the last one before we added MathJax3: https://github.com/penrose/penrose/commit/3f077d9a55fbe498424dc6d3822132f69ded1436

(I really need to learn how git bisect works. Ive been using git checkout to do this manually)

hypotext:
[...]

PR #339, opened 2020-04-22
Fix newline support for MathJax v3

maxkrieger:
Due to MathJax v3 not supporting newlines (bug report: https://github.com/mathjax/MathJax/issues/2418), several examples (eg `linearMap_add.sub`) cannot optimize fully due to their horizontal length.

PR #1445, opened 2023-06-01
feat: accurate `Equation` measurements

### LINKED ISSUE #1448: "SVG (TeX)" export in `editor` needs to align `Equation`s by baseline
**Describe the bug**

#1433 added support for TeX-importable SVGs. However, he TeX `svg` package is drawing labels with the `y` on the baseline (right), whereas we are drawing them with the y on the geometric center (left).

image

To work around this, our TeX SVG export need to shift the `Equation` center to the baseline. To do so, we need to know the descent of the `Equation`, and shift `center` by `(0, height/2 - descent)`.

wodeni:
# Description

Resolves https://github.com/penrose/penrose/issues/1438, resolves #1448.
Related: https://github.com/penrose/penrose/pull/842

For #340, we introduced a way to scale MathJax-generated SVGs such that `Equations` have visually-consistent sizes in a single diagram. However, #1438 points out that this scaling conversion doesn't produce labels of accurate font size. This PR does exact conversion from MathJax SVG units to pixels for `width`, `height`, `descent`, and `ascent` of our `Equation` shape.

[The CSS standard](https://www.w3.org/Style/Examples/007/units.en.html#units) defines all the units, which is implemented by the MathJax source code.

When MathJax generates an SVG, the `width`, `height`, `viewBox`, and `vertical-align` is computed in https://github.com/mathjax/MathJax-src/blob/a5ae9485cb7441fdd5ea59645cfbd1c12b7d53e1/ts/output/svg.ts#L239. 

Our implementation computes the `width`, `height`, `descent`, and `ascent` by using these values from above. See inline comments for more details.

https://tex-measure.penrose-72l.pages.dev/try/?gist=5c192c1186607e52d56b86e48ab7c3f5
https://tex-measure.penrose-72l.pages.dev/try/?gist=5e4f9cd917276e11ebaa8e798a4a62c2

# Checklist

# Open questions

* Now that we parse the `fontSize` of `Equation`, only absolute units are accepted for now. The conversion to relative units such as `em` and `pt` needs font-specific information, and there's no guarantee that `Text` is going to render the same way. It is possible to support them, but this is requires careful work on both `Text` and `Equation`. Would love some help on supporting them! Given my own time constraints I'm inclined to convert existing relative `fontSize`s to absolute units so we can merge this PR.
* I reverted the fix for #339 because it causes MathJax to return a constant height no matter the content of the label. We may consider supporting it, but this requires parsing the content and finding TeX linebreaks. I think this is out of scope as well.


wodeni:
Note on relative units: although `pt` is nominally an "absolute" unit, MathJax treats it as a "relative" unit like `em`. I'm not sure if this has any implications to Penrose, but I am erring on the safer side of not providing support for it for now. More experimentation is needed to find out how `pt` is handled by both `Text` and `Equation` to make sure there's a one-to-one relationship between their sizes.

samestep:
Glancing at the registry diff, what's going on with `exterior-algebra/vector-wedge.svg` and the `matrix-ops` examples?

samestep (approved):
Looks great to me! Thanks so much for doing this deep dive :smiley:

PR #1452, opened 2023-06-03
fix: baseline center for TeX SVG `Equation`s

wodeni:
# Description

Resolves #1445.

Follow up on #1445, this PR shifts all `Equation`s exported as TeX SVG by `h/2 - descent` to get the center on the baseline, like the TeX `svg` package does. 

image


keenancrane:
LGTM! And tested/validated in real-world context.  Squashing & merging…

PR #1325, opened 2023-02-27
docs: start a blog

samestep:
# Description

Resolves #issue.

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

# Implementation strategy and design decisions

In the Wasm blogpost, here are the specific commits where the "performance difference" charts come from:

- Before: #1172
- After: #1183

# Examples with steps to reproduce them

# Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new ESLint warnings
- [ ] I have reviewed any generated registry diagram changes

# Open questions

Questions that require more discussion or to be addressed in future development:

PR #1457, opened 2023-06-05
fix: use `Serializer::json_compatible()`

samestep:
# Description

I didn't know this [existed](https://docs.rs/serde-wasm-bindgen/0.4.5/serde_wasm_bindgen/struct.Serializer.html#method.json_compatible) until writing more in #1325 just now. As a followup to #1191, this allows us to use maps in our exported Rust types and have [ts-rs](https://github.com/Aleph-Alpha/ts-rs) type them properly. I don't think we currently have any plans to do that, but it'll be nice to eliminate this potential footgun in case we eventually do.

# Checklist

### DIFF
--- a/packages/optimizer/src/lib.rs
+++ b/packages/optimizer/src/lib.rs
@@ -780,7 +780,7 @@ fn contains_nan(number_list: &[f64]) -> bool {
 }
 
 fn to_js_value(value: &(impl Serialize + ?Sized)) -> Result<JsValue, serde_wasm_bindgen::Error> {
-    // ts-rs expects `Option::None` to become `null` instead of `undefined`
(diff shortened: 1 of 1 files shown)


samestep:
> LGTM! Looks like the `null` vs. `undefined` thing is still true from [the docs](https://docs.rs/serde-wasm-bindgen/latest/serde_wasm_bindgen/struct.Serializer.html#method.json_compatible)? Not sure if this will matter.

Yep! According to [GitHub](https://github.com/cloudflare/serde-wasm-bindgen/tree/v0.4.5#serializer-configuration-options):

> You can also use the `Serializer::json_compatible()` preset to create a JSON compatible serializer. It enables `serialize_missing_as_null` and `serialize_maps_as_objects` under the hood.

PR #1466, opened 2023-06-07
chore: Revert "fix: Diagram panel empty on iOS Chrome and iOS/macOS Safari"

keenancrane:
Reverts penrose/penrose#1465

This PR caused display issues in our gallery (SVG gradients not rendering properly).

Since it's a minor change, and the most recent one, I'm rolling it back until we have a fix.


keenancrane:
This apparently did not actually fix the issue (seems I read too much into the `gallery` output…).

Still maybe better to determine the root issue before re-reverting the reversion. :-P

PR #1472, opened 2023-06-08
feat: additional constructor declaration syntax in Domain and Substance

### LINKED ISSUE #1402: Constructors should have the name of the type they construct
Currently, the `constructor` and `function` types in Domain schemas feel fairly redundant.  For instance, what is the difference between
```
function MakeTriangle( Point a, Point b, Point c ) -> Triangle
```
and
```
constructor MakeTriangle( Point a, Point b, Point c ) -> Triangle
```
Also, especially in the absence of sugared notation, we struggle to find good names for constructors (e.g., `MakeSegment`, `MakeTriangle`, etc.).  This ends up feeling like we're following the [OOP pattern in a language that doesn't explicitly support it](https://www.codementor.io/@michaelsafyan/object-oriented-programming-in-c-du1081gw2), i.e., we're just cooking up some convention for how to consistently name our functions so we know they're constructors.  But this isn't good for novice users, especially since Substance is our "novice language."

#### Proposal 2: Change the syntax used to invoke a constructor from `Type a := Constructor(...)` to `a := Type(...)` so that the type name doesn't redundantly appear in every Substance line that invokes a constructor.

For instance, running with the example above, a Domain schema might now look like
```
type Point
type Triangle
constructor Triangle( Point a, Point b, Point c )
```
and the corresponding Substance code would be
```
Point x, y, z
t := Triangle(x,y,z)
```
Here, `t` gets the type of the constructor (`Triangle`).

The Style selector wouldn't need to look much different in this case.  E.g., it might be
```
forall Triangle tri; Point a; Point b; Point c
where tri := Triangle(a,b,c)
```
The only small change here is that a type name (`Triangle`) can now be used on the righthand side of `:=`.

rjainrjain:
# Description

Resolves #1402.

This PR addresses the problem of "the absence of sugared notation" for constructor declaration and usage raised in #1402 by adding support for syntax that reflects the following proposals from #1402:

> **Proposal 1: Change the definition of `constructor` so that it always has the same name as the type it constructs.
Proposal 2: Change the syntax used to invoke a constructor... so that the type name doesn't redundantly appear in every Substance line that invokes a constructor.**

With respect to Proposal 1, we allow constructors to be declared in Domain without an output type if the name of the constructor is the same as the type it constructs. With respect to Proposal 2, we allow constructor decl-binds to be written without a type prefixing the variable, using a `Let` keyword instead, if the name of the constructor is the same as the type it constructs.

We also update the constructor syntax for all [geometry-domain](https://github.com/penrose/penrose/tree/f925044f7bba5c563aaee51f7b655a7b20bb3719/packages/examples/src/geometry-domain) examples to reflect and demonstrate these new additions.

Since this PR allows additional syntax rather than replacing the original constructor syntax, we add a rule to `constructor_decl` in [Domain.ne](https://github.com/penrose/penrose/blob/f925044f7bba5c563aaee51f7b655a7b20bb3719/packages/core/src/parser/Domain.ne#L4) such that the postprocessing step creates a `TypeConstructor<C>`with the constructor name as the typename and gives that to the `type` field of the `Decl<C>`. In [Substance.ne](https://github.com/penrose/penrose/blob/f925044f7bba5c563aaee51f7b655a7b20bb3719/packages/core/src/parser/Substance.ne), we add an `if` check to the `decl_bind` rule that checks for the `Let` keyword—in which case it sends the constructor name as the type, wrapped in a `TypeConsApp<C>`, to the `Decl<C>`.

In the Penrose editor, select the Triangle Incenter example. The updated constructor section in the .domain file will begin like this: 

```
constructor Segment(Point p, Point q)
constructor Ray(Point base, Point direction)
constructor Line(Point p, Point q)
constructor Midpoint(Linelike l) -> Point
```

rather than the original syntax, which was 

```
constructor MkSegment(Point p, Point q)
constructor MkRay(Point base, Point direction)
constructor MkLine(Point p, Point q) -> Line
constructor MkMidpoint(Linelike l) -> Point
```

Note that constructors using the same name as their type can omit the output type. Where constructors do not do so, they must still include the output type. 

In the Substance file, we can note the decl-bind `Let JKL := Triangle(J, K, L)`. This is interchangeable with, yet less redundant than, the original `Triangle JKL := Triangle(J, K, L)` syntax—both of which compile to the same effect. However, the syntax stays the same where the constructor name is different from the output type, such as with `Angle PML := InteriorAngle(P, m, L)`.

# Checklist

# Open questions

Supporting overloaded constructors/functions, if the convention is to move towards constructor names being the names of their types.


wodeni (changes requested):
Great job again on this PR! The implementation looks great and thanks for efforts on refactoring the geometry domain to use this new feature! Requesting changes on two small details on the implementation.

wodeni (approved):

PR #1473, opened 2023-06-08
fix: reuse resolver when example becomes local

### LINKED ISSUE #1463: SVG load fails after edit in web IDE
On the IDE deployed at https://penrose.cs.cmu.edu/try/, SVGs are initially properly loaded for examples.  E.g., the lightbulb in this example:

image

However, if the code is edited and recompiled, the image reference breaks.  E.g., here I just change one of the labels in the Substance program:

image

(Note that the camera is drawn directly in the Style program, rather than imported from SVG, hence doesn't break).

IIRC we had this behavior in the past (and fixed it), but I can't find the associated issue or PR.

My best guess is that it has to do with #1444, since this PR also added back in preview icons.  However, imported SVGs also don't show up properly in these icons—e.g., the green balls here should be shaded:

image

samestep:
# Description

Resolves #1463.

# Checklist

### DIFF
--- a/packages/editor/src/components/DiagramPanel.tsx
+++ b/packages/editor/src/components/DiagramPanel.tsx
@@ -114,7 +114,10 @@ export const pathResolver = async (
     case "gist":
       return undefined;
     case "local": {
-      return fetchResource(relativePath, workspace);
+      const { resolver } = location;
+      return resolver
+        ? resolver(relativePath)
+        : fetchResource(relativePath, workspace);
     }
   }
 };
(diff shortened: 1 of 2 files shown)


wodeni (approved):
LGTM! Thanks for fixing this!

For future reference: the issue is caused by `editor` not caching images in `localforage` at some point in our history. In the past, the `editor` saves fetched images whenever it loads up an example, and these images persist whenever the user starts editing the example (i,e, switching to a "local" workspace). However, this caching behavior was removed due to unknown reasons. 

The reason why it took so long to find: `localforage` can live on your browser for a long time. Many of us probably have browsers that have run a correct version in the past and therefore the cached images. @keenancrane did run this on a newer machine and caught it recently in #1463.

PR #1475, opened 2023-06-09
fix: list whitespace parsing in Style

### LINKED ISSUE #1454: Compilation ultra-slow even when code is commented out (!)
Here's a really messed up example:

A program (attached below) that compiles ultra slow, but if a block of **commented-out code** is deleted, it suddenly compiles really fast.

My guess is that there's some uncaught error in the compilation process.  But it's strange that the commented code has _any_ effect at all on the compilation process.  (Shouldn't it just be preprocessed out of existence?)

This process hints at something possibly deeper broken in the compiler.  The reason I even found this bug is that compilation suddenly became suuuuuupper slow without any major change being made.  Then I tried to comment things out to simplify the code—and noticed it wasn't until I _deleted_ the commented-out code that things became fast again…

[slow-compile-comments.domain.txt](https://github.com/penrose/penrose/files/11652926/slow-compile-comments.domain.txt)
[slow-compile-comments.style.txt](https://github.com/penrose/penrose/files/11652927/slow-compile-comments.style.txt)
[slow-compile-comments.substance.txt](https://github.com/penrose/penrose/files/11652928/slow-compile-comments.substance.txt)

wodeni:
# Description

Resolves #1454.

The `list` rule in the parser handles whitespaces around list elements incorrectly: `expr_list` consumes all the whitespaces already, and the rule ambiguously consumes extra whitespaces, leading to a blowup of ambiguous parses.

Although #1454 observes this only when the comments are present. The number of parsed trees exceeds 4000 with or without them. 

# Checklist

### DIFF
--- a/packages/core/src/parser/Style.ne
+++ b/packages/core/src/parser/Style.ne
@@ -545,7 +545,7 @@ expr_literal
   # |  transformExpr 
 
 list -> "[" expr_list "]" {% 
-  ([lbracket, , exprs, , rbracket]): List<C> => ({
+  ([lbracket, exprs, rbracket]): List<C> => ({
     ...nodeData,
     ...rangeBetween(lbracket, rbracket),
     tag: 'List',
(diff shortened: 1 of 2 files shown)


keenancrane:
Thanks @wodeni!

PR #1482, opened 2023-06-13
fix: type exports in core API

wodeni:
# Description

`api.ts` mixes type and value exports, which causes an error when clients try to import `core` functions:

```
api.ts:66 Uncaught SyntaxError: The requested module '/try/@fs/Users/nimo/Code/penrose/packages/core/src/engine/BBox.ts' does not provide an export named 'BBox' (at api.ts:66:10)
```

This PR separates exported types and values in `api.ts`.


joshsunshine:
I'll merge this to get the site up and running again, but we need to figure out what is going.

PR #1485, opened 2023-06-14
feat: `DraggablePoint` in `solids`

wodeni:
# Description

Resolves #issue.

This PR adds a `DraggablePoint` component to `@penrose/solids`, which is inspired by [Movable points in Mafs](https://mafs.dev/guides/interaction/movable-points). 

* Separate out drag-handling code into `DraggablePoint.tsx` and use it in `eigenvalues.tsx`
* Fix a small build issue with `rose` now that it's introduced by #1636.

# Checklist

# Examples and how to reproduce them

Run `nx run solids:dev` to see the eigenvalue example. All draggable points use `DraggablePoint".

# Open questions

* The API is not the cleanest: `DraggablePoint` will need a ref to the outer SVG to compute the right transforms. I think usually people use contexts to pass these things around. 
* `transform` is technically not needed if `x` and `y` are in the SVG screen space. It's only needed because `DraggablePoint` updates `x` and `y` after dragging. Should we expose them as callbacks instead?


samestep (approved):
LGTM! I didn't review the code in too too much detail though, so let me know if there are any specific parts you want me to take a look at.

Issue #1420, opened 2023-05-23
Restore stroke on Equation


keenancrane:
Stroked equations were added in #1026, but were then reverted in #1208 to make life easier for bundling.

This feature is however pretty critical for making diagram labels legible in a number of scenarios, e.g.,

<img width="842" alt="image" src="https://github.com/penrose/penrose/assets/3604525/f1b5e90f-4d7d-4e44-b669-bbb1c0d7385e">

keenancrane:
Also worth noting that the solution to use Text and Unicode is a poor substitute for the real thing—apart from aesthetic issues, Unicode math covers only a tiny fraction of labels used in practice.

keenancrane:
Perhaps the simplest solution is just to absorb the necessary routines from `svg-flatten` into the renderer. It's pretty short (about 200 lines), and is covered by an ISC license that lets us copy, modify, and redistribute: https://opensource.org/license/isc-license-txt/

samestep:
@keenancrane Sounds reasonable to me! If you wanna open a PR doing that, I'd be happy to review it.

keenancrane:
Some notes about interactions between stroke on `Equation` and font outlining in TeX:

We currently have an experimental path that allows TeX equations to be stroked (even if they are not stroked in SVG) via the field `texContourColor`, whose value is a string providing the name of a LaTeX color (see `RenderEquation()` in `Equation.ts`).

Since this feature is experimental, this field is currently supported via SVG passthrough, i.e., it is not officially part of the `Equation` type definition.  Once proper stroke support is added to `Equation` on the SVG path, we should instead just use the `strokeColor` field for this purpose.  In particular, we will need to:

- Convert the `strokeColor` to a string encoding the color as either RGB values in [0,1] or a hexadecimal value
- Prepend a command `\definecolor{contourColor}{COLOR_FORMAT}{COLOR_STRING}` immediately before the `\contour` command, where `COLOR_FORMAT` and `COLOR_STRING` define the color model and specify the color, as [detailed here](https://www.overleaf.com/learn/latex/Using_colours_in_LaTeX).
- Change the first argument of `\contour` to `{contourColor}`, rather than inserting the value of the `texContourColor` field.

In particular, we should replace this line of the Equation renderer
```
        txt = placeholderString(
          `\\contour{${propVal.contents}}{$${getAdValueAsString(
            shape.string
          )}$}`,
          [x, baselineY],
          shape
        );
```
with something like this
```
        txt = placeholderString(
[...]

PR #1493, opened 2023-06-16
feat: Walk on stars

keenancrane:
image

This PR extends the `walk-on-spheres` example to handle the _walk on stars_ algorithm, prompted by the development of [this tutorial](https://github.com/GeometryCollective/wost-simple), which is illustrated completely in Penrose.  In the process, it adds features and makes fixes needed in order to get these examples working.  Note that **not all diagrams from the tutorial have been added to the Penrose repo**—but are stored in the tutorial repo linked above.  However, the `walk-on-spheres` example has been updated.

Specific changes include:

### Examples

- The structure of the `walk-on-spheres` domain (and examples) has been changed to more explicitly declare points and balls via separate constructors.
- Labeling in the `walk-on-spheres` example is now done in a separate stage.
- Support for walk on stars has been added to the `walk-on-spheres` domain.

### Functions

image

- Added functions `closestSilhouetteDistance` and `firstRayDistance` that give the scalar distance to the first query point (rather than the point itself).  Variants for abstract shapes are also exposed in Style.
- We now correctly implement `disjoint` for `Line-Line` pairs (previously done via AABB approximation).
- We now correctly support `disjoint` for the `Polyline` shape with `Circle` and `Rectlike` (whereas before we just treated a `Polyline` as an AABB).
- We now correctly support padding on `contains` for `Rectangle-Rectangle` pairs (previously padding was ignored).
- Added a `circularArc` convenience function that generates path data for a standard circular arc using just a circle center, circle radius, and start/end angles.  This function properly handles arc orientation and values outside the range $[-\pi,\pi)$.

image

### Shapes

image

- For `Equation` shapes, the passthrough field name `texContourColor` now triggers TeX export to wrap the equation string in a `\contour{color}` command, where the `color` string is given by the value of the `texContourField`.  This behavior affects only downloads from the IDE via the `SVG (TeX)` button and a `roger` invocation with `--tex-labels`.

### Editor

- The editor now prioritizes display of constraint and objective names in the debugging panel.  Previously constraints/objectives were given the same field width as the energy and "is satisfied" fields, which meant they were truncated to a very short string that in most cases did not identify the constaint/objective.
- The number of decimal digits in the energy value has also been truncated (since we don't need ~12 digits of precision for debugging).

### Open issues

- Stubs for two examples (`geometric-queries/signed-distance` and `geometric-queries/disjoint`) were added to provide more thorough testing of geometric queries (see _below, left_ and _below, center_), but significant coverage still needs to be added.  These examples will be built up in later PRs (and are hence not added to the registry).
- The ball shading in the walk on stars examples is not clipped by the domain boundary, since there appears to be a bug with clipping imported images and/or gradients (see image _below, right_).  I will file an issue about this bug; for now the walk on stars examples show the shading in grey and the clipped ball in green, which is reasonable (albeit not ideal) from a stylistic point of view.  (This example is anyway not in the Gallery…)

image


packages/core/src/contrib/Queries.ts
```diff
@@ -281,6 +282,145 @@ export const rectLineDist = (
...
+//
+//   return convexPolygonOriginSignedDistance([
+//     [x0, y0],
+//     [x1, y1],
+//     [x2, y2],
+//     [x3, y3],
+//     [x4, y4],
+//     [x5, y5],
+//   ]);
+// };
```
samestep, line 422: This commented-out code seems like just a duplication of the definition of `rectLineDist` above; could you delete it?

```suggestion
```

keenancrane: Sure thing—not sure what was going on there…

packages/core/src/contrib/Queries.ts
```diff
@@ -368,6 +415,42 @@ export const shapeDistance = (s1: Shape<ad.Num>, s2: Shape<ad.Num>): ad.Num => {
...
+    );
+  } else if (t2 === "Polyline" && t1 === "Circle") {
+    return shapeDistanceCirclePolyline(
+      s1.center.contents,
+      s1.r.contents,
+      s2.points.contents
+    );
+  } else if (t1 === "Polyline" && isRectlike(s2)) {
+    const bbox = bboxFromShape(s2);
+    const corners = BBox.corners(bbox);
```
liangyiliang, line 432: There is a function called `bboxPts(s)` that computes the bounding box of `s` and returns the four points of the bounding box (top-right, top-left, bottom-left, bottom-right). By the way, this function apparently assumes that the rectangles have zero rotation value - maybe that's something that can be written in the documentation?

packages/core/src/renderer/Equation.ts
```diff
@@ -40,11 +40,27 @@ const RenderEquation = (
...
-    const txt = placeholderString(
+
+    let txt = placeholderString(
       `$${getAdValueAsString(shape.string)}$`,
       [x, baselineY],
       shape
     );
+
+    // If the Equation has a texContourColor passthrough value, give the
+    // string a contour with the specified color
```
liangyiliang, line 51: I'm a bit unsure whether `texContourColor` should be a passthrough value, since this field and value appear to be something that Penrose directly handles.

I would suggest perhaps modifying the definition of `Equation` to include this field. [This file](https://github.com/penrose/penrose/blob/6b2550eee5afdcc3e106bdb41e129bead6a51947/packages/core/src/shapes/Equation.ts#L15) is the definition of `Equation`. [Here ](https://github.com/penrose/penrose/blob/6b2550eee5afdcc3e106bdb41e129bead6a51947/packages/core/src/compiler/shapeChecker/CheckShape.ts#LL143C31-L143C31) is the Shape Typechecker which would also need to be adapted to include the new field.

packages/core/src/contrib/Queries.ts
```diff
@@ -281,6 +282,52 @@ export const rectLineDist = (
   ]);
 };
 
+export const polygonSignedDistance = (
```
liangyiliang, line 285: Is this `polygonSignedDistance` function used anywhere? Notice that `Functions.ts` already has `signedDistancePolygon` which apparently does the same thing?

packages/core/src/contrib/Queries.ts
```diff
@@ -409,14 +492,46 @@ export const shapeDistanceRectLine = (
...
+    add(cx, halfW),
+    add(cy, halfH),
+    start[0],
+    start[1],
+    end[0],
+    end[1]
   );
 };
 
+const shapeDistanceRectlikePolyline = (
```
liangyiliang, line 506: For these branches of `shapeDistance`, is it possible to include these into `compDict` as well? For example, `shapeDistanceCircleLine` is in `Queries.ts`, but we also add this into `compDict` so that the Style DSL can call this function.

packages/core/src/contrib/Functions.ts
```diff
@@ -2810,6 +2865,47 @@ export const compDict = {
     },
     returns: valueT("Real2"),
   },
+  rayIntersectDistance: {
```
liangyiliang, line 2868: Is it possible to expose different variants of `rayIntersectDistance` to the Style DSL as well? For example, aside from `rayIntersect`, we also have `rayIntersectCircle`, etc. Same goes for `closestSilhouetteDistance`. Otherwise there is no way for the Style writers to call `rayIntersectDistance` on abstract parameters ...

liangyiliang (commented):
Hi Keenan, thanks for doing this!

My main concerns lie on the consistency of the new functions `rayIntersectDistance` and `closestSilhouetteDistance` with the abstract shape paradigm which we are encouraging. Perhaps expose branches of these functions to the Style DSL, just like how we exposed `rayIntersect` branches to the Style DSL?

I'm also a bit concerned about the treatment of the additional `Equation` field `texContourColor` as passthrough. I think passthrough should only be used to handle properties that Penrose cannot handle. But since we explicitly handle on the value of this field, we should instead make `texContourColor` as a `Equation` shape field so that it can be properly typechecked.

Once again, thanks for doing this!

packages/core/src/contrib/Queries.ts
keenancrane: @liangyiliang The function `bboxPts()` just calls `BBox.corners`, as is done here, right?

```
export const bboxPts = (bbox: BBox.BBox): ad.Pt2[] => {
  const { topLeft, topRight, bottomLeft, bottomRight } = BBox.corners(bbox);
  return [topRight, topLeft, bottomLeft, bottomRight];
};
```

I don't see a version that takes a shape directly as an argument.  Most places it is used like `      bboxPts(bboxFromShape(s))`.  Wouldn't it therefore be better to everywhere write

```
BBox.corners(bboxFromShape( s ))
```

rather than

```
bboxPts(bboxFromShape(s))
```
?

Basically I don't understand the point of the `bboxPts` function (esp. since there are no comments).  Is it just to do a type conversion?

keenancrane: By the way, a lot of these things (like how corners are accessed, whether rect-line functions support rotations, etc.) are not introduced in my PR; I'm just extending to Polylines using existing functions.  I would prefer if we address things like rotations in a future PR, since this one is already quite complicated.

packages/core/src/contrib/Queries.ts
keenancrane: No, good catch.  It seems this was just kept by the automatic merge (Git did not detect that it was renamed by the abstract shape PR).

packages/core/src/contrib/Queries.ts
keenancrane: Sure thing.

packages/core/src/contrib/Functions.ts
keenancrane: Done.

packages/core/src/renderer/Equation.ts
keenancrane: I'm going to leave this one alone, because the whole "export to TeX" thing is experimental at this point, and this is also connected to #1420.  Long term the right solution is just to restory the `strokeColor` field to `Equation`, and then use this field to trigger a `\contour` command on TeX export.  I started going down this road, but it requires quite a few changes to the shape definitions and renderer.  So, better to put that in a different PR than to make this one even more complex.

keenancrane:
@liangyiliang Thanks for the very detailed feedback!  I took care of all your issues except for (i) rectangle rotation, which we should do in a separate PR (there is just too much here already…), and (ii) promoting the LaTeX contour color to an "official" part of the `Equation` shape definition, since this feature is experimental and interacts with #1420.  I did however write a [detailed comment here](https://github.com/penrose/penrose/issues/1420#issuecomment-1595709007) that discusses what the right path forward is, which will anyway replace this experimental field.

PR #1495, opened 2023-06-16
fix: bad clipping on transformed shape

### LINKED ISSUE #1494: bug: Clipping renders wrong for imported images and/or gradients
image

An example is shown above, where the gray shaded balls (which come from an imported SVG) are supposed to be clipped by the grey polygon, but get clipped at the wrong location.  The green balls (which are just `Circle` shapes) are properly clipped, within the same group.

It looks like what may be happening is just that there's a transform that's not properly accounted for, i.e., the clipping path is for some reason translated by some amount before clipping.

Here's the relevant snippet, from `walk-on-spheres.style` (in `examples/`):
```
   shape St.ball = Circle {
      center: p.location
      r: R
      fillColor: rgba( f[0], f[1], f[2], 1 )
      strokeColor: rgba( s[0], s[1], s[2], 1 )
      strokeWidth: Global.sphereStrokeWidth
      ensureOnCanvas: false
      opacity: .35
   }

   shape St.shading = Image {
      href: "ball-shading.svg"
      center: St.ball.center
      width: 2*St.ball.r
      height: 2*St.ball.r
      ensureOnCanvas: false
   }

   layer St.shading above D.geometry
   layer St.ball above St.shading

   shape St.domainGeometry = Polygon {
      points: D.geometry.points
      fillColor: none()
      ensureOnCanvas: false
   }
   -- this group should ideally also include St.shading, but
   -- there appears to be a bug with clipping imported images
   -- (or perhaps gradients)
   shape St.G = Group {
      shapes: [St.ball] -- [St.ball,St.shading]
      clipPath: clip(St.domainGeometry)
      ensureOnCanvas: false
   }
```

The corresponding Substance program is `WalkOnStars.substance`.

liangyiliang:
# Description

Resolves #1494.

This PR fixes a bug where shape clipping fails for the `Image` shape. Through the fix, we realized that this bug occurs whenever a clipped SVG shape has a `transform` field. This PR ensures that clipping on SVG shapes with `transform` field behaves well by enforcing the transformation to be applied before clipping.

Currently, the `Image` shape is rendered as follows:
```svg
<g
    transform="  ... transformation necessary to make sure this image has the right center, width, height ... " 
    clip-path="...">

  <svg> ... imported svg ...</svg>

</g>
```

Apparently, in SVG, `clip-path` is performed on the pre-transformed coordinate, not the post-transformed coordinate.

The fix simply involves wrapping the transformed shape into another `<g>` tag and perform clipping on that outer `<g>` tag.

```svg
<g clip-path = "...">
  <g transform=" ... transformation necessary to make sure this image has the right center, width, height ...">
    <svg> ... imported svg ... </svg>
  </g>
</g>
```

This forces the clipPath to be applied at the post-transformed coordinates. For a StackOverflow post about this, see [this](https://stackoverflow.com/questions/17388689/svg-clippath-and-transformations).

Here is a Style file.
```
canvas {
    height = 700
    width = 800

    r = Rectangle {
        center: (0, 0)
        width: 100
        height: 50
        rotation: 50
    }

    r2 = Rectangle {
        center: (-100, 100)
        width: 150
        height: 150
    }

    g = Group {
        shapes: [r]
        clipPath: clip(r2)
    }
}
```

In the current branch, nothing is shown on the canvas, which is the incorrect behavior. In this branch, a small triangular wedge is shown, which is correct.
# Checklist


wodeni (approved):
Looks great. Thanks for the fix!

keenancrane:
@liangyiliang I updated the `walk-on-stars` example to take advantage of the fix.

keenancrane:
Otherwise, this looks great—thank you!

PR #1498, opened 2023-06-21
feat: function warnings

### LINKED ISSUE #1451: Issue warnings for incomplete/imperfect features
image

Often we implement an approximate algorithm (e.g., in the layout engine) just to provide a baseline level of support.  A good recent example is that `disjoint(Line,Line)` had been implemented via bounding boxes.  Although this version of the constraint formally worked (Lines become disjoint), it also has some unexpected behavior (e.g., Lines have a strong tendency to be aligned with the horizontal or vertical axis).  In turn, Style programs are difficult to develop and debug.

Of course, to make progress on the system we sometimes have to start with an imperfect solution.

A good middle ground, then, would be to issue warnings when approximations are used.  E.g., if `disjoint(Line,Line)` is invoked in a Style program, we could issue a warning like
```
warning (StyleWarning)
Constraint `disjoint` is approximated using bounding boxes for shape types `Line` and `Line` (at line 31, column 9 of Style program)
```

Such warnings can be enabled/disabled via a compiler flag like `--noApproximationWarnings`, which can also be exposed in the **options** panel of the IDE.

### LINKED ISSUE #1489: Style Function Warnings
# Motivation

Many Style functions use approximations such as treating shapes as their bounding boxes. But, these approximations are often not communicated clearly with the users of these functions. This violates the principle of least-surprise.

We should allow Style functions to fire warnings whenever unexpected approximations are used. These warnings can then be shown to the users like in the IDE. We should _not_ restrict to approximation warnings only, as there might be other types of warnings that are appropriate for functions.

# Question: Where Should Warnings Be Issued?

One question is, _where_ should we issue the warnings?

We use the `shapeDistance` function as an example of this question. Assuming PR #1361 is merged, the `shapeDistance` function essentially has three variants:

**Variant 1: Entry in `compDict`**

The `shapeDistance` entry in `compDict` is as follows:
```ts
  shapeDistance: {
    name: "shapeDistance",
    description: "Return the distance between two shapes.",
    params: [
      { name: "s1", type: shapeT("AnyShape"), description: "a shape" },
      { name: "s2", type: shapeT("AnyShape"), description: "a shape" },
    ],
    body: (
      _context: Context,
      s1: Shape<ad.Num>,
      s2: Shape<ad.Num>
    ): FloatV<ad.Num> => floatV(shapeDistance(s1, s2)),
    returns: valueT("Real"),
  },
```
This entry represents a "Style function" that is callable from our Style DSL. Often, this is just a wrapper of the actual implementation function: here, it calls the TypeScript implementation `shapeDistance(s1, s2)` (see Variant 2) and wraps it in `body`.

**Variant 2: TypeScript Implementation**

The TypeScript implementation of the function performs the actual branching based on the shape types. It looks something like this:
```ts
export const shapeDistance = (s1: Shape<ad.Num>, s2: Shape<ad.Num>): ad.Num => {
  const t1 = s1.shapeType;
  const t2 = s2.shapeType;
  if (t1 === "Circle" && t2 === "Circle") {
    return shapeDistanceCircles(...);
  } else if (isPolygonlike(s1) && isPolygonlike(s2)) {
    return shapeDistancePolys(...);
  } else if (isRectlike(s1) && t2 === "Circle") {
    return shapeDistanceRectCircle(...);
  } 
  // ...
  else {
    // bounding box approximation here
    return shapeDistanceRects(
      bboxPts(bboxFromShape(s1)),
      bboxPts(bboxFromShape(s2))
    );
  }
};
```

Each branch calls a version of the `shapeDistance` function using abstract shape parameters (see Variant 3). The last branch assumes the bounding-box approximation.

**Variant 3: Abstract Parameters**

The final variant of the `shapeDistance` function are its smaller, abstract-parameter branches like `shapeDistanceCircles`, `shapeDistancePolys`, and `shapeDistanceRects`. These functions are implemented separately in TypeScript, but are also are also exposed in the Style DSL in `compDict`.

These functions themselves do not perform any approximations that would surprise the users. For example, to call `shapeDistanceRects` on bounding boxes of shapes, the user would have to manually compute the bounding box coordinates using the `bboxPts` Style function before sending the bounding box coordinates into `shapeDistanceRects`. In other words,the users would have to explicitly write,
```
shapeDistanceRects(bboxPts(s1), bboxPts(s2))
```
and hence will not be surprised at all when the expression actually calculates the distance between the bounding box of `s1` and the bounding box of `s2`, not the actual distance between `s1` and `s2`.

**The _Question_**

Where should we fire the approximation warnings?

First, "Variant 3" functions of `shapeDistance` won't need to issue any approximation warnings, since these functions make no approximations except those explicitly declared by the users.

If "Variant 2" of `shapeDistance` is called only through the wrapper "Variant 1", then we can put the warnings in either variants.

But the whole reason why we splitted the "Variant 2" out of "Variant 1" is because we _might_ want to call `shapeDistance` directly from the Penrose API without indexing into `compDict`. If that is the case, since approximation warnings are Style-level warnings, we should only issue the warnings in "Variant 1", the variant that the Style DSL uses. Otherwise, API calls to `shapeDistance` would also receive Style-level warnings which violates the whole goal of having an API.

liangyiliang:
# Description

Resolves #1451.
Resolves #1489.

We add a new type of warning (`FunctionInternalWarning`) that can be triggered within Style functions. This type of warning currently has one variant, called `BBoxApproximationWarning`, which is particularly used when some Style functions approximates some shapes using their bounding boxes. Unlike function errors which interrupt the entire execution of the functions, warnings are to be recorded, but should not interrupt the functions.

# Style Function Library Changes

## General Design for All Function Warnings

Currently, we only have one type of `FunctionInternalWarning`, but this section applies even with multiple types of warnings.

We considered two solutions:
1. Each Style function should return both a value (`ad.Num` or `Value<ad.Num>`) and a list of warnings (empty if no warning is fired). This would require changes to every single Style function.
2. Each Style function should have two "sub-functions" - one `body` function that actually computes the return value, and one `warning` function that just computes the list of warnings. By default the `warning` function returns empty list.

Originally we thought that Solution 2 would yield better result since it does not require changes to any functions that do not issue warnings, since we provide a default `warning` function. But upon experimenting with Solution 2, we realized that in many functions, the decision of whether or not to issue a warning are intertwined with a long chain of `if ... else if ... else if ... ... ... else` statements in the `body` function. The `warning` function, then, needs to replicate all these logic, which yields very ugly code.

So we eventually opted with Solution 1: each function return both a value and a list of warnings.

We define type `MayWarn<T>` that encodes both a value and a list of warnings:
```ts
type MayWarn<T> = {
  value: T,
  warnings: FunctionInternalWarning[]
}
```
Then, Solution 1 mandates that a computation function returns `MayWarn<Value<ad.Num>>` while an objective / constraint function returns `MayWarn<ad.Num>`.

For the vast majority of functions that do not return warnings, we defined two convenience functions:
* `noWarn(x)` takes `x` and wraps into `MayWarn` with an empty `warnings` list. A function that returns value `v` and does not issue warnings can then just return `noWarn(v)`.
* `noWarnFn(f)` takes `f` (which has type `Input -> Output`) and wraps it into a function with type `Input -> MayWarn<Output>` with an empty `warnings` list. This is used to wrap an entire function, not a single return value.

As an example, the `acos` computation function becomes:
```ts
  acos: {
    name: "acos",
    description: "Return `acos(x)`.",
    params: [{ name: "x", description: "`x`", type: realT() }],
    body: (_context: Context, x: ad.Num): MayWarn<FloatV<ad.Num>> => {
      return noWarn({
        tag: "FloatV",
        contents: acos(x),
      });
    },
    returns: valueT("Real"),
  },
```
Notice that the `body` function now returns `noWarn(...)`. The `equal` constraint function becomes:
```ts
  equal: {
    name: "equal",
    params: [
      { name: "x", description: "First value", type: realT() },
      { name: "y", description: "Second value", type: realT() },
    ],
    body: noWarnFn(equal),
  },
```
Notice that we use `noWarnFn` to wrap around the implementation function `equal`. Originally, function `equal` has type `(x: ad.Num, y: ad.Num) => ad.Num`; now, `noWarnFn(equal)` has type `(x: ad.Num, y: ad.Num) => MayWarn<ad.Num>`.

## Bounding Box Approximation Warning Specifics

In many other programming languages, errors and warnings maintain a `trace` of what happened. Here, we also maintain a trace for `BBoxApproximationWarning`, because many functions call other library functions which, in turn, issue `BBoxApproximationWarning`. For example, here is how the `disjoint` library function works:
* `disjoint` calls `overlapping`
* `overlapping` calls `shapeDistance`
* `shapeDistance` may use bounding-box approximations in some cases and issue warning.

The warnings issued by `shapeDistance` should propagate to `disjoint` using a trace.

The `BBoxApproximationWarning` is defined as follows:
```ts
export interface BBoxApproximationWarning {
  tag: "BBoxApproximationWarning";
  // tail is the top of stack
  stack: [BBoxApproximationWarningItem, ...BBoxApproximationWarningItem[]];
}

export interface BBoxApproximationWarningItem {
  signature: string;
  location?: SourceRange;
}
```

A few notes about this definition:
* `BBoxApproximationWarning` essentially stores a "stack" of function calls (each represented by a `BBoxApproximationWarningItem` object).
* Each `BBoxApproximationWarningItem` object stores the `signature` of the function call and the optional source location.
  * For example, if `shapeDistance(Circle, Polygon)` issues a warning, then `signature` would be the string `shapeDistance(Circle, Polygon)`.
  * The source location is optional for two reasons.
    * During the execution of the function, we do not have access to the source location of the function call. Only after the termination of the function can we "attach" a source location.
    * If `overlapping` calls `shapeDistance` which then issues a warning, the call to `shapeDistance` does _not_ have a source location, since it is done within the TypeScript implementation of `overlapping`.

As an example, suppose we call `overlapping(Circle, Polygon)`.
* Suppose `overlapping(Circle, Polygon)` calls `shapeDistance(Circle, Polygon)`.
* Suppose `shapeDistance(Circle, Polygon)` issues a `BBoxApproximationWarning`. This warning would have a stack of one item, consisting of the signature `shapeDistance(Circle, Polygon)`.
* The value of `shapeDistance(Circle, Polygon)` and the warning are both passed back to `overlapping(Circle, Polygon)`.
* Function `overlapping(Circle, Polygon)` now knows that `shapeDistance(Circle, Polygon)` issued a `BBoxApproximationWarning`. So `overlapping(Circle, Polygon)` pushes its own signature `overlapping(Circle, Polygon)` onto the stack, and returns the new warning alongside the correct `overlapping` value.

Now, `overlapping(Circle, Polygon)` returns a `BBoxApproximtionWarning` whose stack consists of two items: one by `shapeDistance(Circle, Polygon)` and one by `overlapping(Circle, Polygon)`. The Style compiler would then attach a location information to top of the stack, `overlapping(Circle, Polygon)`.

In the end, the warning stack is pretty-printed out as follows:
image

# IDE and Trio Changes

Issue #1451 also discusses about the potential of disabling certain warnings. We expose this feature in the IDE and the Trio JSON files.

In the IDE "options" panel, we add a selector for the types of warnings to be disabled.

options panel with selectors for which warnings to disable

Note that we support multiple-selection. The list of options are exported from `@penrose/core`.

The IDE collects the selected warning types into a string of arrays, and sends this list to the compiler. Near the end of the compilation process, the compiler goes through the list of warnings and removes those warnings that appear in the selected list.

Aside from IDE, the trios have also been adapted to support an _optional_ list of warnings to be discarded. (If this list is not provided, by default, all warnings are preserved). For example, the `Two Perpendicular Vectors` example now has such a field:
```json
{
  "substance": "./twoVectorsPerp-unsugared.substance",
  "style": ["./linearalgebra-paper-dashes.style"],
  "domain": "./linear-algebra.domain",
  "variation": "",
  "excludeWarnings": ["BBoxApproximationWarning"]
}
```
## Change to `roger`
`roger` now prints (on `console.warn`) all warnings that are reported by the compiler (except the excluded ones). It also takes into account the `excludeWarnings` field of `trio` files, passing them into the compiler or the IDE as necessary.

# Checklist

# Open questions

As of now, we are _not_ supporting saving the `excludeWarnings` field in the local and Gist workspaces. This is the broader issue of diagram-related elements like `variation` not being saved to local and Gist workspaces. We would like to, generally, support saving of these fields into local and Gist workspaces.


packages/roger/watch.ts
```diff
@@ -68,7 +68,12 @@ export default async function (port = 9160): Promise<void> {
           // get containing dir of the trio file
           const parentDir = parse(path).dir;
           const contents = await fs.readFile(path, "utf8");
-          const { substance, style, domain } = JSON.parse(contents);
+          const pathContents = JSON.parse(contents);
+          const { style, domain, substance } = pathContents;
+          let { excludeWarnings } = pathContents;
+          if (excludeWarnings === undefined) {
+            excludeWarnings = [];
+          }
```
wodeni, line 76:
```suggestion
          const { style, domain, substance, excludeWarnings } = JSON.parse(contents);        
```

For brevity mostly, also see below

packages/roger/watch.ts
```diff
@@ -104,6 +109,7 @@ export default async function (port = 9160): Promise<void> {
                           : relative(".", resolve(parentDir, style[0])), // HACK: use the first style path as the path to the combined Style because the combined Style doesn't exist in the file system
                     },
                   },
+                  excludeWarnings,
```
wodeni, line 112:
```suggestion
                  excludeWarnings ?? [],
```

packages/core/src/contrib/Constraints.ts
```diff
@@ -184,41 +193,69 @@ export const overlapping = (
...
+    return {
+      value: add(dist, overlap),
+      warnings: warnings.map((warning) => {
+        if (warning.tag === "BBoxApproximationWarning") {
+          return {
+            ...warning,
+            stack: [
+              ...warning.stack,
+              {
+                signature: `overlapping(${t1}, ${t2})`,
```
wodeni, line 249: The location is not added because we don't have access to the source range here, and the location will be populated by the `call*` functions later?

Probably good to leave a comment somewhere if that's the case. Maybe in the `call*` functions

packages/core/src/contrib/Constraints.ts
```diff
@@ -468,18 +569,68 @@ export const containsGroupShape = (
...
     // (2) the clipping shape contains the other shape.
-    return andConstraint(
-      minN(vals),
-      constrDict.contains.body(clip.contents, s2)
-    );
+    const clipResult = contains(clip.contents, s2);
+    const clipEnergy = clipResult.value;
+    const clipWarnings = clipResult.warnings.map((warning) => {
+      if (warning.tag === "BBoxApproximationWarning") {
+        const newStack: [
```
wodeni, line 611: The type is just `BBoxApproximationWarningItem[]` here?

packages/core/src/types/errors.ts
```diff
@@ -255,6 +258,17 @@ export interface GroupCycleWarning {
   cycles: string[][];
 }
 
+export interface BBoxApproximationWarning {
+  tag: "BBoxApproximationWarning";
+  // tail is the top of stack
+  stack: [BBoxApproximationWarningItem, ...BBoxApproximationWarningItem[]];
```
wodeni, line 264: Similarly, what's the advantage of this type here (vs. `BBoxApproximationWarningItem[]`)?

packages/core/src/utils/Util.ts
```diff
@@ -1021,3 +1023,31 @@ export const getValueAsShapeList = <T>(val: Value<T>): Shape<T>[] => {
...
+  f: (...args: T) => S
+): ((...args: T) => MayWarn<S>) => {
+  return (...args: T) => noWarn(f(...args));
+};
+
+//#endregion
+
+//#region warnings
+
+export const allWarnings: Warning["tag"][] = [
```
wodeni, line 1044: Possible hack to avoid this list: https://stackoverflow.com/questions/44480644/string-union-to-string-array?

wodeni (changes requested):
Looks good overall! Thanks for all the hard work on this PR. Though we are not issuing a lot of warnings in functions now, having this is very useful. 

See comments for some mostly cosmetic suggestions.

packages/core/src/contrib/Constraints.ts
liangyiliang: I'd like to make this more general. The idea is, a function warning might eventually have multiple variants even though we currently only have one. I'm putting this check here for this purpose.

packages/core/src/types/errors.ts
liangyiliang: This requires the `BBoxApproximationWarning` to have at least one item in the stack.

packages/core/src/contrib/Constraints.ts
liangyiliang: No, the type is `BBoxApproximationWarning`. This type is a variant of `FunctionWarning` which is a variant of `StyleWarning`. `BBoxApproximationWarningItem` is a single entry in the stack of `BBoxApproximationWarning`.

packages/roger/watch.ts
liangyiliang: done

packages/roger/watch.ts
liangyiliang: This works: `excludeWarnings: excludeWarnings ?? []`.
Done

packages/core/src/contrib/Constraints.ts
liangyiliang: Done

packages/core/src/utils/Util.ts
liangyiliang: Did this instead:
```ts
export const allWarnings = [
  "BBoxApproximationWarning",
  "GroupCycleWarning",
  "ImplicitOverrideWarning",
  "LayerCycleWarning",
  "NoopDeleteWarning",
  "ShapeBelongsToMultipleGroups",
] as const;

// These are type-level assertions that allWarnings
// covers each variant in StyleWarning
type AssertedWarningTags = (typeof allWarnings)[number];
type ActualWarningTags = StyleWarning["tag"];

type IsSubset<T, U> = T extends U ? true : false;
type AreUnionsEqual<T, U> = IsSubset<T, U> extends true
  ? IsSubset<U, T>
  : false;

// If this fails, then allWarnings and the actual tags of StyleWarning variants
// are different.
const _warningTagsCheck: AreUnionsEqual<
  AssertedWarningTags,
  ActualWarningTags
> = true;

```

wodeni (approved):
LGTM!

PR #1500, opened 2023-06-22
chore: Remove old assets

keenancrane:
This PR removes a collection of SVG files from `core/assets` that were used in the examples from our SIGGRAPH paper.  These examples no longer exist, and the assets are unused in current examples.  (If for some strange reason we need to reconstruct these examples, the assets are still available in the Git history, and in a pinch can even be extracted from the archival PDF hosted by ACM.)


samestep (approved):

joshsunshine:
Thanks for doing this cleanup task!

PR #1503, opened 2023-06-22
Additional curve support

keenancrane:
# Description

Resolves #issue.

# Checklist

# Open questions

### DIFF
--- a/packages/core/src/contrib/Functions.ts
+++ b/packages/core/src/contrib/Functions.ts
@@ -1087,11 +1087,13 @@ export const compDict = {
 
        const path = new PathBuilder();
        path.moveTo( points[0] );
-       for( let i = 0; i < n-1; i++ ) {
+       const m = (pathType === "open" ? n-1 : n);
+       for( let i = 0; i < m; i++ ) {
+          const j = (i+1) % n;
           path.bezierCurveTo(
              ops.vadd( points[i], tangents[i] ),
-             ops.vsub( points[i+1], tangents[i+1] ),
-             points[i+1]
+             ops.vsub( points[j], tangents[j] ),
+             points[j]
           );
        }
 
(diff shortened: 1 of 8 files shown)


packages/core/src/contrib/Functions.ts
```diff
@@ -1008,6 +1008,115 @@ export const compDict = {
...
+    body: (_context: Context, points: ad.Num[][]): VectorV<ad.Num> => {
+      return {
+        tag: "VectorV",
+        contents: points[points.length - 1],
+      };
+    },
+    returns: valueT("Real2"),
+  },
+
+  averagePoint: {
```
liangyiliang, line 1041: We already have `sumVectors` defined as follows:
```
  sumVectors: {
    name: "sumVectors",
    description: "Return the sum of vectors in a list of vectors.",
    params: [{ name: "vecs", description: "vectors", type: realNMT() }],
    body: (_context: Context, vecs: ad.Num[][]): MayWarn<VectorV<ad.Num>> => {
      if (vecs.length === 0) {
        throw new Error("Expect a non-empty list of vectors");
      }
      const vlen = vecs[0].length;
      const zeros: ad.Num[] = new Array(vlen).fill(0);
      return noWarn(
        vectorV(vecs.reduce((curr, v) => ops.vadd(curr, v), zeros))
      );
    },
    returns: realNT(),
  },
```
Is the `averagePoint` function necessary here?

liangyiliang (commented):
Looks great! Thanks!

packages/core/src/contrib/Functions.ts
keenancrane: Yes—see my description in the PR. ;-)

keenancrane: Yes—see my description in the PR. ;-)

liangyiliang: Sounds good!

liangyiliang (approved):

wodeni (approved):
Love it. Thanks, @keenancrane!

keenancrane:
Merged—thanks all!

PR #1513, opened 2023-06-27
feat: error and warning markings in IDE editor

liangyiliang:
# Description

Since many types of warnings and errors include source location information, we can utilize these information to mark them in the IDE editor.

1. We write a function, `errLocs`, that extracts the location information (which file, which row, which column) from each error and warning, _except `StyleErrorList`_. `StyleErrorList` essentially stores multiple Style errors, so we handle them specially.
2. For errors and warnings with location information, the IDE editor now draws red squiggly lines at places for errors, and yellow squiggly lines at places for warnings.
3. These squiggly lines update after every time the "compile" button is pressed.

# Checklist

### DIFF
--- a/packages/core/src/utils/Error.ts
+++ b/packages/core/src/utils/Error.ts
@@ -624,7 +624,6 @@ export const errLocs = (
       return [];
     }
     case "ParseError":
-      // Don't know in Style, Substance, or Domain
       if (e.fileType === undefined || e.location === undefined) {
         return [];
       } else {
--- a/packages/core/src/utils/Error.ts
+++ b/packages/core/src/utils/Error.ts
@@ -31,6 +31,7 @@ import {
   RuntimeError,
   SelectorFieldNotSupported,
   StyleError,
+  StyleErrorList,
   StyleWarning,
   SubstanceError,
   SymmetricArgLengthMismatch,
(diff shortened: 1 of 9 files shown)


packages/core/src/utils/Error.ts
```diff
@@ -608,6 +612,246 @@ canvas {
   }
 };
 
+export const errLocs = (
```
wodeni, line 615: Another way would be to add this to an umbrella type so every error is required to identify the offending source location. This is also fine, but the downside is yet another place to update when a new error is added.

packages/core/src/utils/Util.ts
```diff
@@ -1021,3 +1022,32 @@ export const getValueAsShapeList = <T>(val: Value<T>): Shape<T>[] => {
...
 //#endregion
+
+//#region AST
+
+export type ErrorLoc = {
+  type: NodeType;
+  range: SourceRange;
+};
+
+export const toErrorLoc = (node: {
```
wodeni, line 1033: Is this function only renaming and grouping some values? Looks like you had to do some work when feeding them to monaco anyway. Why is this step necessary?

wodeni (approved):
Looks good! Thanks for getting to this issue finally. Testing this on the preview deployment felt great 💯

PR #1515, opened 2023-06-28
docs: homepage tweaks

wodeni:
# Description

This PR makes a few changes to our homepage:

* Add discord server invitation
* Tweak landing page design
  * Better `Demo` component positioning and sizing
  * More demo examples: hexagonal graph, caffeine, 3D euler diagram
* Update footer to include copyright and license info   
* Gallery page: rounded corners and dark-mode shadow

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have reviewed any generated registry diagram changes

PR #1518, opened 2023-06-29
feat: cropping picks smaller viewbox in gallery and `editor`

rjainrjain:
# Description

This PR resolves the problem of diagram previews being cropped to the tightest bounding box even when the intended view box is zoomed in on a portion of the total shapes, which makes some previews look wholly unlike the actual diagrams in the gallery and the editor.

We introduce a check to compare the areas of (1) the existing view box and (2) the tightest bounding box of all shapes, and set the SVG `viewBox` to the smaller of the two.

Since this is done after the SVGs have been produced, we parse the widths and heights of the `viewBox` and `croppedViewBox` strings from these SVGs to calculate the area. We only set the `viewBox` to the values in `croppedViewBox` if the area of the latter is smaller than the former. 

Here is a screenshot of the previously badly cropped diagram in the gallery (Brownian Motion Absorbed at Boundary), which is now fixed:
Screen Shot 2023-06-29 at 10 34 54 AM
And here it is in the editor:
Screen Shot 2023-06-29 at 10 35 11 AM

# Checklist

### DIFF
--- a/packages/components/src/Gallery.tsx
+++ b/packages/components/src/Gallery.tsx
@@ -23,15 +23,10 @@ const Container = styled.div`
   }
 `;
 
-const Example = ({
-  example,
-  ideLink,
-}: {
-  ideLink: string;
-  example: TrioWithPreview;
-}) => {
-  // determine whether to crop the SVG
-  const svgDoc = parser.parseFromString(example.preview!, "image/svg+xml");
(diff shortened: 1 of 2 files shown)


wodeni (approved):
Looks great! BTW I'm removing some redundant calls to `cropSVG` in #1515 soon.

PR #1527, opened 2023-07-06
refactor!: more readable `core` language API

wodeni:
# Description

This PR is a prereq for #1525

This PR improves the `core` "language API" (i.e. any `@penrose/core` functions for interacting with Penrose DSLs programmatically). Here are the main breaking changes:

* Rename `compileTrio` to `compile`.
* Delete `prepareState`, which effectively does text measurement, and move its functionality into `compile`.
* Rename `stepUntilConvergence` to `optimize`.
* Change the return type of `step` and `stepNextStage` to `Result<State, PenroseError>`.
* Rename `stepState` to `stepTimes` and change its return type to `Result<State, PenroseError>` too.
* Rename `stateConverged` to `isOptimized`.

* Some non-breaking changes
  * Add `isError(State) -> boolean` helper function
  * Add a lower level `step` function that takes a `State` and an early-exit condition `until: () => boolean`. This function is a wrapper of `step_until` of `@penrose/optimizer` that wraps optimizer errors with a `PenroseError`.
* Since label collection is a part of compilation now, we don't report the labeling time in the performance graph

# Checklist


packages/core/src/index.ts
```diff
@@ -44,37 +40,60 @@ export const resample = (state: State): State => {
   });
 };
 
-const step = (state: State, numSteps: number): State => {
+const step = (
+  state: State,
+  until: () => boolean
```
samestep, line 45: Here `until` is a positional parameter rather than named; in contrast, in `ad.ts` I have the user pass `until` inside of an object:

https://github.com/penrose/penrose/blob/b3bbd666e70128601751517916fb909f58e73e0b/packages/core/src/types/ad.ts#L248-L251

This way, they have to write the word `until`, so in the code it's hopefully more clear that the function returning `true` is what's supposed to stop the stepping.

Could we do a similar thing here?

samestep (commented):
Looks great to me, thanks so much for doing this refactor! Just had one suggestion.

samestep (approved):

PR #1534, opened 2023-07-12
docs: revise constr/obj doc page

### LINKED ISSUE #1488: Outdated "Writing Constraints & Objectives" Docs-Page
This [documentation page](https://penrose.cs.cmu.edu/docs/ref/constraints), which teaches contributers how to write constraints and objectives, has the following problems.

* It still uses shape tuples (`[shapeType, props]`) which haven't existed since the consolidation of shape types.
* It does not use function-as-data.
* We removed the example about `minSize` and `maxSize` because `shapeSize` does not behave well. We should add a better example.

samestep:
# Description

Resolves #1488.

# Checklist

### DIFF
--- a/packages/docs-site/docs/ref/constraints.md
+++ b/packages/docs-site/docs/ref/constraints.md
@@ -1,12 +1,6 @@
 # Writing Constraints & Objectives
 
-## Introduction
-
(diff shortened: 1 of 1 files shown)


wodeni (approved):
Looks great! Thanks for updating this page 🚀

PR #1552, opened 2023-07-16
docs: improve blog excerpt styling

wodeni:
# Description

For some reason tailwind's dark mode doesn't play well with our site. This PR makes sure we vitepress instead of tailwind for dark/light mode styling. We also change from `home` to `page` layout so the left bar is always visible.


samestep (approved):
Looks good to me, thanks!

PR #1560, opened 2023-07-17
fix: layering on nonexistent objects

### LINKED ISSUE #1557: Uncaught error when `layer` directives refer to nonexistent shapes
**Describe the bug**

When the `layer` directives refer to nonexistent shapes, the compiler throws an unhandled error.

**To Reproduce**

```
canvas {
    height = 800
    width = 700

    layer a above b
}
```

image

**Expected behavior**
It should yield an error visible to the IDE denoting that, for example, `a` and `b` don't exist.

liangyiliang:
# Description

Resolves #1557 .

This PR fixes the bug of uncaught error when a `layer` directive is applied to nonexistent objects. As an example,
```
canvas {
    height = 800
    width = 700

    layer a above b
}
```
The compiler doesn't check that `a` and `b` actually exist, nor does it confirm that indeed, `a` and `b` are shapes.

When translating `layer` statements, the compiler now resolves each expression and ensures that each object exists and are of `Shape` type. If not, it reports an error.

The Style code used in the example now properly reports an error when necessary.

# Checklist

### DIFF
--- a/packages/core/src/compiler/Style.test.ts
+++ b/packages/core/src/compiler/Style.test.ts
@@ -927,6 +927,9 @@ delete x.z.p }`,
            x.z = 1.0
            x.y = x.z.p
 }`,
+        `forall Set x {
+          layer AAA above BBB
+        }`,
       ],
       CanvasNonexistentDimsError: [
         `foo {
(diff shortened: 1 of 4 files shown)


wodeni (approved):
Thanks for catching this bug and fixing it quickly!

PR #1562, opened 2023-07-17
build: compile `@penrose/roger` for distribution

wodeni:
# Description

As reported by a community member, running `roger` when it's already a local dependency results in a `ts-node` error about not finding `tsconfig.json`. While the exact cause is unclear, this PR compiles `@penrose/roger` before packing it for npm publish. 

* Run plain `node` in `run.js` binary
* Remove `ts-node` from `roger`'s dependencies
* 

Run `npm unlink @penrose/roger` if you have it installed locally. Run `npm link` under `packages/roger` to re-link it. The behavior of the binary should stay the same.

# Checklist

# Open questions

Should we include the source maps in the package?

### DIFF
--- a/packages/roger/package.json
+++ b/packages/roger/package.json
@@ -3,13 +3,16 @@
   "version": "3.0.0",
   "description": "",
   "type": "module",
-  "main": "index.tsx",
+  "main": "dist/index.js",
   "scripts": {
-    "build": ":",
+    "build": "tsc",
     "build-decls": ":",
     "start": "ts-node --esm --swc ./index.ts",
     "typecheck": "tsc"
   },
+  "files": [
+    "dist/**/*.js"
+  ],
   "bin": {
     "roger": "./bin/run.js"
   },
@@ -37,7 +40,6 @@
     "prettier": "2.8.8",
     "regenerator-runtime": "^0.13.11",
     "true-myth": "^4.1.1",
-    "ts-node": "^10.9.1",
     "ws": "^8.6.0",
     "yargs": "^17.7.2"
   },
--- a/packages/roger/bin/run.js
+++ b/packages/roger/bin/run.js
@@ -1,3 +1,2 @@
-#!/usr/bin/env -S npx ts-node --esm --swc
-
-import "../index.js";
+#!/usr/bin/env node
+import "../dist/index.js";
--- a/packages/roger/tsconfig.json
+++ b/packages/roger/tsconfig.json
@@ -1,13 +1,11 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "noEmit": true,
+    "outDir": "dist",
+    "sourceMap": true,
     "noImplicitThis": true,
     "resolveJsonModule": true,
     "strictNullChecks": true,
     "strict": false // COMBAK: strict disabled for now due to errors
-  },
-  "ts-node": {
-    "swc": true
   }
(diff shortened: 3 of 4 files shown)


packages/roger/package.json
```diff
@@ -3,13 +3,15 @@
...
-    "build": ":",
+    "build": "tsc",
     "build-decls": ":",
-    "start": "ts-node --esm --swc ./index.ts",
-    "typecheck": "tsc"
+    "typecheck": ":"
   },
+  "files": [
+    "dist/**/*.js"
+  ],
```
samestep, line 14: Does this exclude `bin/run.js`?

wodeni: Not super sure tbh

jeremyong:
> Should we include the source maps in the package?

This is an interesting question. My understanding is that most packages don't do this, because the source maps will increase the package size by roughly a factor of 2 (give or take obviously). Personally? I think disk space is a small price to pay for improved debuggability, and letting users step through code in a sensible manner (when I distribute C++ libraries, I always ensure the PDBs are available for example, even though the PDBs are significantly larger than the binaries themselves). This is, admittedly, a personal preference though.

One thing that occurred to me however, is whether you'd want `@penrose/roger` to be packaged for consumption both as a CLI tool *and* as a library. If consumed as a library, you'd want the TS files for all the type information to be available at edit time. As a CLI tool, you'd want to strip the TS dependency altogether as done here.

wodeni:
> I think disk space is a small price to pay for improved debuggability, and letting users step through code in a sensible manner 

Thanks for the suggestions @jeremyong! Given that the unpacked size is 19.5kB without and 35kB with source maps. I'll include the source maps for now. We can always remove them if it affects DX in the future.

> One thing that occurred to me however, is whether you'd want `@penrose/roger` to be packaged for consumption both as a CLI tool _and_ as a library.

Right now `@penrose/roger` doesn't really export anything as it's a simple wrapper around `core`, so I'm distributing the `.js` files for the next release.

packages/roger/package.json
wodeni: including `bin/run.js` and tested with `npm pack`:

```
npm notice
npm notice 📦  @penrose/roger@3.0.0
npm notice === Tarball Contents ===
npm notice 1.6kB  README.md
npm notice 47B    bin/run.js
npm notice 10.4kB dist/index.js
npm notice 9.8kB  dist/index.js.map
npm notice 44B    dist/types.js
npm notice 98B    dist/types.js.map
npm notice 6.4kB  dist/watch.js
npm notice 5.6kB  dist/watch.js.map
npm notice 1.0kB  package.json
npm notice === Tarball Details ===
npm notice name:          @penrose/roger
npm notice version:       3.0.0
npm notice filename:      penrose-roger-3.0.0.tgz
npm notice package size:  9.2 kB
npm notice unpacked size: 35.0 kB
npm notice shasum:        ba19590180cfd42aa5eb365f3db56c8f23121bc0
npm notice integrity:     sha512-M8+/vRrSM5RJv[...]FJIb65Cc3Nk3A==
npm notice total files:   9
npm notice
penrose-roger-3.0.0.tgz
```

samestep:
Somehow I missed this discussion about source maps; sorry! I think that right now we're distributing the source maps but not the actual source? Should we add `index.ts` to the list of included files in `packages/roger/package.json`?

PR #1570, opened 2023-07-19
docs: fix anchors for functions

logan12358:
# Description

On the "Style Functions" reference page the function heading elements don't have unique IDs, so they can't be permalinked to and the "On this page" links on the right hand side don't work.

This change sets a unique ID on each of the headings, so each function can be linked to individually.

I'm not super familiar with `markdown-it`, but I couldn't find a nice way to get the ID from markdown `###` syntax, so we just write a `<h3>` element manually.

There are these issues in VitePress, but I don't think they handle this yet. vuejs/vitepress#785 and vuejs/vitepress#960. Maybe in the future this commit can be reverted.

# Checklist

# Open questions

This change might break if we can't assume that:
a. Function names will be valid in a URL
b. Function names are unique within each section (which they seem to be at the moment)

### DIFF
--- a/packages/docs-site/docs/ref/style/functions.md
+++ b/packages/docs-site/docs/ref/style/functions.md
@@ -10,7 +10,10 @@ const { objDict, compDict, constrDict } = data;
 
 <div v-for="f in constrDict">
 
-### {{ f.name }}
+<h3 :id="`constraint-${f.name}`">
+  {{ f.name }}
+  <a class="header-anchor" :href="`#constraint-${f.name}`" :aria-label="`Permalink to constraint &quot;${f.name}&quot;`">&ZeroWidthSpace;</a>
+</h3>
 
 <Function :name="f.name" :description="f.description" :params="f.params" :returns="f.returns" />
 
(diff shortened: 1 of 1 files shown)


joshsunshine:
Thank you @logan12358!

wodeni:
This is great. Thanks so much @logan12358! Can't tell you how many times I tried to fix this with `markdown-it` and failed. Makes much more sense to just write custom HTML for this. One suggestion: can we add anchor links to the headers for each functions like so:

```vue
<h3 :id="`constraint-${f.name}`">
  {{ f.name }}
  <a :href="`#constraint-${f.name}`">#</a>
</h3>
```

It's probably useful for sharing links to functions a little quicker.

wodeni (approved):

logan12358:
Good point, looking again it seems like [`markdown-it-anchor`](https://www.npmjs.com/package/markdown-it-anchor#permalinks) was adding that anchor link for markdown headings - I've just manually copied the `<a>` tag it was generating.

wodeni:
> Good point, looking again it seems like [`markdown-it-anchor`](https://www.npmjs.com/package/markdown-it-anchor#permalinks) was adding that anchor link for markdown headings - I've just manually copied the `<a>` tag it was generating.

Just checked it out. Looks awesome! Thanks again for fixing this @logan12358. Merging now.

keenancrane:
Also wanted to chime in (even if a bit late…) and say "thank you!" for figuring this out, @logan12358!  As a heavy Style user, this has really improved my day-to-day. 😄

Issue #1516, opened 2023-06-29
bug: Arithmetic operations `+` and `-` in Style require spaces between constant operands


keenancrane:
The Style compiler currently reports an error when two numerical constants are added or subtracted without putting spaces around the operator, resulting in an error of the form

```
Unexpected float_literal token: "+5". Instead, I was expecting to see one of the following:

    ws token
    "*"
    "/"
    ".*"
    "./"
    "^"
```

This error does not occur for the operands `*` and `/`, nor if the operands are paths rather than constants.  Here's a full example:

```
canvas {
    width = 100
    height = 100

   scalar a = 5
   scalar b = 5

    shape C = Circle {
        center: (0,0)
        -- r: 5+5 -- causes an error
        -- r: 5-5 -- causes an error
        -- r: 5*5 -- works fine
        -- r: 5/5 -- works fine
        -- r: a+b -- works fine
        -- r: a-b -- works fine
    }
}
```

The issue is probably that we are parsing, e.g., `5-5` as a token `5` next to a token `-5`, rather than as a binary operation `-` on `5` and `5` (and likewise for `+`).  So, the fix may simply be to change the order in which unary and binary `-` and `+` get parsed.

keenancrane:
Simply reordering the unary ops and arithmetic ops in `Style.ne` doesn't resolve the issue.

liangyiliang:
It appears that the issue also occurs in Substance indexed sets: see [discussion](https://github.com/penrose/penrose/pull/1572#issuecomment-1662379522).

PR #1572, opened 2023-07-19
feat: Substance indexed sets

### LINKED ISSUE #1407: Support sequences in Substance
We have quite a few examples of Substance programs that create long lists of indexed variables by hand.  For instance, in the `random-sampling` example we have programs of the form
```
Interval I
Point p1 = sampleInterval( I )
Point p2 = sampleInterval( I )
Point p3 = sampleInterval( I )
Point p4 = sampleInterval( I )
Point p5 = sampleInterval( I )
...
```

Although it's tempting to introduce an "array" or "list" construct (a la most programming languages), the more natural choice for a language that is meant to resemble mathematical prose is a [sequence](https://en.wikipedia.org/wiki/Sequence).  Notation for sequences, which is remarkably standard across different areas of mathematics, is both flexible and concise:

image

## Sequence declarations

In Substance, we might declare a sequence like this:

```
Point p_1, ..., p_10
```

This declaration would be equivalent to declaring ten variables of type `Point`, indexed from 1 to 10 (inclusive).  Internally (i.e., in our implementation), this declaration can simply be expanded to a list of `Point` variables (as in the original example above).

In general, we would adopt the following conventions:

- The character `_` cannot be used in an identifier (which seems uncommon anyway).
- A sequence declaration takes the form `Type id_i0, ..., id_iN`, where `id` is a valid identifier and both `i0`, `iN` are integers such that `i0 < iN` (not necessarily positive, or even signed).
- Any element of a sequence can be referenced as `id_k`, where `k` is an index between `i0` and `iN` (inclusive).
- Referencing an element outside the sequence is no different from referencing any other undeclared variable.

For instance, one can now write

```
p_4 = sampleInterval( I )
```

even though the identifier `p_4` was never explicitly declared (only the sequence it comes from).

One feature of sequences that is quite different from arrays is that the indexing scheme is more flexible.  For instance, in most programming languages indices are universally fixed to either a 0-based or 1-based convention.  But it's not at all uncommon in mathematics to start indexing at other numbers, e.g., -1.  The proposed scheme naturally allows this kind of indexing, e.g., a declaration

```
SimplicialSet K_-1, ..., K_3 -- elements of a tetrahedral mesh
```

would allow one to address `K_-1`.

## Sequence iteration

In order for sequences to be useful, we also need some way to iterate over them.  Again, though it's tempting to use conventions from programming languages, we should push ourselves to think about syntax that resembles mathematics.  Here, the notation `...` continues to be ubiquitous:

image

Likewise, a reasonable approach in Substance would be something like

```
p_k = sampleInterval( I ), for k = 1, ..., 10
```

Like the sequence declaration, this expression ultimately just expands into a list of ordinary Substance statements

```
p_1 = sampleInterval( I )
p_2 = sampleInterval( I )
p_3 = sampleInterval( I )
...
```

Note that unlike a map/reduce paradigm, we are not required to iterate over the entire sequence.  For instance, even if the declaration is `Point p_1, ..., p_100`, it's perfectly fine to write `p_k = sampleInterval( I ), for i = 1, ..., 10` (and then perhaps `p_k = sampleDisk( D ), for k = 11, ..., 20`, etc.).

## Implementation

A basic implementation of sequences shouldn't be too difficult, as long as we stick with constructions that basically amount to list expansion.  Of course, there may be other features we want to support down the line.  But as usual, I would advocate for: get a basic version, don't sweat the design too much, see how we like using it, and iterate.

wodeni:
# Description

Resolves #1407

This PR introduces _indexed sets_ in Substance, a new syntax for rapidly declaring Substance objects and relations. See #1407 for examples and motivation. Here we outline the design of _indexing statements_, which in the case of sets indexed by a contiguous range of values is similar to Haskell's [list comprehension](https://wiki.haskell.org/List_comprehension).

An indexing statement is a Substance statement with templated identifiers and an indexing clause:

```substance
Vector v_i for i in [0, 2]
-- is equivalent to Vector v_0, v_1, v_2
```

Templated identifiers are regular Substance identifiers consisting alphanumeric characters and underscores, except that the last underscore and the substring following it denote an indexed variable. For example, `_i` denotes an index variable `i` taking on the values in the range of `[0, 2]`, i.e. `0, 1, 2`. When templated identifiers collide with regular identifiers, the latter is shadowed in indexing statements:

```substance
Vector v_i, v_j, vec1
Vector v_i for i in [0, 10] -- ok
Vector v_j for i in [0, 1] -- error: the range of `j` is not defined by `i in [0, 1]`
Orthogonal(v_i, vec1) for i in [0, 10] -- ok
Orthogonal(v_i, v_j) for i in [0, 1], j in [1, 2] -- ok, equivalent to `Orthogonal(v_0, v_1); Orthogonal(v_1, v_2); ...; Orthogonal(v_1, v_2)` 
Orthogonal(v_i, v_j) for i in [0, 5] -- error: the range of `j` is not defined by `i in [0, 5]`
```

When multiple template variables and ranges are present, Substance takes all combinations (cartisian product) of the indices in the ranges. As noted in #1407, this PR also adds support for comparison expressions and boolean logical expressions to allow other kinds of index sets:

```substance
Vector v_i for i in [0, 10] where i % 2 == 0 -- even number: 0, 2, 4, 6, 8, 10
Orthogonal(v_i, v_j) for i in [0, 2], j in [0, 2] where i <= j -- triangular range: [0, 0], [0, 1], [0, 2], [1, 1], [1, 2], [2, 2]
Orthogonal(v_i, v_j) for i in [0, 3], j in [0, 3] where i + 1 == j -- consecutive pairs: [0, 1], [1, 2], [2, 3]
Orthogonal(v_i, v_j) for i in [0, 3), j in [0, 3] where i % 2 == 0 && j == i + 1 -- disjoint pairs of 2: [0, 1], [2,3]
```

An indexing statement generates a list of Substance statements. The existing semantics on duplicates apply to them as well.

```
Vector v_0
Vector v_i for i in [0, 2] -- error: `v_0` is declared and cannot be re-declared
Vector v_i, v_j for i in [0, 2], j in [0, 2] -- error: `v_0` is declared and cannot be re-declared
Orthogonal(v_0, v_1)
Orthogonal(v_i, v_j) for i in [0, 2], j in [0, 2] where i != j -- ok, because duplicated predicates are ok
```

An indexing statement generates identifiers by replacing index variables with strings of integer values. Therefore, generated identifiers can be used just like regular identifiers as long as they exist:

```
Vector v_i for i in [0, 2]
LinearlyDependent(v_0, v_2) -- ok
``` 

## Previous design

There are 2 notable alternatives to the current design:

1. Reserve `_` in regular identifiers: the only difference from the current design is that template identifiers such as `v_i` gets expanded to `v0`, `v1` etc. The main advantage of this approach is that template identifiers are syntactically distinct from regular identifiers. The main downside is that users cannot include `_` in regular identifiers.
2. Use special indexing operators such as `v{i}` and expand to `v0`, `v1` etc: this is similar to 1 but without the cost of removing `_` from regular identifiers
3. Use special indexing operators such as `v{i}` and allow template identifiers everywhere in Substance: Instead of `v1`, users refer to generated identifiers as `v{1}`, which has similar syntax as the template identifiers. This requires the Substance grammar to accept `{}` anywhere a regular identifier could occur and will require the checker to treat `v{1}` as a regular identifier internally.

Notably, both 1 and 2 have the usability disadvantage of visually different syntaxes between template idenfiers and substituted ones. However, 3 requires significant grammar changes and an implicit conversion from indexed identifiers to strings, potentially complicating Style selector matching. The current design attempts to avoid the downsides of 1, 2, and 3 while retaining the current identifier syntax. 

* In addition to indexing clauses, this PR also introduces integer literals and arithmatic (`+`, `-`, `*`, `/`, `%`) and comparison expressions (`<`, `<=`, `>`, `>=`, `==`, `!=`) to Substance.
  * Currently, both expressions can only occur in comparison expressions of indexing clauses

# Checklist

# Open questions


keenancrane:
This is great! Thanks @wodeni (and others) for all the design work done here.

A couple “future proofing” questions:

1. Is this design extensible to non-contiguous index sets (say, only odd numbers in a given range, or Fibonacci numbers in a range?). Feels like the answer is “yes”, as long as there’s a mechanism for specifying the given index set.
2. Does it extend in a natural/easy way to multiple indices? The Cartesian product examples above still refer to pairs from a linear sequence. But suppose you want to declare, say, entries of a matrix. It seems you may run up against the same issues as in item (3). For instance, one natural way to declare such a collection is

```
Type A_{i,j}, for i in [0,5], j in [0,5]
```

Likewise, an element of this set needs to be addressed in some way, e.g., `A_{2,3} `. Given how seemingly tricky parsing is for a single index, just want to be sure we won’t hit a wall with multiple indices.

Thanks!

wodeni:
> 1. Is this design extensible to non-contiguous index sets (say, only odd numbers in a given range, or Fibonacci numbers in a range?). Feels like the answer is “yes”, as long as there’s a mechanism for specifying the given index set.

The current design will only be as expressive as the comparison expressions, which does cover a lot of ground. Obviously we can always add language features, but I think it's beyond the scope of this PR. 

> 2. Does it extend in a natural/easy way to multiple indices? The Cartesian product examples above still refer to pairs from a linear sequence. But suppose you want to declare, say, entries of a matrix. It seems you may run up against the same issues as in item (3). For instance, one natural way to declare such a collection is
> 
> ```
> Type A_{i,j}, for i in [0,5], j in [0,5]
> ```
> 
> Likewise, an element of this set needs to be addressed in some way, e.g., `A_{2,3}`. Given how seemingly tricky parsing is for a single index, just want to be sure we won’t hit a wall with multiple indices.

The semantics of sequence expressions give you a list of tuples, and substituting multiple of them is conceptually the same as inserting strings into a templated string. So this should be okay. Again not this PR but can be a follow-up one.

keenancrane:
> The semantics of sequence expressions give you a list of tuples, and substituting multiple of them is conceptually the same as inserting strings into a templated string. So this should be okay. Again not this PR but can be a follow-up one.

I guess my question was more about the difficulty of parsing a reasonable syntax (i.e., how would multiple indices be declared, and how would they be referenced after declaration), which has been the major stumbling block in terms of implementation.  E.g., is the assertion

```
Type my_var_i_j for i in [0,5], j in [0,5]
```

going to be impossible to parse in a context-free way due to "maximal munch"?  Alternatively, if the set declaration were

```
Type A_{i,j} for i in [0,5], j in [0,5]
```

(which seems to imply a perfectly fine context-free grammar), would it then be possible to refer to elements of this set via a parallel syntax like

```
A_{2,3}
```

? Design alternative (3) in the PR description makes it sound like this would be a big challenge from an implementation point of view.

To be clear, I'm not pushing for one syntax or another—just trying to make sure that there's _some_ reasonable multi-index syntax that is both readable and not a nightmare for us to implement further down the line.

wodeni:
> is the assertion
> 
> ```
> Type my_var_i_j for i in [0,5], j in [0,5]
> ```
>
> going to be impossible to parse in a context-free way due to "maximal munch"? 

No, because the current design is not "context-free," as it involves breaking down a token (i.e. regular IDs) into sub-tokens (i.e. template identifiers where "_" is significant).

> Alternatively, if the set declaration were
> 
> ```
> Type A_{i,j} for i in [0,5], j in [0,5]
> ```
> 
> (which seems to imply a perfectly fine context-free grammar), would it then be possible to refer to elements of this set via a parallel syntax like
> 
> ```
> A_{2,3}
> ```
> 
> ? Design alternative (3) in the PR description makes it sound like this would be a big challenge from an implementation point of view.

Yes, having an identical syntax for template and regular identifiers will inevitably result in context-sensitive grammars. However, my understanding is this is one of our core design requirements. I personally am okay with different syntaxes for templating vs. references, e.g. `A_{i}_{j}` expanding to `A_0_1`, `A_0_2` etc. 

> To be clear, I'm not pushing for one syntax or another—just trying to make sure that there's _some_ reasonable multi-index syntax that is both readable and not a nightmare for us to implement further down the line.

The context-sensitive implementation of multi-index can require a lot of care if we keep insisting on identical syntax for templating and references. The simple heuristic of "anything after the last `_` is an index" will no longer hold, and we need to navigate shadowing carefully.

keenancrane:
> identical syntax for template and regular identifiers … my understanding is this is one of our core design requirements … The context-sensitive implementation of multi-index can require a lot of care if we keep insisting on identical syntax for templating and references.

I don't think that needs to be a core requirement that they're identical, does it?  If that's a blocker to having a reasonable implementation, let's do something else.

For instance, it would seem fine if multi-index declaration was something like
```
Point A_ij for i,j in [0,5] x [0,5]
```
and then to refer to a specific element you'd write
```
A_{1,2}
```
Still don't have a great feel for what's easy/hard to parse…

keenancrane:
> Again not this PR but can be a follow-up one.

Sure I get that.  And if the multi-index discussion is causing headaches, we can put a pin on it and go ahead with the single-index design (which looks great, BTW).  Just wanted to take a moment to look down the road a bit before pulling the trigger.

wodeni:
> > Again not this PR but can be a follow-up one.
> 
> Sure I get that. And if the multi-index discussion is causing headaches, we can put a pin in it and go ahead with the single-index design (which looks great, BTW). Just wanted to take a moment to look down the road a bit before pulling the trigger.

That’d be my preference. The syntax discussion is a time sink and design space is quite constrained. The current design already gets us pretty far, and code can always be refactored IMO.

keenancrane:
> That’d be my preference. The syntax discussion is a time sink and the design space is quite constrained. The current design already gets us pretty far, and code can always be refactored IMO.

Ok! SGTM.

keenancrane:
This is looking good!  Especially exciting to see the diagrams @liangyiliang is coming up with.

One "PR review" comment:

As discussed in [this comment](https://github.com/penrose/penrose/issues/1407#issuecomment-1652758762), the concept being implemented by this PR really is that of an _indexed set_ rather than a _sequence_.  This [comment from StackExchange](https://math.stackexchange.com/questions/1593384/what-is-the-difference-between-an-indexed-family-and-a-sequence) nicely summarizes the difference:

image

I suppose we could continue to call our feature a "sequence" and later add some language to the docs like

> Note that in Penrose the term "sequence" can actually refer to more general _indexed sets_ …

But why not just get it right at the beginning?  Users who already know the difference between these concepts will immediately get the right idea; users who don't think much about sequences vs. indexed sets will also be fine, since "indexed set" is a quite clear and literal English description of the feature we provide (a set of values that you can refer to by index).

To bring the Penrose terminology in line with established terminology, I've substituted `sequence` with `indexed set` and `sequence statement` with `indexing statement` in the PR description.  Ideally, we should also make the following substitutions in the implementation before merging the PR:

```
#region Substance Sequences -> #region Substance IndexedSets
stmt_seq -> stmt_iset
sequence -> indexedset
seq -> iset
StmtSeq -> StmtISet
sequence -> indexedset
Sequence -> IndexedSet
```

If this is too much of a time sink, I guess we can leave it like this in the code—but let's at least get the terminology right in user-facing stuff like the docs.

keenancrane:
Looks good to me, though I can't comment on a lot of the implementation issues.

As noted above, I've changed the user-facing documentation (and test names) so that this feature is more clearly described in terms of indexed sets.

Leo and I just spoke about this in person, and clarified the important distinction between an **indexed set** and an **index set** (without the ed).  Specifically, the latter is the set I of indices i that can be used to index the former.  The concept introduced by the PR is the former, i.e., an _indexed set_.

keenancrane:
As also discussed with @liangyiliang, I added `mod` to the parser as an additional way to express the modulo operation, so people have the choice of writing either

```
Edge( x_i, x_j ) for i in [0, 9], j in [0, 9] where j == (i+1) mod 10
```

or

```
Edge( x_i, x_j ) for i in [0, 9], j in [0, 9] where j == (i+1) % 10
```

In the process I updated the type names in the parser/compiler to reflect the "indexed set" nomenclature, simplified the spectral-graph/torus example, added a Möbius strip example that tests `mod`, and updated the documentation also with `mod`.

image

keenancrane (approved):
Looks good to me!  I've left more extensive comments on the PR.

keenancrane:
Also added `box` and `grid` spectral examples, which provide more thorough testing of @liangyiliang's "calculator." (Thanks to @jiriminarcik for suggesting this pattern.)

image

keenancrane:
There appears to be one small issue with the parser: it seems that operands must be separated by a space character.  E.g., it's not possible to write `i+1`, only `i + 1`.  I'm uncertain whether this is specific to this PR and @liangyiliang, or is just a general issue with the way we parse numbers in Penrose (because we get similar errors in Style).

As a concrete example, if we write

```
Node n_i for i in [0,9]

Edge e_i := MakeEdge( n_i, n_j ) for i in [0,9], j in [0,9] where j == (i+1) % 10
```

then we get the following error

```
Error: Syntax error at line 3 col 74:

  Edge e_i := MakeEdge( n_i, n_j ) for i in [0,9], j in [0,9] where j == (i+1
                                                                           ^
Unexpected int_literal token: "+1". Instead, I was expecting to see one of the following:

    ws token
    ")"
    "+"
    "-"
…
```

If the `i+1` is replaced by `i + 1`, the error goes away and the program works as expected.

liangyiliang:
The issue seems like the parser is parsing "+1" as a single token of a positive integer.

Perhaps I need to give a different order to the `moo` tokenizer.

I'll look into this.

liangyiliang:
@keenancrane The spacing issue seems to be across the board and happens in Style as well, see #1516. So we should fix it altogether, but not in this PR.

keenancrane:
Sounds good to me. @liangyiliang can you put a `:::warning` in the docs about this?

liangyiliang:
@keenancrane done - warning (see below) is added to both Substance and Style docs.
image

keenancrane:
@liangyiliang Excellent, thank you!  There's a small typo `requried` -> `required`.  Otherwise I have nothing more to say about this PR! Looks good.

packages/core/src/compiler/Substance.ts
```diff
@@ -326,37 +396,616 @@ const checkStmt = (stmt: SubStmt<A>, env: Env): CheckerResult<SubStmt<A>> => {
...
+    else if (operator === "<=")
+      return ok(closeEqual(lVal, rVal) ? true : lVal <= rVal);
+    else if (operator === ">=")
+      return ok(closeEqual(lVal, rVal) ? true : lVal >= rVal);
+    else if (operator === "==") return ok(closeEqual(lVal, rVal));
+    else return ok(!closeEqual(lVal, rVal));
+  }
+};
+
+const closeEqual = (x: number, y: number): boolean => {
```
wodeni, line 536: Please note why the `closeEqual` is needed for floating-point comparison here

wodeni (commented):
Still haven't done with `Substance.ts` and `Style.ts`, but here are some comments to help clean up the monadic code in the type checker.

packages/core/src/types/substance.ts
```diff
@@ -137,3 +163,90 @@ export type ApplyPredicate<T> = ASTNode<T> & {
...
+
+export type UnaryBooleanExpr<T> = ASTNode<T> & {
+  tag: "UnaryBooleanExpr";
+  operator: "!";
+  arg: BooleanExpr<T>;
+};
+
+export type BooleanConstant<T> = ASTNode<T> & {
+  tag: "BooleanConstant";
+  value: "true" | "false";
```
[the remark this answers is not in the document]
liangyiliang: Nope, no reason. I guess I didn't want to do the conversion from string to boolean in the parser itself ...

wodeni:
Also the textmate grammar for the docs site can be improved with highlighting for operators:
```diff
diff --git a/packages/vscode/syntaxes/substance.tmGrammar.json b/packages/vscode/syntaxes/substance.tmGrammar.json
index 4e3b2519e..a7c7e8fa3 100644
--- a/packages/vscode/syntaxes/substance.tmGrammar.json
+++ b/packages/vscode/syntaxes/substance.tmGrammar.json
@@ -4,10 +4,11 @@
   "scopeName": "source.penrose-substance",
   "fileTypes": [".substance"],
   "patterns": [
+    { "include": "#comments" },
     { "include": "#keywords" },
     { "include": "#latex" },
     { "include": "#strings" },
-    { "include": "#comments" }
+    { "include": "#numbers" }
   ],
   "repository": {
     "strings": {
@@ -31,16 +32,24 @@
       "patterns": [
         {
           "name": "keyword.control.penrose",
-          "match": "\\b(AutoLabel|Label|NoLabel|All|Let)\\b"
+          "match": "\\b(AutoLabel|Label|NoLabel|All|Let|for|where|in)\\b"
         },
         {
-          "name": "keyword.operator",
-          "match": ":="
+          "name": "keyword.operator.penrose",
+          "match": ":=|>=|<=|>|<|==|\\|\\||&&|!|\\+|\\*|-|%|/|mod|="
         }
       ]
     },
     "latex": {
       "patterns": [{ "name": "constant.language.penrose", "match": "\\$.*\\$" }]
+    },
+    "numbers": {
+      "patterns": [
+        {
+          "name": "constant.numeric.float",
+          "match": "\\b[+-]?(?:\\d+(?:[.]\\d*)?(?:[eE][+-]?\\d+)?|[.]\\d+(?:[eE][+-]?\\d+)?)\\b"
+        }
+      ]
     }
   }
 }

```

packages/core/src/compiler/Substance.ts
```diff
@@ -326,37 +396,616 @@ const checkStmt = (stmt: SubStmt<A>, env: Env): CheckerResult<SubStmt<A>> => {
...
+  }
+  if (b.tag === "BooleanConstant") {
+    const { value } = b;
+    return ok(value === "true");
+  } else if (b.tag === "BinaryBooleanExpr") {
+    const { operator, left, right } = b;
+    if (operator === "&&") {
+      const lValRes = evalCond(left, subst);
+      if (lValRes.isErr()) return err(lValRes.error);
+      if (!lValRes.value) return ok(false);
```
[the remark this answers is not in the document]
liangyiliang: The linter is giving me this error - 
image

packages/core/src/compiler/Substance.ts
```diff
@@ -326,37 +396,609 @@ const checkStmt = (stmt: SubStmt<A>, env: Env): CheckerResult<SubStmt<A>> => {
...
+  n: NumExpr<A>,
+  subst: ISetSubst,
+): Result<number, SubstanceError> => {
+  const result = evalNumHelper(n, subst);
+  if (result.isErr()) return err(result.error);
+
+  const value = result.value;
+
+  // This is the case of NaN
+  if (value !== value) {
```
wodeni, line 547: Use [`isNaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN) instead?

packages/core/src/compiler/Substance.ts
```diff
@@ -326,37 +396,609 @@ const checkStmt = (stmt: SubStmt<A>, env: Env): CheckerResult<SubStmt<A>> => {
...
+  if (underscorePos === -1) {
+    return ok(v);
+  }
+
+  const prefix = v.slice(0, underscorePos);
+  const isetVarName = v.slice(underscorePos + 1);
+
+  const isetVarValue = substISetVarNumber(isetVarName, location, subst);
+  if (isetVarValue.isErr()) return err(isetVarValue.error);
+  return ok(`${prefix}_${isetVarValue.value}`);
```
wodeni, line 634:
```suggestion
  return andThen((idx) => ok(`${prefix}_${idx}`), isetVarValue);
```

packages/core/src/compiler/Substance.ts
```diff
@@ -326,37 +396,609 @@ const checkStmt = (stmt: SubStmt<A>, env: Env): CheckerResult<SubStmt<A>> => {
...
+  subst: ISetSubst,
+): Result<Identifier<A>, SubstanceError> => {
+  const substContents = substISetVarStr(id.value, id, subst);
+  if (substContents.isErr()) return err(substContents.error);
+
+  return ok({
+    ...id,
+    value: substContents.value,
+  });
+};
```
wodeni, line 648:
```suggestion
): Result<Identifier<A>, SubstanceError> =>
  andThen(
    (substitutedID: string) =>
      ok({
        ...id,
        value: substitutedID,
      }),
    substISetVarStr(id.value, id, subst),
  );
```

packages/core/src/compiler/Substance.ts
```diff
@@ -326,37 +396,609 @@ const checkStmt = (stmt: SubStmt<A>, env: Env): CheckerResult<SubStmt<A>> => {
...
+
+const substISetDeconstructor = (
+  deconstr: Deconstructor<A>,
+  subst: ISetSubst,
+): Result<Deconstructor<A>, SubstanceError> => {
+  const { variable } = deconstr;
+  const substVariable = substISetId(variable, subst);
+  if (substVariable.isErr()) return err(substVariable.error);
+  return ok({ ...deconstr, variable: substVariable.value });
```
wodeni, line 699: Can refactor to one-liner similar to above

packages/core/src/compiler/Substance.ts
```diff
@@ -326,37 +396,609 @@ const checkStmt = (stmt: SubStmt<A>, env: Env): CheckerResult<SubStmt<A>> => {
...
+  const { variable } = labelDecl;
+
+  const substVariable = substISetId(variable, subst);
+  if (substVariable.isErr()) return err(substVariable.error);
+
+  return ok({
+    ...labelDecl,
+    variable: substVariable.value,
+  });
+};
```
wodeni, line 771: Similar refactor to reduce line counts

packages/core/src/compiler/Substance.test.ts
```diff
@@ -277,12 +277,93 @@ NoLabel B, C
...
+      const env1 = envOrError(domainProg);
+      const prog1 = `Set a_i for i in [1, 10] where i % 2 == 0.5`;
+      const res1 = compileSubstance(prog1, env1);
+      expect(res1.isOk()).toBe(true);
+      if (res1.isOk()) {
+        expect(res1.value[1].vars.size).toBe(0);
+      }
+
+      const env2 = envOrError(domainProg);
+      const prog2 = `Set a_i for i in [11, 10]`;
```
wodeni, line 307: This is because the index ranges need to be ascending?

wodeni (commented):
Finished reviewing the whole thing. Great job on the PR and thanks for making this a reality @liangyiliang! There are some additional spots to refactor if you want to improve the readability and/or make the code a little shorter. 

One question I didn't totally get: why do the checker functions return lists of statements, instead of the previous interface? Probably missed that while reviewing

liangyiliang:
@wodeni 

> why do the checker functions return lists of statements, instead of the previous interface?

One substance-statement can be compiled into multiple compiled-substance-statements in the resultant `CompiledSubProg` code.

Example 1: aggregated statements. A single Substance statement
```
Edge e := MakeEdge(n1, n2)
```
compiles into two Compiled Substance statements
```
Edge e
e := MakeEdge(n1, n2)
```

Example 2: substance indexed sets.
```
IsValid(n_i) for i in [1, 3]
```
compiles into
```
IsValid(n_1)
IsValid(n_2)
IsValid(n_3)
```

To make the code more general I am allowing the checkers to return more than one statements.

wodeni:
ahh makes sense. thanks for clarifying @liangyiliang!

packages/core/src/compiler/Substance.test.ts
liangyiliang: Yes. My interpretation is that `i in [11, 10]` means `i` is an integer and `i >= 11` and `i <= 10` which is impossible.

packages/core/src/compiler/Substance.ts
liangyiliang: What's the difference between `andThen(..., isetVarValue)` and `isetVarValue.andThen(...)`? I find the latter to be more intuitive.

packages/core/src/compiler/Substance.test.ts
wodeni: got it. can you add a quick thing to the docs on this?

packages/core/src/compiler/Substance.ts
wodeni: never knew that existed! if that works i agree it’s more intuitive for sure

liangyiliang: yep it works

packages/core/src/compiler/Substance.test.ts
liangyiliang: sounds good, will do it

PR #1583, opened 2023-07-24
feat: `numberof` and `nameof`

### LINKED ISSUE #1577: Count of substance collection
We should add an expression that counts the number of elements in a substance collection. For example,
```
collect Object o into os {
  c = countof os
}
```
We already have the Collection Access expression `listof someprop from collection` which _can_ give us the same thing, but it requires a dummy variable:
```
forall Object o {
  o.dummy = 1
}
collect Object o into os {
  c = count(listof dummy from os)
}
```

It would be nice to add the `countof` expression without a dummy variable.

liangyiliang:
# Description

Resolves #1577.

We add two keywords, `numberof` and `nameof`, to give us access to more metadata about the style-substance substitutions.

If `xs` is a substance variable collection, then `numberof xs` gives us a `FloatV` with the number of elements in the collection. This can be useful in computing, for example, the spacing between two objects arranged in a line.

If `x` is a Style variable, then `nameof x` gives us a `StrV` with the name of the Substance variable that is matched by `x`. This can potentially be useful for showing errors as a part of the diagram. For example, we might want to show an error when a set is both a subset and not a subset of another set. We can write,
```
forall Set a; Set b
where IsSubset(a, b); Not(IsSubset(a, b)) {
  errorMessage = Text {
    string: "Set " + (nameof a) + " is both a subset and not a subset of Set " + (nameof b)
  }
}
```

This may allow the users to detect some inconsistencies (which indicate errors) in the Substance program.

Since `listof` (already exists), `numberof`, and `nameof` all operate upon Style-variables, we add a new category of expressions, called `sty_var_expr`. The grammar is as follows:
```
sty_var_expr := listof <prop_name> from <coll_name>
              | numberof <coll_name>
              | nameof <var_name>
```

Handling of `listof` remains unchanged. We handle `numberof` by returning the length of the collection; we handle `nameof` by extracting the name of the corresponding Substance variable.

https://countof.penrose-72l.pages.dev/try/?gist=43123d97f7063f94b84414bdae5217d8

This diagram of the dining philosophers problem uses both `numberof` (for determining the spacing between two philosophers) and `nameof` (for showing error message).

# Checklist

### DIFF
--- a/packages/components/src/editing/languages/StyleConfig.ts
+++ b/packages/components/src/editing/languages/StyleConfig.ts
@@ -50,6 +50,7 @@ const styleCustoms = {
     "where",
     "with",
     "listof",
+    "countof",
     "from",
     "delete",
     "as",
--- a/packages/components/src/editing/languages/StyleConfig.ts
+++ b/packages/components/src/editing/languages/StyleConfig.ts
@@ -51,6 +51,7 @@ const styleCustoms = {
     "with",
     "listof",
     "countof",
+    "nameof",
     "from",
     "delete",
     "as",
--- a/packages/components/src/editing/languages/StyleConfig.ts
+++ b/packages/components/src/editing/languages/StyleConfig.ts
@@ -50,7 +50,7 @@ const styleCustoms = {
     "where",
     "with",
     "listof",
-    "countof",
+    "numberof",
     "nameof",
     "from",
     "delete",
--- a/packages/core/src/compiler/Style.test.ts
+++ b/packages/core/src/compiler/Style.test.ts
@@ -1354,7 +1354,7 @@ delete x.z.p }`,
         }
         collect T t into ts {
           Circle {
-            r: countof ts
+            r: numberof ts
           }
         }
       `;
--- a/packages/core/src/compiler/Style.test.ts
+++ b/packages/core/src/compiler/Style.test.ts
@@ -1080,7 +1080,7 @@ delete x.z.p }`,
         }
         `,
       ],
-      UnexpectedCollectionAccessError: [
+      NotSubstanceCollectionError: [
         `forall Set a {
           a.c = 10
           x = listof c from a
@@ -1354,7 +1354,7 @@ delete x.z.p }`,
         }
         collect T t into ts {
           Circle {
-            r: count(listof value from ts)
+            r: countof ts
           }
         }
       `;
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -131,13 +131,13 @@ import {
   err,
   invalidColorLiteral,
   isErr,
+  notCollectionError,
   ok,
   parseError,
   redeclareNamespaceError,
   safeChain,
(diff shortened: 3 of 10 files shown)


wodeni:
Will check the code out soon. Meanwhile, could you add this to the language docs too?

liangyiliang:
> Will check the code out soon. Meanwhile, could you add this to the language docs too?

Yes I will do that

liangyiliang:
@wodeni docs added

liangyiliang:
Out of camelCase and snake_case, I personally prefer snake_case simply because no other keyword in the Style language uses upper-case letters. Though I also don't particularly like the idea of underscores in keywords, I guess we have already violated that principle with `match_id` and `match_total`, so I'm fine with snake_case.

wodeni (approved):
Thanks for adding this feature! I do want to note one alternative to this syntax: pre-populated Style fields, i.e. `countof A` => `A.count` and `nameof A` => `A.name`. Suggesting this mainly because we already do this with label texts, where `A.label` has the string label defined by the Substance program. This obviously has a disadvantage of possible `delete` and `override` by the Style program itself, so I don't think this option is without downsides. 

In terms of the word choices, my preference is to merge the current design as-is. Generally, I'd like to figure out an open process for the broader community, in addition to core maintainers, to give feedback on new language features.

liangyiliang:
@keenancrane @wodeni

Would it be a good idea to add the operator `labelof` (a easy thing to add), while still keeping (but discouraging) the use of `x.label` until the next major version?

If so, I will change all the occurrence of `x.label` in the examples to use `labelof`, and disable `x.label` in Version 4.0.

keenancrane:
> Would it be a good idea to add the operator labelof (a easy thing to add in this PR), while still keeping (but discouraging) the use of x.label until the next major version?

@liangyiliang Interesting—I would have thought the easiest path would be to just rename `countof` -> `count_of` and `nameof` -> `name_of` in your PR, so there's some consistency with the naming of existing keywords (`match_id` and `match_total`).  That's just some search/replace in a few of the files you've modified, right?

Deprecating `x.label` sounds like a bigger change that's not suuuuuuper related to this PR…

keenancrane:
> I personally don't like how x.label is a language feature but is represented just like any other user-defined field. I think generally it is a good idea to support labelof (or, of course, label_of).

Yeah, I can get behind that.

> This labelof is easy to add within this PR.

Can we discuss this in a separate issue / PR?  When I think about using `label_of` as proposed here, it reads a bit odd—see the example below.  There are some design choices here, e.g., should it be `label_of x` or should it be `label_of(x)`?  These questions are quite orthogonal to adding `count_of` and `name_of`.

```
forall Point p
where p has label {
   shape p.labelText = Text {
      string: label_of p
   }
}
```

> I am not talking about disallowing x.label in this PR - it is just that x.label is something that I think should be deprecated.

Sure.  Let's just address all of this in a separate issue / PR.  Generally better to have smaller, targeted PRs anyway.

Thanks @liangyiliang!

PR #1589, opened 2023-07-25
test: add `eslint-plugin-unary-minus`

samestep:
# Description

I created https://www.npmjs.com/package/eslint-plugin-unary-minus to catch bugs like this:

```typescript
(x: ad.Num) => -x
```

This PR adds that plugin as a dependency.

The new plugin depends on pretty recent versions of TypeScript and ESLint, so I also needed to update those in our own dependencies. That resulted in a couple minor lint and formatting changes (I guess TypeScript v5.1 now lets us automatically sort `export`s as well as `import`s).

Copy-paste the above code into a TypeScript file in `packages/core/src/`.

# Checklist

# Open questions

As suggested by https://github.com/typescript-eslint/typescript-eslint/issues/7076#issuecomment-1652240079, this lint might become subsumed into typescript-eslint itself; if that happens then we can remove this dependency.


wodeni (approved):
Great! Thanks for building the plugin :D

PR #1593, opened 2023-08-01
build: add rust-toolchain.toml for automatic toolchain installation

jhvst:
# Description

While [packaging penrose for Nix](https://github.com/jhvst/nix-config/commit/d749ea86b31782f51c0cf10c0b407b636063bd84), I found that a common practice to add Rust toolchains or select specific version is by doing [overrides](https://rust-lang.github.io/rustup/overrides.html). Nix in particular likes the toolchain file, which if present, will automatically run the equivalent "rustup" command. In penrose case, the file can be used to add the wasm toolchain required to compile the roger binary.

As this helped me, I was wondering if you would be interested in merging this request and/or possibly pinning the Rust version furthermore using the toolchain file. For me, having this file on upstream skips a patch over the source repository.

N/A

N/A?

# Open questions

See above

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,12 +43,6 @@ Be sure you have these tools installed:
 
 - [Rust][]
 
-  - the WebAssembly target for Rust:
-
-    ```sh
-    rustup target add wasm32-unknown-unknown
-    ```
-
   - `wasm-bindgen` CLI v0.2.84:
 
     ```sh
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,5 @@
+[toolchain]
+channel = "stable"
+components = []
+targets = [ "wasm32-unknown-unknown" ]
+profile = "default"
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "stable"
+channel = "1.71.0"
 components = []
 targets = [ "wasm32-unknown-unknown" ]
 profile = "default"


samestep (changes requested):
Awesome, thanks for doing this! I was going to but have been a bit lazy 😅

> I was wondering if you would be interested in merging this request and/or possibly pinning the Rust version furthermore using the toolchain file.

Definitely. If you want to go ahead and pin the Rust version in this PR (e.g. to v1.71.0) then that'd be great.

Minor note: you need to edit your PR title to conform to the "conventional commits" style; if you look at the [failing CI log](https://github.com/penrose/penrose/actions/runs/5724641909/job/15511541733?pr=1593) you should be able to see the constraints:

> Available types:
>  - feat: A new feature
>  - fix: A bug fix
>  - docs: Documentation only changes
>  - style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
>  - refactor: A code change that neither fixes a bug nor adds a feature
>  - perf: A code change that improves performance
>  - test: Adding missing tests or correcting existing tests
>  - build: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
>  - ci: Changes to our CI configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs)
>  - chore: Other changes that don't modify src or test files
>  - revert: Reverts a previous commit

Also, since you've included the `wasm32-unknown-unknown` target in the toolchain file, could you remove it from our `CONTRIBUTING.md` instructions?

https://github.com/penrose/penrose/blob/95ab946ea5e3c842ffa399c102c85513235f0059/CONTRIBUTING.md#L46-L50

jhvst:
Thanks, I will make the changes at some point later this week.

samestep (approved):
Looks great, thanks for making the changes!

samestep:
Hey @jhvst, after merging this we've been getting [Cloudflare build errors](https://discord.com/channels/1115349463603617954/1138171222987919420/1138200753018765333) that didn't show up on this initial PR because we don't run the Cloudflare build on forks; any idea what's going on / how to fix this?

> error: component download failed for cargo-x86_64-unknown-linux-gnu: could not rename downloaded file from '/opt/buildhome/.rustup/downloads/fe6fb520f59966300ee661d18b37c36cb3e614877c4c01dfedf987b8a9c577e9.partial' to '/opt/buildhome/.rustup/downloads/fe6fb520f59966300ee661d18b37c36cb3e614877c4c01dfedf987b8a9c577e9'

I unfortunately need to revert this PR until we figure out how to fix this.

samestep:
Not sure exactly; but also, regarding the point about this appearing in Cloudflare but not GitHub Actions, note that the error _does_ occur in GitHub Actions if we remove our explicit `rustup target add` command (which should no longer be necessary after this PR); see #1606.

If you have time to take a a look later then that'd be awesome!

jhvst:
Yeah, I'll take a look; I maintain some Rust projects, so happy to see how to get the file to work correctly.

samestep:
Possibly relevant: we already use a [`rust-toolchain.toml`](https://github.com/rose-lang/rose/blob/e5a34e8af63fcf89a6eb332b6bf9b6ee92f8a03d/rust-toolchain.toml) file in a related project that @ravenrothkopf and I maintain, and we haven't run into this issue; here's our [`Makefile`](https://github.com/rose-lang/rose/blob/e5a34e8af63fcf89a6eb332b6bf9b6ee92f8a03d/Makefile) and [GitHub Actions workflow](https://github.com/rose-lang/rose/blob/e5a34e8af63fcf89a6eb332b6bf9b6ee92f8a03d/.github/workflows/build.yml#L14). One hypothesis I have is that we don't run into the problem there because our pinned nightly Rust version doesn't conflict with anything which might already be on the runner, whereas the latest stable release v1.71.0 may conflict somehow; would that make any sense, or is that reasoning completely misguided?

jhvst:
I can see it conflicting. I don't know how rustup determines which toolchains it downloads by default, but I can see that if the runner expects _not_ to download any overriding "basic" toolchains, then maybe it gets confused. The error suggests it to an extent: the target is "unknown" linux gnu, so maybe setting it to x86 would help -- I don't expect most runners to use ARM, but it's also not great to set it explicitly unless warranted to avoid this projects from not being able to compile on say ARM darwin or ARM Linux. But, as said, I will look into this.

PR #1599, opened 2023-08-03
feat: row-indexing of matrices

### LINKED ISSUE #1509: Support row-indexing of matrices
Currently in Style, we only allow for accessing specific elements of a matrix using two indices. For example, if `m` is a matrix, then we can only do `m[5][6]` to get one element, not `m[5]` to get the entire row (as a vector) at index 5.

This is currently implemented in this code:
```ts
    ...
    case "LListV":
    case "MatrixV":
    case "PtListV": {
      if (indices.length !== 2) {
        return err({ tag: "BadIndexError", expr });
      }
      const [i, j] = indices;
      if (!isValidIndex(coll.contents, i)) {
        return err({ tag: "OutOfBoundsError", expr, indices });
      }
      const row = coll.contents[i];
      if (!isValidIndex(row, j)) {
        return err({ tag: "OutOfBoundsError", expr, indices });
      }
      return ok(floatV(row[j]));
    }
    ...
```
in the `evalAccess` function.

We should be able to access specific rows of a matrix.

liangyiliang:
# Description

Resolves #1509.

This PR allows Style writers to extract an entire row from a matrix. Previously, we only allow indexing for a specific element of matrix with two indices.

We add a case when the number of indices is 1. In that case, after checking for out-of-bound errors, we directly extract that row as a vector.

# Checklist

### DIFF
--- a/packages/core/src/compiler/Style.test.ts
+++ b/packages/core/src/compiler/Style.test.ts
@@ -1,5 +1,6 @@
 import im from "immutable";
 import { describe, expect, test } from "vitest";
+import { numsOf } from "../contrib/Utils.js";
 import { C } from "../types/ast.js";
 import { Either } from "../types/common.js";
 import { Env } from "../types/domain.js";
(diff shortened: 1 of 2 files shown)


wodeni (approved):
LGTM!

PR #1638, opened 2023-09-21
feat: initial values for unknown variables

### LINKED ISSUE #1633: Support specifying initial value to varying values.
We used to have a feature called `VARYING_INIT` that specifies an initial value to a varying variable. See [old wiki](https://github.com/penrose/penrose/wiki/Language-Specification#specifying-a-custom-initialization-value-for-varying-variables). This was removed somewhere in the production.

We should add it back, but in a way that elegantly plays well with staged layout. One can already write things like `? in stage`, so varying variables with initial values should look similar to that.

Some ideas for syntax are `?(init_value)` or `?[init_value]`. This syntax constructs a varying variable that has an initial value. Then one can also write `?(init_value) in stage` or `?[init_value] in stage`.

liangyiliang:
# Description

Resolves #1633.

This PR allows users to provide initial values for unknown variables.

Users now can write `?[some_init_value]` to construct a variable with unknown value, optimized using the initial value `some_init_value`. The previous behavior of `?` (without the explicit initial value) remains unchanged.

Whenever the optional `some_init_value` is provided, the compiler construct an unknown value variable using an "init sampler" that only outputs that `some_init_value` instead of the original uniform sampler.

This PR also "sprinkled" some uses of this new syntax in the tests.

The reference pages have also been updated to reflect the new syntax.

# Checklist

# Open questions

Do we need to support custom initial ranges? E.g., `?[10, 20]` will have the initial value sampled between 10 and 20. My gut tells me this is not necessary, but I'm not sure.


wodeni (commented):
The implementation looks good to me! Pretty straightforward. The syntax is a pretty new to me. Have you searched for syntaxes in other languages for similar things? The most conventional thing is probably the constructor-argument syntax for initial values: `varying(5.0)`, but there might be others. I'm okay with merging this for now and be open to syntax suggestions in the near future.

liangyiliang:
Yup, I came up with the syntax `?[0.5]` on the fly simply because it looks similar to `?`. If we can type `? in stage` then I think it would make sense to type `?[0.5] in stage`. Given that `?` already exists, other syntax like `varying(0.5)` might be confusing.

liangyiliang:
I think the important part of the syntax is the "unknown" part (represented by `?`). The initial value is simply an "annotation" to this "unknown" value.

wodeni:
> I think the important part of the syntax is the "unknown" part (represented by `?`). The initial value is simply an "annotation" to this "unknown" value.

Yep I understand the motivation here. I have no good alternatives to suggest. We can PR again if others have suggestions.

liangyiliang:
Sounds good, merging it now

PR #1651, opened 2023-10-26
chore: rename `contrib` to `lib` in `core`

### LINKED ISSUE #1597: chore: rename `/packages/core/src/contrib` to `lib/`
This path was originally named `contrib` because we imagined it might be an entry point for contributors, but might now be better called `lib`.

wodeni:
# Description

Resolves #1597.


samestep (approved):
Makes sense to me; thanks!

PR #1171, opened 2022-12-10
feat: Make SVGs "Penrose-editable"

lmxlucy:
# Description

Related issue/PR: #1129 

Provide editability of any (SVG) diagram created in Penrose.
1. SVGs exported from Penrose contain (in metadata) the entire source code used to generate the diagram (Domain, Style, and Substance), along with the seed and software version.
2. The Penrose IDE can directly import an SVG, and automatically import this metadata.

# Implementation strategy and design decisions

Penrose SVG Export:
1. Figure out where SVGs are exported
2. Programmatically retrieve the Penrose version, variation string, substance code, style code, and domain code 
3. Add this code to the exported SVG metadata

Penrose SVG Import:
1. Figure out how to upload SVG (dropzone in a new SVG tab)
2. Figure out how to load trios (focus on In-browser storage)
3. Extract trios content from SVG and load this code to trios in browser

# Examples with steps to reproduce them

Export Penrose SVG
1. Load word-cloud trios from _Roger_
2. Click _Compile_ to compile diagram from triosScreen Shot 2022-12-09 at 8 41 38 PM
3. Click _SVG_ to export the diagram in svg
4. Open the exported svg in text editor to see added Penrose tagScreen Shot 2022-12-09 at 8 44 45 PM

Import Penrose SVG
1. For comparison purpose, load walk-on-spheres from _Roger_  and compileScreen Shot 2022-12-09 at 8 45 33 PM
2. Upload the word-cloud svg downloaded above in the _svg_ tab and CompileScreen Shot 2022-12-09 at 8 48 13 PM
3. Check trios content and the newly compiled diagram match the uploaded svg (the titles of trios stay as walk-on-spheres)Screen Shot 2022-12-09 at 8 48 49 PM

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new ESLint warnings
- [x] I have reviewed any generated changes to the `diagrams/` folder

PR #1652, opened 2023-10-26
fix: duplicate `<penrose>` tags in exported SVGs

### LINKED ISSUE #1586: bug: two <penrose> tags in Penrose-exported SVGs
**Describe the bug**

Here is the SVG code of a Penrose-exported diagram:

image

As seen, the code contains two `<penrose>` blocks. The first one only contains the cropped view box. The second one contains the same cropped view box, as well as the source code information.

wodeni:
# Description

Resolves #1586.

The cause is an extra `svg.appendChild(metadata)` call at the end of the `SVGAddCode` function introduced in https://github.com/penrose/penrose/pull/1171. This line should only be executed when there isn't already a `<penrose>` node.

### DIFF
--- a/packages/editor/src/components/DiagramPanel.tsx
+++ b/packages/editor/src/components/DiagramPanel.tsx
@@ -174,11 +174,8 @@ const SVGaddCode = (
   versionStr: string,
   variationStr: string,
 ): void => {
-  console.log(svg);
-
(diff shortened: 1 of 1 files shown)


kyleliangus:
Looking forward to this bugfix!

samestep (approved):
Good catch, thanks!

PR #1674, opened 2023-12-11
chore: add default canvas block to `editor`

wodeni:
# Description

On start, `editor` starts the trio with empty strings, and it's a bit tedious to type `canvas { ... }` every time. This PR inserts a default canvas block to the Style pane by default and exposes a code snippet `canvas` that inserts a template for the canvas block:

```
canvas {
  width = 400
  height = 400
}
```

* Go to deployment IDE
* Hit "compile"
* Observe no errors are thrown because the canvas is already pre-populated.

# Checklist

# Open questions

### DIFF
--- a/packages/editor/src/state/atoms.ts
+++ b/packages/editor/src/state/atoms.ts
@@ -200,7 +200,10 @@ export const currentWorkspaceState = atom<Workspace>({
       },
       style: {
         name: ".style",
-        contents: "",
+        contents: `canvas {
+  width = 400
+  height = 400
+}`,
       },
       domain: {
         name: ".domain",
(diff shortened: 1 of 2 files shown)


packages/editor/src/state/atoms.ts
```diff
@@ -200,7 +200,10 @@ export const currentWorkspaceState = atom<Workspace>({
       },
       style: {
         name: ".style",
-        contents: "",
+        contents: `canvas {
+  width = 400
```
hsharriman, line 204: is the spacing correct on these lines? they look like they're not indented enough in the PR

wodeni:
Note after testing a bit: having something in the Style file on start does trigger an alert box when switching to an example because the alert basically checks if the editor panes are empty. Prolly the most correct thing to do, but I'd prefer to push this decision a later time when we re-examine the state management system.

Issue #1110, opened 2022-09-24
`checkSymmetricArgs` checks empty type graphs


wodeni:
**Describe the bug**

When checking `symmetric`, the checker attempts to use `typeGraph` by calling `isSubtype` for both arguments to make sure they are identical. When they are actually identical types, this is okay. However, when one is a subtype of another, `isSubtype` returns `false` for both `isSubtype(MyType, MySubType)` and `isSubtype(MySubType, MyType)`. 

```
type MyType
type MySubType <: MyType
symmetric predicate MyBadPredicate(MyType a, MySubType b)
```

**To Reproduce**

Print out `console.log(isSubtype(type1, type2, env), isSubtype(type2, type1, env));` in `areSameTypes` when running the Domain tests and observe that only `true true` and `false false` show up. In one of the cases, it should print out `true false` or `false true`. 

**Expected behavior**
`isSubtype` cannot be called within the Domain checker, especially while the checker is going through the program. This is because `computeTypeGraph` is not called until the checking is done. 

Because `checkSymmetricArgs` doesn't want subtypes at all. It could check the names of predicate arguments instead. 

Alternatively, run the symmetry check _after_ the Domain checker finishes its pass through the Domain program.

Or, construct the type graph as the checker runs, but one will need to pay attention to where `<:` statements show up in the program, because a naive implementation will need them to appear before any symmetric predicates.

liangyiliang:
> It could check the names of predicate arguments instead.

One issue is that our types might be `TypeFunctions` which are present in the language ... I guess it's some type polymorphism feature which we haven't thoroughly tested?

wodeni:
> One issue is that our types might be `TypeFunctions` which are present in the language ... I guess it's some type polymorphism feature which we haven't thoroughly tested?

Fun fact: Penrose used to have dependent types! But now it's basically generics that have some tests in Domain and Substance test suites. No end-to-end examples nor Style tests for sure.

liangyiliang:
> > One issue is that our types might be `TypeFunctions` which are present in the language ... I guess it's some type polymorphism feature which we haven't thoroughly tested?
> 
> Fun fact: Penrose used to have dependent types! But now it's basically generics that have some tests in Domain and Substance test suites. No end-to-end examples nor Style tests for sure.

Hmmm ... if so, shouldn't we remove type polymorphism completely?

samestep:
[...]

PR #1677, opened 2023-12-13
fix!: separate Substance environment and remove unused features

### LINKED ISSUE #742: Bracket choice for "generics" in Domain programs
In https://github.com/penrose/penrose/pull/737#discussion_r777599227 @wodeni suggested that it may be more readable to use angle brackets instead of our current mix of round and square brackets. So for instance, this:
```
constructor Cons ['X] ('X head, List('X) tail) -> List('X)
```
would become this:
```
constructor Cons<'X> ('X head, List<'X> tail) -> List<'X>
```

### LINKED ISSUE #1251: Remove type polymorphism
Many of the parser and compiler code are designed to work with type polymorphism (i.e. with type constructors, type arguments, etc.), but as discussed in #1110 , type polymorphism only has unit tests but no overall tests or end-to-end examples, and the @samestep suggests that "generics just don't work at all currently."

If there is no reasonable use case for type polymorphism, should we just remove it?

### LINKED ISSUE #1676: Improper handling of `Prop`-typed expressions in Substance
The `Prop` Substance type is handled incorrectly.

See [example](https://penrose.cs.cmu.edu/try/?gist=a401ea6a90b4538a82d1c1890d19b761).

Here, `Pred` is defined as `predicate Pred(T, T)`, but in the Substance program it is given `Pred(t, Pred(t, t))`. It should fire an error but it doesn't.

liangyiliang:
# Description

**Issues this PR eliminates by removing relevant features:**
Resolves #742 
Resolves #1676 

**Issues resolved by this PR:**
Resolves #1251

This PR is a general refactoring of the designs and implementations of the Domain, Substance, and Style languages. It removes some features that were half-baked into the system (at varying levels of support) during the initial stages of design, which we later discovered are not useful nor necessary. It also overhauls the Substance compiler to output useful information that can be used by other parts of Penrose.

# "Features" removed

- Nested predicates and function calls; the `Prop` type. Instead of writing `Not(Intersecting(a, b))` one can just have a different predicate `NotIntersecting(a, b)`, so we aren't losing much here. Removing this feature reduces the complexity of both the Substance compiler and the Style compiler (since it makes style-substance matching easier).
- Polymorphic types (types with type arguments) (see #1251). Newer features of Penrose just don't work with polymorphic types. According to #1110, it had Domain and Substance support but was never tested in Style.
- "Dead" language features that had parser support but are ignored by compiler:
  - Prelude values
  - Notations
  - `EqualExprs`
  - `EqualPredicates`

Removal of these features involve removing parser support as well as compiler support. TypeScript types concerning these features are also gone.

# Changes to Substance compiler

The Domain compiler outputs an `Env` (henceforth called `DomainEnv`) which encodes information about the Domain, like which types, predicates, constructors, and functions can be used.

The Substance compiler takes in the `DomainEnv` and a Substance program, checks the program against the `DomainEnv` (e.g. does the Substance program respect the types in the Domain?), and outputs **both** a `DomainEnv` and a `SubstanceEnv`.

Observing the `DomainEnv` and the `SubstanceEnv`: 
image

Fields of the two environments concerning the features we removed (see above) are crossed out. 

The more significant problem is that the two lines for `vars` and `varIDs` (boxed) actually store information about the Substance program -- `vars` maps Substance variables to their types, and `varIDs` is a list of names of all Substance variables -- but they are placed within `DomainEnv` for some reason.

As a result, `SubstanceEnv` contains no useful information other than the compiled AST of the Substance program.

This PR removes unnecessary information from both environments, and rearranges the fields so that `DomainEnv` and `SubstanceEnv` are clearly separated. The idea is, `DomainEnv` should only be modified by the Domain compiler, and can only be accessed but not modified by the Substance compiler. The Substance compiler would modify the `SubstanceEnv`.

Here is the resultant changes to the two environments:
image

We modify the Substance compiler to output the `SubstanceEnv` that stores the useful information.

# Changes to Style header checker

Since a Style header resembles a Substance program, to check whether or not it is well-formed, one can just convert each part of the Style header into a Substance statement and use the Substance checkers. In fact, this is already partially what we do, so this PR just goes a bit further along this line, reusing Substance checkers as much as possible with a few caveats, namely:

- `a has label` relations are not checked by Substance checkers but separately within Style compiler, and
- aliases for predicate applications are checked separately too.

The Style compiler checks a Style header by:
- start by constructing a fake `SubstanceEnv`
- go through each statement in Style:
  - convert that statement into its equivalent form in Substance
  - use the `SubstanceEnv` and the Substance checker to check that Substance statement, and get a new `SubstanceEnv` with the updated variable bindings, etc.
  - do special handling for `a has label` relations and predicate aliases, as necessary.

# Checklist

# Open questions


wodeni (changes requested):
Thanks for the PR! Still reviewing `Style.ts` but wanted to show some older comments first. Also this PR is technically a breaking change, so I'd suggest labeling it `fix!` and renamed it to something like "fix!: separate Substance environment and remove unused features."

packages/edgeworth/src/components/Content.tsx
```diff
@@ -153,7 +154,13 @@ export class Content extends React.Component<ContentProps, ContentState> {
             );
           }
         }
-        const synth = new Synthesizer(env, setting, subResult, seed);
+        const synth = new Synthesizer(
```
wodeni, line 157: `Synthesizer` is really supposed to take in a `DomainEnv` and an optional `SubstanceEnv`. Since we have additional PRs depending on this one. I'd prefer to do this myself in another PR after this one is merged.

wodeni (approved):
Trying to speed up reviewing so I won't go through this PR line-by-line. Skimming the PR seemed great to me. I'm okay with merging this one for now and will note (in comment below) that I'll need to do another PR to refactor `edgeworth` a bit more.

PR #1678, opened 2023-12-13
fix: download functionality in the editor

hsharriman:
# Description
Motivation: Update and consolidate the functionality surrounding downloading a diagram that is created in the Penrose editor.
Screenshot 2023-12-13 at 5 02 03 PM
Screenshot 2023-12-13 at 5 02 23 PM

Changes made:
* Added "export" and "save as Penrose SVG" buttons to the header of the editor
* Added functionality to save a Penrose trio as a zipped file
* Moved the old buttons to save as SVG, SVG (TeX), PDF, and PNG and put their functionality in the header
* Created a streamlined `DropdownButton` component
* Fixed bug where SVG files exported from Penrose contained no metadata necessary for importing
* Moved all the logic for downloading SVG, SVG TeX, PDF, and PNG files into `downloadUtils.ts` and `callbacks.ts` and out of `DiagramPanel.tsx`
* Added a new atom called `Canvas` to `atoms.ts` to track the canvasRef of the `DiagramPanel` in application state. Originally I tried to add the ref to the existing `Diagram` atom but it had to be removed to avoid an endless circular update pattern.
* Added dependencies to **JSZip and file-saver** packages, ran `npm audit fix`

To make these changes, I needed to refactor the code that was previously kept in `DiagramPanel` that related to downloading diagrams in SVG, SVG TeX, PDF, and PNG file formats. All of the logic that relies on grabbing from recoil state have been moved into the `callbacks.ts` file, and all of the other logic has been moved to `downloadUtils.ts`.

TODO

# Checklist

# Open questions
* ATM I have only separated out the option to "save as Penrose SVG" on the header. Should I also include the "SVG (TeX)" file format option there or leave it in the export dropdown?


package.json
```diff
@@ -52,5 +52,9 @@
       "**/babel-loader",
       "**/webpack"
     ]
+  },
+  "dependencies": {
+    "file-saver": "^2.0.5",
+    "jszip": "^3.10.1"
```
hsharriman, line 58: Should these have been added to devDependencies instead?

wodeni: I don't think it matters as much for `@penrose/editor`, but I think they should be in `dependencies`, as they are included in the final bundle in production.

wodeni (approved):
Looks great! Thanks for going through the troubles to refactor things. Much better organized now. I like how prominent the export options are now. 

A small question is whether material UI is being tree-shaken properly and affects startup time, definitely a future investigation.

wodeni:
Found one bug: downloading Penrose SVG will cause pdf download to fail like this:
image

PR #1687, opened 2023-12-19
fix: improper handling of flipped index ranges

### LINKED ISSUE #1685: Incorrect handling of indexed sets with flipped ranges
**Describe the bug**

When a substance indexed set has flipped ranges (e.g., `for i in [35, 32]`), the range should contain zero elements. But it seems to be not the case.

**To Reproduce**

https://penrose.cs.cmu.edu/try/?gist=5dcb8754142fa595cbfc52bb16368c41

**Expected behavior**

I expect nothing to be displayed on the canvas.

**Screenshots**

liangyiliang:
# Description

Resolves #1685.

This PR adds a check to `evalISet` - for a range `i in [low, high]`, if `high < low`, then return empty set since there are no valid `i`s.

We modified the test `indexed set decl - no satisfying indices`. The modified test would fail in the current version but will pass in this PR.

 
# Checklist

### DIFF
--- a/packages/core/src/compiler/Substance.test.ts
+++ b/packages/core/src/compiler/Substance.test.ts
@@ -304,7 +304,7 @@ NoLabel B, C
       }
 
       const env2 = envOrError(domainProg);
-      const prog2 = `Set a_i for i in [11, 10]`;
+      const prog2 = `Set a_i for i in [20, 10]`;
       const res2 = compileSubstance(prog2, env2);
       expect(res2.isOk()).toBe(true);
       if (res2.isOk()) {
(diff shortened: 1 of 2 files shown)


wodeni (changes requested):
Thanks for the fix! I should have documented the expected behavior tbh, and i’m okay with the one in this PR. one small request: can you add this case to the docs? Also, I imagine an empty set $\\times$ any list is just empty set? I don’t remember having a test case for this but it’s probably good to test that out

liangyiliang:
> can you add this case to the docs

It has already been documented! 

image

The previous implementation does not respect the docs.

liangyiliang:
> Also, I imagine an empty set 
> ×
>  any list is just empty set? I don’t remember having a test case for this but it’s probably good to test that out

Will add a test.

wodeni:
> > can you add this case to the docs
> 
> It has already been documented!
> 
> image
> 
> The previous implementation does not respect the docs.

Ahh awesome! I totally missed the example after the explanation haha.

wodeni (approved):
Looks great! Thanks for the speedy fix

Issue #1695, opened 2024-01-12
`saveWorkspaceEffect` in editor functionality should be split based on whether state update should be immediate or debounced


hsharriman:
**Describe the bug**
A clear and concise description of what the bug is. If there is an error in the inspector pane, please paste that string in.

**To Reproduce**
Please include what repository you are working from (e.g. if it is a fork, include URL), the branch of the repository (if not `main`), and the specific commit where this occurs.

Please also include the program triple (Domain, Substance, Style) that triggers the error. Specific line number(s) are very helpful. If you open the diff of a commit on GitHub and click the line number(s) on the left, it will generate a URL linking directly to that code. ([Example](https://github.com/srossd/penrose/commit/4a8574a4bcad94d5773e4a7c8dfc70c180cdbc46#diff-e3f90ad59d8a07aa564e67ef3a7ca3e72aed7f28e946545a651aac3e8da8227dR352))

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information if applicable):**

**Smartphone (please complete the following information if applicable):**

**Additional context**

Issue #1683, opened 2023-12-15
Multiple bugs related to tracking state changes surrounding saving drafts and workspaces in the editor


hsharriman:
**Describe the bug**
- renaming a workspace should not automatically create a new workspace if you're already working with a workspace
- importing: add confirmation if you are trying to import a file while in draft state
- importing: trio is imported but the title/top header does not update
	- if you share imported trio the stale title will be included
- creating a gist from a draft does not save the changes locally but it does encapsulate the changes in the gist, it should also save your changes as a workspace
- there's no confirmation that you're abandoning work if you navigate to someone else's gist while in draft mode
- there's no confirmation that you're abandoning work if you navigate to new webpage while in draft mode
- if you import a file while viewing a workspace, imported files overwrite your workspace's changes instead of saving your workspace and putting you in draft state
- FLAKY bugs (not a consistent repro): 
	- upload is flaky? sometimes it takes like 5 or 6 tries for the upload to go through. no confirmations/indicators appear in this situation
	- duplicating workspaces too quickly causes errors in retrieving those workspaces

**To Reproduce**
These behaviors repro on main

Please also include the program triple (Domain, Substance, Style) that triggers the error. Specific line number(s) are very helpful. If you open the diff of a commit on GitHub and click the line number(s) on the left, it will generate a URL linking directly to that code. ([Example](https://github.com/srossd/penrose/commit/4a8574a4bcad94d5773e4a7c8dfc70c180cdbc46#diff-e3f90ad59d8a07aa564e67ef3a7ca3e72aed7f28e946545a651aac3e8da8227dR352))

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information if applicable):**

**Smartphone (please complete the following information if applicable):**

**Additional context**
Too many bugs to include repro instructions for all at the moment.

hsharriman:
changes:
* couldn't repro on 1/10/24 ~renaming a workspace should not automatically create a new workspace if you're already working with a workspace~
* **fixed**. importing: add confirmation if you are trying to import a file while in draft state
* importing: trio is imported but the title/top header does not update
  * (most likely related): if you share imported trio the stale title will be included
  * add callback to upload step that updates the header title
* creating a gist from a draft does not save the changes locally but it does encapsulate the changes in the gist, it should also save your changes as a workspace
* try adding a beforeunload event that confirms that you have no unsaved changes before navigating away from the page
[...]

PR #1696, opened 2024-01-12
fix: resolve state-related bugs in the editor

hsharriman:
# Description

Resolves the non-flaky bugs in #1683. 

(fyi: draft state = user has unsaved changes that are not tracked in a workspace. currently, this is visually indicated by the presence of a "save" button on top left side of the TopBar)
Many changes related to importing a Penrose SVG to the editor:
* added confirmation if you are trying to import a file while in draft state
* changed behavior so that the the title/top header matches the imported file name instead (previously, the title would not update at all when the file was imported)
  * b/c of above change, if you share imported trio the new title will be included
* importing a file now puts user in draft state. This fixes a bug where if you imported a file while viewing a workspace, the imported files would overwrite the workspace's changes.
* fixed bug where if you imported a trio, all the ".domain/.substance/.style" tabs would change to  "SVG import"

Other fixes:
* fixed situation where creating a gist from a draft would not save the changes locally before you're automatically redirected to the gist. Changed behavior so that the draft will be saved to a new workspace before being redirected.
* added a beforeunload event that confirms that you have no unsaved changes before navigating away from the page, addressing situations where:
  * there's no confirmation that you're abandoning work if you navigate to someone else's gist while in draft mode
  * there's no confirmation that you're abandoning work if you navigate to new webpage while in draft mode

I decided not to handle the flaky bugs in this PR because I believe they will require more involved changes to the recoil state, refer to issue #1695 for more information. 

How to get into draft state:
1. go to Examples -> make a change to any example diagram -> the blue "save" button should appear next to the diagram title

Testing changes that involve discarding a draft:
1. from draft state -> go to "upload" -> upload any Penrose SVG -> should receive a confirmation toast warning about unsaved changes
2. from draft state -> click "share" (make sure you're signed into github so you can create a gist) -> should redirect to new gist page -> press "back" button in your browser -> click "saved" tab -> draft should be listed in your workspaces
3. from draft state -> type a new website URL into your browser (i.e. google.com) -> should receive confirmation toast warning about unsaved changes

Testing changes involving importing:
1. click "upload" -> import any Penrose SVG" -> notice that:
  * "save" button should appear next to diagram title indicating that you're in  draft state
  * the diagram title should match the filename of your imported SVG
2. click "saved" -> click any of your existing workspaces -> click "upload" -> import any Penrose SVG -> click "saved" again -> notice that the workspace you had previously selected remains unchanged
 
# Checklist

# Open questions


wodeni:
Just went through the example cases and all seemed good. There's a very noticeable delay after every keystroke. I'm guessing this has something to do with the debounced function?

wodeni:
After some more tinkering, looks like commenting out the `beforeunload` event listener will get rid of the delay. This effect should only trigger once on load, right? If so, `useEffect` should take `[]` as the second argument. I made this change locally, and the delay goes away, but once I switch to another example it comes back. Very mysterious.

hsharriman:
The beforeunload event handler is causing unexpected lag throughout the event handlers in the editor. For now, opted to remove the event handler and reimplement in a future PR.

PR #1700, opened 2024-01-15
feat: Update variation for caffeine example

keenancrane:
> The caffeine example is laughable.

This PR fixes the caffeine example to be less laughable.


wodeni (approved):
Thanks for the quick patch @keenancrane!

PR #1701, opened 2024-01-16
fix: vector subtraction in tutorial3.md

zqm980904:
# Description

Vector subtraction should involve adding the origin, not subtracting. It will cause an offset when `vec2 U.origin` is not `(0., 0.)`.


wodeni:
Good catch @zqm980904! Thanks for contributing to Penrose :P. If you haven't done so, join our [discord server](https://discord.gg/a7VXJU4dfR) to chat with us about other cool things to do in Penrose!

PR #1714, opened 2024-02-15
fix: exclude function body from `FunctionInternalError`

### LINKED ISSUE #1713: Calling `random` with varying or computed values results in uncaught errors in the IDE
**Describe the bug**

When the Style program calls `random` with varying or computed values as arguments, the IDE end up spinning forever due to a worker error with cloning.

**To Reproduce**

[arrays_9.zip](https://github.com/penrose/penrose/files/14300089/arrays_9.zip)

wodeni:
# Description

Resolves #1713.

When there's an internal error in computation functions, the compiler adds the entire function including its body as metadata for the error. This is fine on the main thread but not ok in webworkers because functions are not clonable. This PR omits the function body in the `FunctionInteralError` type.

### DIFF
--- a/packages/core/src/compiler/StyleFunctionCaller.ts
+++ b/packages/core/src/compiler/StyleFunctionCaller.ts
@@ -46,7 +46,17 @@ export const callCompFunc = (
     });
   } catch (e) {
     if (e instanceof Error) {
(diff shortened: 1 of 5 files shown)


liangyiliang (approved):
Interesting, thanks!

PR #1728, opened 2024-04-01
refactor: commented geometry domain

rjainrjain:
# Description

This PR adds comments to each type, constructor, function, and predicate in the geometry domain, explaining its usage with a description and example program. For example, for a `Segment`:

```
constructor Segment(Point p, Point q) -- This constructor creates a line segment from two points. Example usage: 
-- Point A, B
-- Segment AB := Segment(A, B)
-- AutoLabel A, B
```

# Checklist


wodeni (approved):
Thanks for adding the comments! Probably good to add a page on [our site](https://penrose.cs.cmu.edu/docs/ref) for every (stable) diagramming domain as well.

PR #1763, opened 2024-05-30
feat: Added copy trio to clipboard feature

### LINKED ISSUE #1761: Copy trio to clipboard (editor UX)
Suppose a user wanted to copy their trio code for reference (creating a GitHub issue for example). Currently they'd have to copy from each trio tab, tabbing between each and copy and pasting each, which seems inconvenient and time consuming. 

Under the export dropdown menu add a "Copy to Clipboard" option. On click it would copy all three parts of the trio (comment delineated) to the user's clipboard.

KyleleeSea:
# Description

Resolves #1761 

Previously if a user wanted to copy a trio to their clipboard they would have to copy and paste each piece individually which feels time consuming and clunky. 

As a solution, I added a "Copy to Clipboard" option beneath the Export button in the ide. Clicking this option copies each element of the trio separated by a -- .(file name)

### DIFF
--- a/packages/editor/src/state/callbacks.ts
+++ b/packages/editor/src/state/callbacks.ts
@@ -222,6 +222,21 @@ export const useDownloadTrio = () =>
     }
   });
 
+export const useCopyToClipboard = () =>
+  useRecoilCallback(({ set, snapshot }) => async () => {
+    const workspace = snapshot.getLoadable(currentWorkspaceState)
+      .contents as Workspace;
+    const dsl = workspace.files.domain.contents;
+    const sub = workspace.files.substance.contents;
+    const sty = workspace.files.style.contents;
+    const concatenated = `-- .substance \n ${sub} \n 
+    -- .style \n ${sty} \n 
+    -- .domain \n ${dsl} \n`;
+
+    navigator.clipboard.writeText(concatenated);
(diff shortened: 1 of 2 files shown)


packages/editor/src/state/callbacks.ts
```diff
@@ -222,6 +222,27 @@ export const useDownloadTrio = () =>
...
   });
 
+export const useCopyToClipboard = () =>
+  useRecoilCallback(({ set, snapshot }) => async () => {
+    const workspace = snapshot.getLoadable(currentWorkspaceState)
+      .contents as Workspace;
+    const sub = workspace.files.substance.contents;
+    const sty = workspace.files.style.contents;
+    const dsl = workspace.files.domain.contents;
+    const concatenated = `-- .substance \n ${sub} \n
```
wodeni, line 232: This formatting actually introduces extra spaces (see below). I think it'll be okay if you remove the spaces around all new lines
image

wodeni (changes requested):

packages/editor/src/state/callbacks.ts
KyleleeSea: Thanks for the catch! Fixed & tested in latest commit.

wodeni (approved):
LGTM! Thanks for the quick fix :P

PR #1764, opened 2024-05-30
feat: Making gist sharing clearer (editor ux)

### LINKED ISSUE #1760: Make gist sharing clearer (editor UX)
Currently when a user clicks share, it redirects them to a Penrose editor page with the gist id as a URL parameter which acts as a sharable link. However, from a UX standpoint, it is not clear that this page is sharable, until I looked at the codebase I thought it was supposed to redirect to a Github gist link. 

1. Copy link to clipboard on share
2. Toast notification "Sharable link copied to clipboard!" or something similar

KyleleeSea:
# Description

Resolves #1760

Problem: Previously, it was not clear that the link the share button redirected the user to is a shareable link. 
Solution: Once the user clicks share, the shareable link is automatically copied to the user's clipboard and after redirecting, a notification saying "Copied shareable link to clipboard" is shown. See video below for behavior. 

https://github.com/penrose/penrose/assets/85892844/288a4780-86e3-4192-be96-ed75597581ff

The copy link and display notification is triggered by a url parameter. The way the share button flow works is once the user clicks share, they are redirected to a page with the gist id as a url parameter. Since we don't have explicit routing setup, the only way to pass state from the page the share button was clicked on to the page the user is redirected to is by adding a query parameter.

The downside of this, is that if the user copies the link manually that link will include the copy to clipboard query parameter so anyone who clicks that link will have the link copied to their clipboard and see the notification even if they weren't the ones who clicked share.

I do not see this as a major disadvantage, since the notification text isn't specific to sharing, this behavior won't feel like a bug.

### DIFF
--- a/packages/editor/src/state/callbacks.ts
+++ b/packages/editor/src/state/callbacks.ts
@@ -678,8 +678,18 @@ export const usePublishGist = () =>
       toast.error(`Could not publish gist: ${res.statusText} ${json.message}`);
       return;
     }
-    toast.success(`Published gist, redirecting...`);
-    window.location.search = queryString.stringify({ gist: json.id });
+    const queryParameter = queryString.stringify({ gist: json.id });
+    const redirectURL = `https://penrose.cs.cmu.edu/try/?${queryParameter}`;
+    navigator.clipboard
+      .writeText(redirectURL)
+      .then(() => {
+        toast.success("Copied sharable link to clipboard");
+      })
+      .catch(() => {
+        toast.success("Redirecting to gist...");
+      });
+
+    window.location.search = queryParameter;
   });
 
 const REDIRECT_URL =
--- a/packages/editor/src/state/callbacks.ts
+++ b/packages/editor/src/state/callbacks.ts
@@ -563,6 +563,15 @@ export const useCheckURL = () =>
         files,
       };
       set(currentWorkspaceState, workspace);
+
+      // Notification + save to clipboard if redirected from clicking share
+      if ("pub" in parsed) {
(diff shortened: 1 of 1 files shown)


wodeni:
> The downside of this, is that if the user copies the link manually that link will include the copy to clipboard query parameter so anyone who clicks that link will have the link copied to their clipboard and see the notification even if they weren't the ones who clicked share.

Reproduced this by visiting: https://a7fca545.penrose-72l.pages.dev/try/?gist=5e82ed0dfb3066dccc4f26bf38b17331&pub=true. 

IMO this is okay for the same reason

samestep (approved):
This makes a ton of sense, thanks Kyle!

samestep:
> The downside of this, is that if the user copies the link manually that link will include the copy to clipboard query parameter so anyone who clicks that link will have the link copied to their clipboard and see the notification even if they weren't the ones who clicked share.

Is there a way to not do this?

KyleleeSea:
> > The downside of this, is that if the user copies the link manually that link will include the copy to clipboard query parameter so anyone who clicks that link will have the link copied to their clipboard and see the notification even if they weren't the ones who clicked share.
> 
> Is there a way to not do this?

Found a way to fix this actually! Latest commit made two changes:
1. The pub parameter is automatically removed from the URL upon load (there is a brief moment where it's visible but its very very short)
2. Removed the "loading gist..." notification when a user clicks share and loads since before this was a bounce of 3 notifs (redirecting, loading, copied) which looked bad

Behavior:
https://github.com/penrose/penrose/assets/85892844/03c6a619-7a99-4368-b95b-765f4159f4ca

PR #1765, opened 2024-05-30
refactor: OptimizerWorker/worker

griffinteller:
# Description

Parallelism is necessary to ensure a responsive webpage during diagram optimization, which requires the use of HTML5 web workers. Communicating with a worker can be done only through immutable messages or shared flags. This process had been previously encapsulated by the class `OptimizerWorker`, which has clearer semantics than just message passing (e.g, methods such as `compile` or `resample`). However, the implicit state machine was unclear an unsafe, leading to occasional hard-to-reproduce bugs (e.g, hanging 'compile' notification), and augmenting the behaviors of both the encapsulation and the worker (for interactivity, in my case) seemed difficult to achieve in an acceptably non-hacky way.

To remedy this, I've refactored both `OptimizerWorker.ts` and `worker.ts` into a much stricter state machine with clearer request and response semantics. This is not perfect, and quite repetitive and likely too complicated in some parts. But it is more extensible, and shouldn't be difficult to simplify.

There are also small changes to some React components and callbacks to accomodate a slightly modified interface.

## State Machine (abbreviated)
Screenshot 2024-05-30 at 12 29 20 PM

Omitted above are states `WaitingForUpdate` and `WaitingForShapes`, which are moved to after a call to `pollForUpdate` or `computeShapesAtIndex`, and moved out of with an `UpdateResp`. These can likely be merged into one state and one request.

## Design decisions

- States: Represented as discriminated union of types, each of which contain relevant state information. Prevents accessing data in states where it may not be valid, since this wouldn't type check. On the worker side, the three states are represented as an enum for simplicity.
- Responses/Requests: ditto
- `OptimizerWorker` interface: Before, each function in the interface took in some subset of callbacks `onUpdate`, `onComplete`, and `onError`. Really, handling the callbacks should not be OptimizerWorker's job, and the semantics are not clear (is `onComplete` called on the completion of the method, or optimization? What about `onUpdate`?). Instead, I opted to use async functions and promises wherever possible. This allows the caller to use the builtin `await`, `.then`, and `.catch` constructs should they feel like it. ~~The only exception to this is the `onFinish` callback given to `startOptimization`, since it is more consistent with the rest of the interface to resolve the returned promise when the method has completed. It might be an option to return two promises: one for the start of optimization, and one for the finish.~~ Done!
- Optimization interrupting: Before, we had to manage shared memory to interrupt optimization. This is not necessary, and prevents sending information with an interrupt (say, a step index to compute shapes with). Instead, we can just queue the next optimization step at the back of the JS engine's task queue via
```ts
while (!isOptimized(state)) {
    /* queues resolve as next in the event queue, after onmessage if a message
     has been received. await-ing `optStep` is not enough, since this will queue optStep
     as a _microtask_, which will still run before onmessage */
    await new Promise<void>(resolve => {
      setTimeout(resolve, 0);
    });

    if (shouldFinish) {  // set by onmessage if we need to stop
      log.info('Optimization finishing early');
      shouldFinish = false;
      break;
    }

    optStep();  // run an optimization step
  }
```

If a message is received, it will be queued in the task queue. Then, on awaiting our new promise, the JS engine will execute `onmessage` _before_ resolving our promise. The `setTimeout(resolve, 0)` line is what actually queues `resolve` into the task queue. Otherwise, the promise is queued on the higher priority _microtask_ queue, which is emptied before  executing the next task (potentially onmessage). [More info](https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide/In_depth)

## `OptimizerWorker` Interface

```ts
isInit(): boolean
```

Returns whether the state is no longer `WaitingForInit`.

```ts
async waitForInit(): Promise<void>
```
Returns promise which resolves when the the state is no longer `WaitingForInit`. Used by react callbacks.

---

Unless otherwise specified, the following functions will wait until they are in a valid state before executing. 

```ts
async compile(
    domain: string,
    style: string,
    substance: string,
    variation: string
): Promise<string>
```
- In state `Init`: compiles diagram (blue and gray paths above)
- In state `Compiled`: recompiles diagram (omitted above: moves back to state `InitToCompile` and compiles)
- In state `Optimizing`: interrupts optimization (InterruptReq -> FinishedResp) and then recompiles

Returns task id

```ts
async startOptimizing(): Promise<{ onStart: Promise<void>, onFinish: Promise<UpdateInfo> }: OptimizerPromises }>
```

- In state `Compiled`: begins optimization (red and yellow paths). 

Returns two promises: `onStart` for when optimization has begun, and `onFinish` for when optimization has ended. If `onStart` rejects, `onFinish` will neither resolve nor reject.

```ts
async resample(
    jobId: string,
    variation: string,
): Promise<{ onStart: Promise<void>, onFinish: Promise<UpdateInfo> }: OptimizerPromises }>
```

- In state `Compiled`: requests worker resamples varying values and begins optimization (omitted above) 
- In state `Optimizing`: interrupts optimization, then resamples

Returns two promises: `onStart` for when optimization has begun, and `onFinish` for when optimization has ended. If `onStart` rejects, `onFinish` will neither resolve nor reject.

```ts
async pollForUpdate(): Promise<UpdateInfo | null>
```

- In states `Compiled`, `Optimizing`: polls for an update from worker (omitted above). 

Returns `UpdateInfo` on worker response, or null if the diagram has not changed.

```ts
async computeShapesAtIndex(
    i: number
): Promise<RenderState>
```

- In states `Compiled`, `Optimizing`: polls for the shapes at step `i`.

Returns `RenderState` from the appropriate optimization step. Probably this function can be merged with `pollForUpdate` to simplify.

## Misc Bug Fixes

- Timeline slider no longer throws NaNs or falls behind
- "Compiling..." and "Resampling..." hang around forever less often (hopefully never)

# Checklist

# To Do/Questions

- Code is quite repetitive in parts (a bunch of nasty switch, very similar behavior between many interface methods). Definitely much of this can and should be abstracted.
- Documentation 🤡 
- ~~Gracefully handling errors in `worker.ts`. Right now these are not sent back or caught, but should be. Before the were sent back in an error response, which makes sense. But we'll need to figure out what the desired behavior is. Probably if an error is thrown during optimization it is fine to return to the `Compiled` state? Similarly if an error is thrown in `Compiled` we can return to `Init`. And for each waiting state we can determine what might be best.~~ Done!


packages/editor/src/state/atoms.ts
```diff
@@ -20,8 +20,8 @@ import {
 } from "recoil";
 import { v4 as uuid } from "uuid";
 import { layoutModel } from "../App.js";
-import { RenderState } from "../worker/message.js";
-import OptimizerWorker from "../worker/OptimizerWorker.js";
+import { RenderState } from "../worker/common";
+import OptimizerWorker from "../worker/OptimizerWorker";
```
samestep, line 24: Looks like the `.js` suffix got lost here; @wodeni I forget, do we still want to use this? I feel like I remember having some way to automatically lint this in the past, but we got rid of it for some reason.

packages/editor/src/state/callbacks.ts
```diff
@@ -84,34 +86,50 @@ const _compileDiagram = async (
...
+  const onError = (error: any) => {
+    toast.dismiss(compiling);
+    toast.error(error.message);
+    set(diagramWorkerState, {
+      ...diagramWorkerState,
+      optimizing: false,
+    });
+  };
+
+  // ugly `then` chain allows for one catch at the end
```
samestep, line 98: I see that a lot of your usage code uses `.then`/`.catch` instead of `async`/`await`; is this necessary? Is there no way to write this using normal `async`/`await` code?

samestep (approved):
This makes a ton of sense, thanks for doing this refactor. I don't remember why we didn't build a `Promise`-oriented API in the first place, but I suspect it was probably just because we got too caught up in the details of what primitives are allowed when using Web Workers.

I haven't thought about the design too hard yet, but the idea of microtasks feels a bit strange; do we just need these specifically because the worker is ultimately still one thread? I'm thinking about other message-passing concurrency architectures I've worked with, and I feel like the more elegant solution might be to have more lightweight queues/channels in between things to handle prioritization while the heavy worker thread is running. Curious to hear your thoughts @griffinteller!

But that's all just musings; regardless, this PR seems like a great improvement from the status quo.

packages/editor/src/state/callbacks.ts
griffinteller: In some parts the `thens` are necessary because I don't want to synchronously wait for the promise to resolve, I just want to attach a callback. In this particular section I was being silly--I wanted to save on catches for every single promise, but really I think you're right that it is more readable to do a bunch of awaits and then `try`/`catch` around that. I'll fix that in a bit.

griffinteller:
> This makes a ton of sense, thanks for doing this refactor. I don't remember why we didn't build a `Promise`-oriented API in the first place, but I suspect it was probably just because we got too caught up in the details of what primitives are allowed when using Web Workers.
> 
> I haven't thought about the design too hard yet, but the idea of microtasks feels a bit strange; do we just need these specifically because the worker is ultimately still one thread? I'm thinking about other message-passing concurrency architectures I've worked with, and I feel like the more elegant solution might be to have more lightweight queues/channels in between things to handle prioritization while the heavy worker thread is running. Curious to hear your thoughts @griffinteller!
> 
> But that's all just musings; regardless, this PR seems like a great improvement from the status quo.

If you're talking about the `setTimeout` trick in the worker, the details of tasks and microtasks don't matter/aren't relied on, the comment was just a note about why this works. The goal is to yield the optimization loop execution to `onmessage` if anything has been recieved; unfortunately `await`-ed promises are given higher priority than events. `setTimeout(/* func */, 0)` queues the function as an event with the same priority as the message, so get queued by JS behind `onmessage`, if there are any pending messages.

PR #1770, opened 2024-05-31
feat: join paths

### LINKED ISSUE #1767: feat: Add functionality to join paths
### Issue

There are many situations in which it is necessary to concatenate multiple paths into a single path.  A common case is connecting several open paths into a single closed path that can be filled.  Consider for instance this figure, where each of the circular arcs is drawn as a `Path` shape using the `circularArc` function:

image

Suppose we now want to fill the region bounded by the three arcs with a solid color.  Currently there is no way to do this in Style.

### Proposal

A straightforward solution is to implement Style functions that can join path data into a new path.  There are several possibilities for how such a function might work.  For instance:

* `joinPaths( [ p1, p2, … ] )` — takes a list of path data instances $p_1$, $p_2$, … and produce a single path where the final point in each path $p_i$ has been replaced with the first point in path $p_{i+1}$.  This function should perhaps also take an `"open"/"closed"` flag which determines whether the final vertex is included in the path description (since a closed SVG path will automatically connect back to the initial point).
* `connectPaths( [ p1, p2, … ] )` — takes a list of path data instances $p_1$, $p_2$, … and produce a single path where the final point in each path $p_i$ is connected to the first point in path $p_{i+1}$ by some connecting geometry (by default, a straight line segment).
* `concatenatePaths( [ p1, p2, … ] )` — takes a list of path data instances $p_1$, $p_2$, … and produce a single path by just concatenating the raw path commands.

Alternatively, we could just have a single function `joinPaths` that automatically determines whether endpoints need to be joined or not depending on whether or not they coincide (perhaps within some user-specified floating point error tolerance).

**Example usage.**

Here's a mockup of usage for the diagram above:

```
shape arc1 = Path {
   d: circularArc( "open", c1, r1, alpha1, beta1 )
}
shape arc2 = Path {
   d: circularArc( "open", c2, r2, alpha2, beta2 )
}
shape arc3 = Path {
   d: circularArc( "open", c3, r3, alpha3, beta3 )
}
shape region = Path {
   d: joinPaths( "closed", [ arc1, arc2, arc3 ] )
   fillColor: blue;
}
```

griffinteller:
# Description

**Draft**

(Partially) resolves #1767.

This implements two style functions, `concatenatePaths` and `connectPaths` which both turn a list of `PathDataV` into one `PathDataV`, according to different connection policies

- Added `PathDataListV` as a style list type
- New style functions mirror the format of existing path factory functions
## New Functions
```
concatenatePaths : (PathDataListV) => PathDataV
```
Connect a list of paths by exactly concatenating their SVG commands. Visually, this has the effect of taking the union of the paths.

```
connectPaths : (PathTypeV, PathDataListV) => PathDataV
```
Connect a list of paths by drawing a line between the end and start point of each path. If "closed" is passed as the path type, a line is drawn from the end of the last path to the start of the first.

```
joinPaths: (PathDataListV) => PathDataV
``` 
Connect a list of paths by removing the beginning 'M' command from all but the first path. This has the effect of closing the path, allowing for filling, etc. However, the start and end points of each path must already coincide to give correct results.

# Examples 

Screenshot 2024-05-31 at 7 33 01 PM

Screenshot 2024-05-31 at 7 33 29 PM

# Checklist

# Open questions/Next steps

- ~~A `joinPaths` function where each start point is made to coincide with the previous endpoint would be an obvious next step, as Prof. Crane suggested. But since certain path coordinates may be absolutely specified, we likely couldn't use relative SVG commands. Additionally, it's not clear what should happen if a start point is absolutely specified to be different from the previous endpoint. But it wouldn't be hard to just delete the 'M' command from the beginning of each path, and make the coincidence of start and end points a precondition for the function.~~ implemented, as above.

### DIFF
--- a/packages/core/src/compiler/Style.ts
+++ b/packages/core/src/compiler/Style.ts
@@ -125,6 +125,8 @@ import {
   LListV,
   ListV,
   MatrixV,
+  PathDataListV,
+  PathDataV,
   PropID,
   PtListV,
   ShapeListV,
@@ -170,6 +172,8 @@ import {
   listV,
   llistV,
   matrixV,
+  pathDataListV,
+  pathDataV,
   prettyPrintResolvedPath,
   ptListV,
   resolveRhsName,
@@ -2674,6 +2678,22 @@ const evalShapeList = (
   return ok(shapeListV(elems));
 };
 
+const evalPathDataList = (
+  coll: List<C> | Vector<C> | CollectionAccess<C>,
+  first: PathDataV<ad.Num>,
+  rest: ArgVal<ad.Num>[],
+): Result<PathDataListV<ad.Num>, StyleDiagnostics> => {
+  const elems = [first.contents];
+  for (const v of rest) {
+    if (v.tag === "Val" && v.contents.tag === "PathDataV") {
+      elems.push(v.contents.contents);
+    } else {
+      return err(oneErr({ tag: "BadElementError", coll, index: elems.length }));
+    }
+  }
+  return ok(pathDataListV(elems));
+};
+
 const evalListOrVector = (
   mut: MutableContext,
   canvas: Canvas,
@@ -2707,11 +2727,14 @@ const evalListOrVector = (
           case "TupV": {
             return eval2D(coll, first.contents, rest);
           }
+          case "PathDataV": {
+            return evalPathDataList(coll, first.contents, rest);
+          }
+          case "PathDataListV":
           case "BoolV":
           case "ColorV":
           case "LListV":
           case "MatrixV":
-          case "PathDataV":
           case "PtListV":
           case "StrV":
           case "ShapeListV":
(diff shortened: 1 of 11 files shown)


wodeni (commented):
Thanks for the PR! The changes seem good to me. It's a bit annoying that we need to enumerate these Style data types in this many places. 

Related to this PR tho: would be great to add some tests in, say, `Functions.test.ts` as sanity check. You can probably use [`toPathString`](https://github.com/penrose/penrose/blob/61f2ad766efc471bf552a6cfc3d9fcf9c2f3779a/packages/core/src/renderer/Path.ts#L19) for pretty-printing paths so the checking is a bit easier.

griffinteller:
> Thanks for the PR! The changes seem good to me. It's a bit annoying that we need to enumerate these Style data types in this many places.
> 
> Related to this PR tho: would be great to add some tests in, say, `Functions.test.ts` as sanity check. You can probably use [`toPathString`](https://github.com/penrose/penrose/blob/61f2ad766efc471bf552a6cfc3d9fcf9c2f3779a/packages/core/src/renderer/Path.ts#L19) for pretty-printing paths so the checking is a bit easier.

Tests are passing, so I'm gonna go ahead and merge. Thanks for the review!

Issue #1788, opened 2024-06-10
bug: Can't use a `Path` as a clip mask in a `Group`


keenancrane:
# Issue

_[Note: this issue seems to have a one-line fix; see below.]_

When trying to use a `Path` as a clip mask, the clipped shape does not appear.  Here's an example program:

```haskell
canvas {
  width = 400
  height = 400

  shape C = Circle {
    center: (0,0)
    r: 100
    fillColor: #ff0
    strokeColor: #f00
    strokeWidth: 1
  }
  shape P = Path {
    d: interpolatingSpline("closed", [ (112.508,119.145), (-34.327,78.639), (-112.209,-17.934), (4.291,-47.19), (104.491,-118.829), (80.272,12.394) ] )
      fillColor: #0a0
      strokeColor: #00f
      strokeWidth: 1
  }
  
   -- use the path as a clipping mask for the circle
   shape G = Group {
     shapes: [C]
     clipPath: clip(P)
   }
}
```

Basically this green shape should be used as a mask for the yellow circle:

![shapes](https://github.com/penrose/penrose/assets/3604525/246f6ef0-7ee5-4bcc-ad50-9d63df697616)

But when we run the program, the canvas is blank.

# Solution

Inspecting the SVG file, it seems that the Path (P) is wrapped in group tags `<g> … </g>`.  These group tags are not generated by the Penrose renderer for other types of clipping masks (e.g., if we use a Rectangle rather than a Path for the clipping mask).

```svg
<?xml version="1.0"?>
<svg xmlns="http://www.w3.org/2000/svg" version="1.2" viewBox="0 0 400 400" width="100%" height="100%">
 <g>
   <clipPath id="myClip">
     <g> <!-- THIS GROUP TAG SHOULDN'T BE HERE -->
       <path fill="#00aa00" fill-opacity="1" stroke="#0000ff" stroke-opacity="1" stroke-width="1" stroke-linecap="butt" d="M 312.508 80.855 C 283.85825 64.29375000000002 221.85225 87.09125 165.673 121.361 C 109.49375 155.63075 78.1365 186.47675 87.791 217.934 C 97.4455 249.39125 150.11599999999999 221.96625 204.291 247.19 C 258.466 272.41375 285.49575 333.725 304.491 318.829 C 323.48625 303.933 278.26775 247.0995 280.272 187.606 C 282.27625 128.1125 341.15774999999996 97.41625 312.508 80.855 Z "/>
       <title>canvas.B</title>
[...]

PR #1789, opened 2024-06-10
feat: Arc mesh example

keenancrane:
# Description

arc-triangle-example

This PR adds an example for drawing meshes with arc edges, and annotating associated quantities (example above).

It also fixes issue #1788 (can't use path as clip mask), and removes some old crufty code from the path renderer (dumping unused CSS shadow effects for every path).


keenancrane:
The SVG code for a bunch of diagrams in the registry changed, but are visually identical.  This is basically just because for each for `<path>` element we dropped an unused shadow effect, and removed the grouping around paths without arrowheads.  (Checking the registry in fact helps verify that the logic here works as expected.)

Since these changes are rather minor, and seem to work fine, I'm going to go ahead and squash & merge.

PR #1794, opened 2024-06-12
feat: distribute horizontal and vertical constraint functions

apalliyil:
# Description

I added two additional constraint functions to the constraint function library. These functions, called distributeHorizontally and distributeVertically respectively, both take in a list of shapes and an optional padding margin as parameters. My original use case for this feature was for an icicle plot diagram I was making in Penrose. For context, icicle plots are one neat way to visualize file systems, or any general hierarchy-based system. Previously, I had to use the disjoint and touching functions in different instances to replicate what I was was envisioning. However, these constraints were difficult to balance and I ran into some consistency issues when resampling. This new constraint function simplified my work and allowed the more difficult alignments of my visualization to be computed instead of optimized. These two functions can be generalized to any collection of shapes. 

Both functions ensure that the bounding boxes of all shapes are touching each other by default, or separated by a margin specified by the padding. The padding parameter defaults to 0 if not specified. 

Here is a more in-depth look at how the functions work:

### distributeHorizontally

Require that the bounding boxes for `shapes` within a shape list are evenly spaced horizontally with optional padding `padding` . 

| Name     | Type | Type Description | Description | 
| ----------- | ----------- | ----------- | ----------- |
|      shapes        |      List of Shapes         |     Any list of shapes          |       Collection of shapes           |
| padding          |             ℝ                     |    Real Number        |         padding                                |

### distributeVertically

Require that the bounding boxes for `shapes` within a shape list are evenly spaced vertically with optional padding `padding` . 
| Name     | Type | Type Description | Description | 
| ----------- | ----------- | ----------- | ----------- |
|      shapes        |      List of Shapes         |     Any list of shapes          |       Collection of shapes           |
| padding          |             ℝ                     |    Real Number        |         padding                                |

---

For distributeHorizontally, the function works by iterating over all shapes in the list (collector) and summing the absolute values of the differences between the  sum of the right side of the current box plus padding and the left side of the next box. 

The optimization function that Penrose then solves is 

$$ \sum_{i=1}^{n-1} | (right_i + padding) -  left_{i+1} | = 0  $$

For distributeVertically, the function works similarly, but with different bounding box sides. Instead of the left and right bounds, the function uses the top and bottom bounds, summing the differences between the top bound of a current shape's bounding box and padding and the bottom bound of the next box's padding. 

The optimization function that Penrose solves here is 

$$ \sum_{i=1}^{n-1} | (top_i + padding) -  bottom_{i+1} | = 0 $$

distributeVertically example:

image

distributeVertically test link:  https://6ef225de.penrose-72l.pages.dev/try/?gist=dd0371a1f9d571beb756596fd3e44ad3

---

distributeHorizontally example:

image

distributeHorizontally test link: https://6ef225de.penrose-72l.pages.dev/try/?gist=94aae8ee3b747470fdbca682999b586a

---

file system example:

image

file system test link: https://6ef225de.penrose-72l.pages.dev/try/?gist=387d16129c47db9f81b2ae62cfc84260

---

# Checklist

# Open questions

### DIFF
--- a/packages/core/src/lib/Constraints.ts
+++ b/packages/core/src/lib/Constraints.ts
@@ -1063,7 +1063,7 @@ const constrDictGeneral = {
   hSpacing: {
     name: "hSpacing",
     description:
-      "Requires that the bounding boxes for shapes within a shapeList are touching.",
+      "Requires that the bounding boxes for shapes within a shapeList are evenly spaced with a set padding.",
     params: [
       {
         name: "shapes",
@@ -1071,15 +1071,21 @@ const constrDictGeneral = {
           "List containing the top-left, top-right, bottom-right, bottom-left points (in that order) of the axis-aligned bounding box of a shape",
         type: shapeListT(),
       },
+      {
+        name: "padding",
+        description: "Margin between bounding boxes of shapes",
+        type: realT(),
+        default: 0,
+      },
     ],
-    body: noWarnFn((shapes: Shape<ad.Num>[]) => {
+    body: noWarnFn((shapes: Shape<ad.Num>[], padding: ad.Num) => {
       let spacingSum: ad.Num = 0;
 
       for (let i = 0; i < shapes.length - 1; i++) {
         const currShape = BBox.maxX(bboxFromShape(shapes[i]));
         const nextShape = BBox.minX(bboxFromShape(shapes[i + 1]));
 
-        const difference = absVal(sub(currShape, nextShape));
+        const difference = absVal(sub(add(currShape, padding), nextShape));
 
         spacingSum = add(spacingSum, difference);
 
--- a/packages/core/src/lib/Constraints.ts
+++ b/packages/core/src/lib/Constraints.ts
@@ -1063,7 +1063,7 @@ const constrDictGeneral = {
   hSpacing: {
     name: "hSpacing",
     description:
-      "Requires that the bounding boxes for shapes within a shapeList are evenly spaced with a set padding.",
+      "Requires that the bounding boxes for shapes within a shapeList are evenly spaced horizontally with a set padding.",
     params: [
       {
         name: "shapes",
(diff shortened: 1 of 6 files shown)


liangyiliang:
Thanks for this - can you add some information into the PR description about what these two functions do? Perhaps go through some of the maths behind them? Thanks!~

liangyiliang (approved):
LGTM!

wodeni (commented):
Thanks for the PR @apalliyil! The code looks great to me. I have one suggestion on naming and one potential issue I observed:

* The behavior of these two functions are similar to "distribute" in WYSIWYG tools and [bluefish.js `<Distribute>`](https://bluefishjs.org/docs/reference/components/distribute). I suggest renaming the functions to `distributeV` and `distributeH`. Optionally we can consider longer names like `distrbuteVertically` and `distributeHorizontally`. Curious to hear @keenancrane's thoughts.
* A potential interaction with `ensureOnCanvas`. When the constraint forces shapes out of canvas, the behavior is not super consistent even with `ensureOnCanvas: false`. To repro, copy the style program below to the text test case, resample a bunch, and observe the inconsistent padding around the two middle elements:

```
\-- .substance
Text t1, t2, t3, t4

Paragraph p1

group(t1, p1)
group(t2, p1)
group(t3, p1)
group(t4, p1)



AutoLabel All


-- .style
canvas {
  width = 400
  height = 400
}


forall Text t {
  t.text = Text {
    name: t.label 
    center: (?, 0)
    ensureOnCanvas: false
  }
  t.bbox = Polygon {
    points: bboxPts(t.text)
  }
  
}


collect Text t into ts 
where group(t, p)
foreach Paragraph p {

  ensure hSpacing(listof text from ts, 100)
}


-- .domain
type Text

type Paragraph

predicate group(Text t, Paragraph p)

```

liangyiliang:
Hi @wodeni  - the inconsistency you pointed out is because of this:
```
  t.bbox = Polygon {
    points: bboxPts(t.text)
  }
```

Here, `t.bbox.ensureOnCanvas` is set to true. This is not a bug of the `hSpacing` contraint.

wodeni:
> Hi @wodeni  - the inconsistency you pointed out is because of this:
> 
> ```
> 
>   t.bbox = Polygon {
> 
>     points: bboxPts(t.text)
> 
>   }
> 
> ```
> 
> 
> 
> Here, `t.bbox.ensureOnCanvas` is set to true. This is not a bug of the `hSpacing` contraint.

Aha good catch! Only the naming suggestion remaining then 😀

PR #1796, opened 2024-06-12
feat: interactive widgets + programatically constrained dragging (experiemental)

griffinteller:
# Description

Adds interactivity as an experimental feature, in two forms:
- Edit Mode: Shapes which are translatable or scalable will have a widget appear on click, which can be dragged or scaled.
- Play Mode: Shapes which are translatable and have a 'draggingConstraint' field set will be draggable.

These modes can be toggled in the editor settings panel.

# Edit Mode

## Transformability indication

Objects which are either scalable or translatable will have their bounding box appear on hover:

https://github.com/penrose/penrose/assets/13922490/49c37dc9-df35-4860-b82c-896e9b375feb

## Translation

Clicking on a transformable object will bring up an "interactive widget". Dragging on the shape now will translate the object, if it is translatable:

https://github.com/penrose/penrose/assets/13922490/433da969-236f-4b89-b8ae-8cb7bcc090b0

## Scaling

If an object is scalable, four scaling boxes will appear on the widget which will scale the object on drag:

https://github.com/penrose/penrose/assets/13922490/a455ba6a-614b-4195-bdf0-77c9fb24c6b1

## Pinning 

When an object is transformed, it is automatically "pinned" on mouse up, preventing the optimizer from editing any of its attributes in the future. This is indicated by a red bounding box and transform widget. Right-clicking on a pinned widget will unpin it, and the attributes of the object will again be optimized:

https://github.com/penrose/penrose/assets/13922490/e8c3c23d-c49c-4c74-a6e4-4265961badf4

## Implementation details

Widgets are spawned by an `InteractivityOverlay`, which keeps track of pinned objects, hovered objects, and clicked objects. This spawns `InteractiveWidgets` and `HoverDisplay`s, which attach even listeners to themselves and to the svg in order to handle the appropriate mouse event. Upon an interaction, the widget sends an interaction request to the optimizer which transforms or pins the shape according the interaction in the optimizers state and restarts optimization from the new state.

Currently a shape is translatable if both of its `center` coordinates are optimized parameters, or in the case of a line, if all of its defining points' coordinates are optimized. A shape is scalable if its `width`/`height` or `rx`/`ry` parameters are optimized (and it is not text/equation), or if it is a circle and its `r` parameter is optimized. In the future, we might want to make this configurable from the style language.

# Play Mode

For interactivity meant for the viewer of the diagram (the _diagramee_, if you will), it is often the case that you would like to constrain dragging to some subset of the canvas (e.g, to slide along a curve or not exit a box). In these cases, you can express the constraint as a map from canvas space to canvas space: restricted to the legal positions of the object, this map is the identity map, and otherwise it forms some intuitive projection onto the space of legal positions.

 For example, to constrain a object to the circle $x^2 + y^2 = r^2$, one might express the constraint as 
$$(x, y) \mapsto r \frac{(x, y)}{\left\Vert(x,y)\right\Vert}$$

Say $r = 50$. In Penrose, you could then write

```
icon = Circle {
  --  ...
  draggingConstraint: "const norm = Math.sqrt(x * x + y * y); return [50 * x / norm, 50 * y / norm];"
}
```

`draggingConstraint` is an optional field on all translatable shapes which takes in one line of javascript. This line forms the body of a function with arguments `x: number` and `y: number`, corresponding to the mouse position, and must return a two-element array specifying the mapped position. Penrose will then apply this function before any dragging occurs. In the following diagram, it is impossible to drag the planets off their orbits.

https://github.com/penrose/penrose/assets/13922490/e7572252-0c90-4ccd-8919-4d5694ea1427

This javascript injection technique is for prototyping only, and suffers from several important limitations:
  - Constraints cannot use style variables, and so cannot vary with interaction or even use constants from style.
  - Constraints do not have access to any computation functions from penrose's library.

# Checklist

# Known issues

- You can't click and drag in one motion--you have to click to enable the widget, then click again to drag.
- Performance issues when CodeMirror is open

# Next steps

- History of interaction, both for undo behavior and saving/replaying. 
- Transformability configurable from style
- Widgets, probably at the bottom of the screen, such as sliders and radio boxes for parameters other than position and scale for which interaction might be desirable.


samestep:
This is incredible! Very minor initial comment @griffinteller: can you fix the merge conflict real quick?

.gitignore
```diff
@@ -8,4 +8,4 @@
 *~
 /target/
 dist/
-node_modules/
+node_modules/
```
samestep, line 11: Extraneous diff?

packages/components/src/Simple.tsx
samestep, line 1: Is the idea here just that now there is no `toInteractiveSVG` because you do interactivity via an overlay?

packages/core/src/index.ts
samestep, line 1: So @wodeni if I understand correctly, this isn't a breaking change because the `interactiveDiagram` function isn't documented in our public API? https://penrose.cs.cmu.edu/docs/ref/api

samestep (commented):
Really excited about this! Left a couple minor comments, happy to look in more detail at specific things and/or after the worker protocol is refactored. That should also help fix the weird "visual jittering" that currently happens for the history bar when dragging.

packages/components/src/Simple.tsx
griffinteller: Yeah, this is just old code that isn't relevant to how interactivity works now (entirely handled by the overlay). In fact I thought this might have already been deleted in the main branch? I guess not.

.gitignore
griffinteller: looks like it. doesn't show a diff locally.... ¯\_(ツ)_/¯

samestep: Are you using VS Code? I think there's a button you can toggle to display or hide whitespace changes.

packages/core/src/renderer/Renderer.ts
```diff
@@ -40,74 +38,26 @@ export interface RenderProps {
...
+  { screenX, screenY }: { screenX: number; screenY: number },
+  CTM: DOMMatrix | null,
 ) => {
-  const CTM = svg.getScreenCTM();
   if (CTM !== null) {
-    return { x: (clientX - CTM.e) / CTM.a, y: (clientY - CTM.f) / CTM.d };
+    return { x: (screenX - CTM.e) / CTM.a, y: (screenY - CTM.f) / CTM.d };
   }
   return { x: 0, y: 0 };
 };
```
wodeni, line 59: looks like this function is no longer used?

griffinteller: Yep, good catch.

packages/editor/src/utils/renderUtils.ts
```diff
@@ -25,8 +37,369 @@ export const stateToSVG = async (
...
+    const approxCenterX = bboxX + bboxW / 2;
+    const approxCenterY = bboxY + bboxH / 2;
+    const approxInitDeltaX = startX - approxCenterX;
+    const approxInitDeltaY = startY - approxCenterY;
+
+    const minX = startX - bboxX;
+    const maxX = state.canvas.width - bboxW + (startX - bboxX);
+    const minY = startY - bboxY;
+    const maxY = state.canvas.height - bboxH + (startY - bboxY);
+
```
wodeni, line 171: unused values?

wodeni (commented):
Done looking through everything other than `editor`. Looks great so far! Most comments are minor questions (that can probably be solved by one-liner comments) and removing dead code

packages/core/src/renderer/Renderer.ts
```diff
@@ -40,73 +38,8 @@ export interface RenderProps {
   texLabels: boolean;
   canvasSize: [number, number];
   pathResolver: PathResolver;
+  titleCache?: Map<string, SVGElement>;
```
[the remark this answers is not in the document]
griffinteller: These are used in `InteractivityOverlay` (line 68) to create a map from paths to svg elements. Otherwise, we would nee to loop through every single svg element every update to attach click listeners.

packages/editor/src/optimizer/worker.ts
```diff
@@ -126,47 +127,92 @@ const computeLayout = async (
...
   return result;
 };
 
+const recursivelyDeleteStepSequence = (sequenceId: StepSequenceID) => {
+  let currSeqId: StepSequenceID | null = sequenceId;
+  while (currSeqId !== null) {
+    const info = historyInfo.get(currSeqId);
+
+    historyInfo.delete(currSeqId);
+    historyInfo.delete(currSeqId);
```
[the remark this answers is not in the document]
griffinteller: :clown:

packages/editor/src/optimizer/worker.ts
```diff
@@ -186,28 +234,26 @@ const startOptimize = (stepSequenceId: StepSequenceID) => {
...
-  // on message, we will take a step, but these messages will be queued behind
-  // self.onmessage, so we can interrupt optimization with a message
-  const optStepMsgChannel = new MessageChannel();
 
   // take one optimization step
   const optStep = () => {
     try {
-      if (activeOptimizationId !== stepSequenceId) {
+      if (activeOptimizationId !== startingOptimizationId) {
         // we've been interrupted
```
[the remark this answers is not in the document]
griffinteller: If `startOptimize` is called again, it increments the id and interrupts the old loop. This could happen if we recompile, resample, or interact during optimization.

packages/editor/src/optimizer/worker.ts
```diff
@@ -289,13 +340,158 @@ const resample = (variation: string): ResampleResult => {
...
+  const parentValues = historyValues.get(data.parentHistoryLoc.sequenceId)![
+    data.parentHistoryLoc.frame
+  ];
+
+  let currentValues;
+  let stepSeqId: StepSequenceID;
+  if (parentSeqInfo.child === null) {
+    // this is a new interaction
+    const newInitValues = [...parentValues];
+    stepSeqId = makeNewStepSequence(
```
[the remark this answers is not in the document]
griffinteller: Correct. And when we call interact with a parentLoc that already has a child, we know we're continuing an interaction (a mouse move), so we just extend that child sequence.

packages/editor/src/optimizer/worker.ts
```diff
@@ -289,13 +340,158 @@ const resample = (variation: string): ResampleResult => {
...
+      parentSeqInfo.variation,
+      newInitValues,
+    );
+
+    currentValues = historyValues.get(stepSeqId)!.at(-1)!;
+  } else {
+    // we're continuing an interaction
+    stepSeqId = parentSeqInfo.child;
+    currentValues = [...historyValues.get(stepSeqId)!.at(-1)!];
+    historyValues.get(stepSeqId)!.push(currentValues);
```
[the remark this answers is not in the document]
griffinteller: Right now whatever the last set of varying values are in the list is used as the list for the penrose state, so its modified. But this is incredibly silly to worry about from outside `startOptimize`; I'll just change it to be duplicated there.

packages/editor/src/optimizer/worker.ts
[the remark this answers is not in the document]
wodeni: probably good to comment on this

packages/editor/src/optimizer/worker.ts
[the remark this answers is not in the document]
griffinteller: Edit: not quite. These are actually the values that are dragged; if we didn't duplicate them, we'd be erasing the last step. But again, in retrospect, I think the dragging functions should mutate this, they should just return a copy. I'll change that too.

packages/editor/src/utils/renderUtils.ts
```diff
@@ -25,8 +37,369 @@ export const stateToSVG = async (
...
+    bottomRightSVG.y - topLeftSVG.y,
+  );
+};
+
+export const interactAndUpdate = async (
+  interaction: Interaction,
+  diagramId: DiagramID,
+  historyLoc: HistoryLoc,
+  setDiagram: (setter: (diagram: Diagram) => Diagram) => void,
+  setWorker: (setter: (worker: any) => any) => void,
```
wodeni, line 98: probably good to define a `DiagramWorkerState` type. very minor tho

packages/editor/src/optimizer/worker.ts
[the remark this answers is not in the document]
wodeni: > Edit: not quite. These are actually the values that are dragged; if we didn't duplicate them, we'd be erasing the last step. But again, in retrospect, I think the dragging functions should mutate this, they should just return a copy. I'll change that too.

that's my guess. did you mean the dragging functions should return a copy instead of mutating them?

packages/core/src/renderer/Renderer.ts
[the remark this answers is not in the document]
wodeni: yep makes sense now! kinda wish we had more metadata to tell what shape type a string path has (e.g. `A.icon -> Circle`), that way we can handle mouseover properly, i.e. using `pointer-event: bounding-box` for `Text` and `Equation`. is there a better way to do this without more boilerplate?

packages/editor/src/optimizer/worker.ts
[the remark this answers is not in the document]
griffinteller: fixed.

packages/editor/src/utils/renderUtils.ts
griffinteller: good catch.

PR #1801, opened 2024-06-14
feat: LSP-style optimization worker(s)

griffinteller:
# Description

LSP inspired optimizer worker protocol. Decreasing statefulness and using a broker-worker(s) paradigm will help eliminate many tricky IDE bugs, as well as improve interactive performance.

# Checklist

# Current issues

- ~~Currently some issues if you resample while its still optimizing and vice versa; this has do with me not polling in a particularly smart way right now. Ultimately we'd like something like if optimizing, poll every 16.67 ms. Right now this a call back that puts itself on a timer until the diagram is no longer optimizing, but this leads to edge cases where the user has recompiled the diagram (which assigns a new diagramID) or resamples (which assigns a new stepSequenceId), but the old callback uses the old ids. Not sure the best way around that.~~ fixed

- ~~Indexing got funky somewhere for the `LayoutTimelineSlider`; current fix is to add 1 to a lot of the indexing but I'll track that down properly.~~ fixed


packages/editor/src/components/SegmentedSlider.tsx
samestep, line 1: These `+1`s are quite spooky, can you add code comments here pointing out that they're hacky / not principled so that people don't get confused trying to decipher them?

packages/editor/src/state/callbacks.ts
```diff
@@ -200,61 +164,99 @@ export const useResampleDiagram = () =>
...
+    //   set(diagramWorkerState, (state) => ({
+    //     ...state,
+    //     optimizing: true,
+    //   }));
+    //
+    //   const info = await optimizer.pollForUpdate();
+    //   if (info !== null) onUpdate(info);
+    // } catch (error: unknown) {
+    //   onError(error);
+    // }
```
samestep, line 259: Is there a reason to keep this code here commented-out instead of just deleting it?

samestep (commented):
@griffinteller I think I'll have an easier time reviewing this if I know a rough order in which to read through the different files. Can you post a quick comment here with a rough guide through the changes so I can review them more effectively?

packages/editor/src/components/SegmentedSlider.tsx
griffinteller: For now, yes; I'll also try to track down where the issue stems from so hopefully we can remove them.

packages/editor/src/state/callbacks.ts
griffinteller: Nope. I'll delete it : )

griffinteller:
@samestep In terms of order, I would go:
1. `optimizer/common.ts`
2. `optimizer/optimizer.ts`
3. `optimizer/broker.ts`
4. `optimizer/worker.ts`
5. `state/callbacks.ts`
6. components, in any order
7. Everything else should be very minor

wodeni:
Haven't gone through all the diffs yet. Tested the branch locally and repro'ed the resample-while-optimizing problem with `?examples=box-arrow-diagram%2Fcomputer-architecture`. I'm getting `InvalidHistoryLocError` on every 2~3 resamples pretty consistently.

Is the `InvalidHistoryLocError` caused by stale layout requests? Looking at `runComputeLayout`, the `setTimeout` did jump out to me. I recommended using `requestAnimationFrame` for better performance, but this will likely not get rid of stale layout requests after resample.

packages/editor/src/components/SegmentedSlider.tsx
griffinteller: Found the issue: I was initializing the length of the each step sequence info to 0 but immediately placing in the unoptimized parameters. fixed, and spooky +1s are no more.

griffinteller:
> Haven't gone through all the diffs yet. Tested the branch locally and repro'ed the resample-while-optimizing problem with `?examples=box-arrow-diagram%2Fcomputer-architecture`. I'm getting `InvalidHistoryLocError` on every 2~3 resamples pretty consistently.
> 
> Is the `InvalidHistoryLocError` caused by stale layout requests? Looking at `runComputeLayout`, the `setTimeout` did jump out to me. I recommended using `requestAnimationFrame` for better performance, but this will likely not get rid of stale layout requests after resample.

Should be fixed; it now keeps track of the most recent three diagrams and step sequences per diagram (configurable, obviously) before deleting. Also, now uses `Ref`s to signal stop conditions to the render loop, instead of `cancelAnimationFrame` which can interleave poorly if the animation frame callback is `async`.

packages/editor/src/optimizer/common.ts
```diff
@@ -0,0 +1,614 @@
...
+}[];
+
+/**
+ * Current state of a sequence of optimizer states:
+ *  - `"Pending"`: optimizer has not finished (ongoing, interrupted). Could
+ *    or will continue.
+ *  - `"Done"`: optimizer finished normally
+ *  - `OptimizerError`: optimizer errored and cannot continue.
+ */
+export type StepSequenceState = "Pending" | "Done" | OptimizationError;
```
samestep, line 64: Is there a reason you don't do this instead?

```typescript
export type StepSequenceState =
  | { tag: "Pending" }
  | { tag: "Done" }
  | OptimizationError;
```

Seems like this would be more convenient/consistent, because then you can just do `switch (stepSequenceState.tag)`.

packages/editor/src/optimizer/common.ts
samestep, line 1: In general, is there a reason you tend to use `type` instead of `interface` in this file? I know there are slight differences between them, dunno which should generally be preferable.

packages/editor/src/optimizer/common.ts
```diff
@@ -0,0 +1,614 @@
...
+export enum MessageTags {
+  Compile = "Compile",
+  Poll = "Poll",
+  ComputeLayout = "ComputeLayout",
+  DiscardDiagram = "DiscardDiagram",
+  DiscardStepSequence = "DiscardStepSequence",
+  Init = "Init",
+  LabelMeasurements = "LabelMeasurements",
+  Resample = "Resample",
+}
```
samestep, line 167: Using an `enum` seems like a good idea! Is there a reason you don't do it for other `tag`s too?

packages/editor/src/optimizer/common.ts
```diff
@@ -0,0 +1,614 @@
...
+export type TaggedErr<E, T> = {
+  tag: T;
+  variant: "Err";
+  error: E;
+};
+
+/**
+ * A Result type, but with a tag. Supertype of `MessageResult`.
+ */
+export type TaggedResult<V, E, T> = TaggedOk<V, T> | TaggedErr<E, T>;
```
samestep, line 230: Do you still need these custom result types if you're using the `true-myth` library?

packages/editor/src/optimizer/common.ts
```diff
@@ -0,0 +1,614 @@
...
+  private nextId = 0;
+
+  next = () => {
+    return this.nextId++;
+  };
+}
+
+// We reexport as different generators for each ID type, in case we decide
+// not to use numbers in the future
+export {
```
samestep, line 291: No need to do this now, but a tool that might be useful in the future, in case we want them to stay numbers but also be different types statically: https://egghead.io/blog/using-branded-types-in-typescript

packages/editor/src/optimizer/common.ts
```diff
@@ -0,0 +1,614 @@
...
+
+  get = (key: K): V | undefined => {
+    return this._map.get(key);
+  };
+
+  set = (key: K, value: V): void => {
+    // if we're about to grow past our maximum length, find the oldest key
+    // and remove it from _keys and _map.
+    if (this._map.size === this.maxLen) {
+      const toDeleteKey = this._keys.shift();
```
samestep, line 588: Great thinking to build this `SizeBoundedMap` type! One comment: this usage of `shift` turns `set` from $O(1)$ to $O(n)$; is there a way you can structure it so it remains $O(1)$?

packages/editor/src/optimizer/optimizer.ts
```diff
@@ -0,0 +1,254 @@
...
+      this.messageIdGenerator,
+    );
+  };
+
+  /**
+   * Create an optimizer.
+   */
+  static create = async (): Promise<Optimizer> => {
+    log.info("Spinning new broker");
+    return new Optimizer(await spinAndWaitForInit("./broker.ts"));
```
samestep, line 102: I'm slightly worried about whether this relative URL will always get correctly transformed, since it's not being used in a literal `new URL` expression. Could you change this so it instead uses an [`import` with Vite's `?worker` suffix](https://v3.vitejs.dev/guide/features.html#import-with-query-suffixes)?

packages/editor/src/optimizer/common.ts
```diff
@@ -0,0 +1,614 @@
...
+    });
+    worker.postMessage(request);
+  });
+};
+
+/**
+ * Launch worker, and wait for a notification `InitData` before resolving.
+ * @param url Worker to spin
+ */
+export const spinAndWaitForInit = async (url: string): Promise<Worker> => {
```
samestep, line 341: When I hear "spin" I think that the thread is blocking, running an infinite `for (;;)` loop or something. Just to clarify, that's not happening here, right? In that case, can you change the name so it doesn't contain the word "spin"?

packages/editor/src/optimizer/optimizer.ts
```diff
@@ -0,0 +1,254 @@
...
+      svgCache = svgCacheResult.value.svgCache;
+      this.svgCaches.set(diagramId, svgCache);
+      this.labelMeasurements.set(diagramId, svgCacheResult.value.optLabelCache);
+    } else {
+      svgCache = this.svgCaches.get(diagramId)!;
+      this.labelMeasurements.set(diagramId, result.value.labelMeasurements);
+    }
+
+    return Result.ok(layoutStateToRenderState(result.value, svgCache));
+  };
```
samestep, line 219: Funny, something about this function breaks syntax highlighting in VS Code!

packages/editor/src/optimizer/common.ts
samestep: Doesn't really matter right now since I see that you have `numWorkersToKeep = 3` and `numStepSequencesToKeep = 3`, but I feel like we've gotten bitten by these sorts of time complexity issues in the past.

packages/editor/src/optimizer/common.ts
samestep: Oh I think I understand, does "spin" mean "spin up"? I think it's fine to say "spin up" then, just not "spin" by itself.

packages/editor/src/optimizer/worker.ts
```diff
@@ -0,0 +1,366 @@
...
+  return parentLoc === null
+    ? [...penroseState!.varyingValues]
+    : [...historyValues.get(parentLoc.sequenceId)![parentLoc.step]];
+};
+
+/**
+ * Compile, but don't start optimizing until we receive `LabelMeasurementsData`
+ * in a notification. After calling, `penroseState` will be non-null.
+ * @param data
+ */
```
samestep, line 78: Can you or @wodeni remind me why it can't compute the label measurements in the worker itself? For some reason I thought this was possible.

packages/editor/src/optimizer/worker.ts
```diff
@@ -0,0 +1,366 @@
...
+  const info: StepSequenceInfo = {
+    layoutStats: [
+      {
+        name:
+          penroseState!.optStages.length === 1
+            ? "default"
+            : penroseState!.optStages[0],
+        // steps, as it is used in the rest of the codebase, indicates the number
+        // of varying values in a list--not the number of optimizer steps. So
+        // we initialize this to one, since one varying value is added initially
```
samestep, line 151: Would it make sense to use a different word here, then?

packages/editor/src/optimizer/worker.ts
```diff
@@ -0,0 +1,366 @@
...
+};
+
+self.onmessage = async ({
+  data,
+}: MessageEvent<MessageRequest | Notification>) => {
+  switch (data.tag) {
+    case "MessageRequest":
+      {
+        const requestData = data.data;
+        log.info(`Worker recieved request ${requestData.tag}`);
```
samestep, line 308:
```suggestion
        log.info(`Worker received request ${requestData.tag}`);
```

samestep (approved):
Awesome work! I left several minor comments and suggestions. One thing I'm not sure if you already told me: given that you limit the number of different diagram workers to 3, how does the diagram panel work properly when there are many more than 3 in the grid?

packages/editor/src/optimizer/common.ts
griffinteller: This is a better idea, I'll change that.

packages/editor/src/optimizer/common.ts
griffinteller: As far as I know (and this is likely far from a complete understanding), the only differences between expressing `interfaces` versus `type` aliases are how extension/intersection is handled. `interface`s will error if you try to `extend` a type alias whose members are not statically known, while `type` intersection will just intersect their union. Honestly though, I just prefer using type everywhere I can for consistency (since you need them for unions), plus I find intersection to be more readable than `extends`.

packages/editor/src/optimizer/common.ts
griffinteller: I wanted to enure that `MessageRequestData`isubtypes have the same tag as their corresponding `MessageResult` subtype, and so the enum really was necessary to prevent a typo in the definition of the types. Mistyping a tag during instantiation will just get caught by the typechecker, so I thought it was less necessary. The only other tags in this file are for `MessageResponse`, `MessageRequest, and `Notification`, but I can add those to a `MessageTypeTag` enum, if you'd like?

samestep: Up to you! Sounds like it's not actually necessary as you mentioned; I was just curious.

packages/editor/src/optimizer/common.ts
samestep: Gotcha :/ yeah that makes sense; up to you if you want to get rid of `true-myth` entirely. I don't think getting rid of the dependency matters much per se.

packages/editor/src/optimizer/common.ts
griffinteller: Oh, sweet! That's perfect. Thanks!

packages/editor/src/optimizer/common.ts
griffinteller: I think launch is maybe just a better word.

packages/editor/src/optimizer/worker.ts
griffinteller: mathjax needs a reference to the window : /

packages/editor/src/optimizer/worker.ts
griffinteller: Sure, I can change it to "frame"; pretty sure this only affects `LayoutTimelineSlider` so shouldn't be a big change.

packages/editor/src/optimizer/worker.ts
samestep: Sad but makes sense!

wodeni: Yep, copying over an old comment from our thread with @YonahGoldberg summarizing the "polyfill" version of the design:

> when main thread creates worker, send offscreen canvas which worker stores (or if the worker can create it directly, then just do that)
use OffscreenCanvas to allow text measurement to run directly on webworker by making measureTextElement an argument of measureText instead of a global
use liteAdaptor to allow mathjax to run directly on the webworker
also run the renderer on the webworker, generate svg string
send that string across the wire when main thread asks for an update
main thread converts that svg string to dom

There were some weirdness around `OffscreenCanvas` and `LiteAdapter` at the time, but I don't think they can't run on workers in theory.

packages/editor/src/optimizer/optimizer.ts
wodeni: same here!

wodeni: I'm guessing it's the ambiguity around `>` that caused the faulty highlighting?

packages/editor/src/optimizer/common.ts
wodeni: was also expecting a `while(true)` before I jumped to definition. `waitForInit` might just be fine?

griffinteller:
> Awesome work! I left several minor comments and suggestions. One thing I'm not sure if you already told me: given that you limit the number of different diagram workers to 3, how does the diagram panel work properly when there are many more than 3 in the grid?

The diagram variations does not plug into this yet--this PR only reaches feature parity with main. When we do connect the diagram variation panel with this, we'll need to change our approach here some.

packages/editor/src/optimizer/common.ts
griffinteller: Good point--javascript should really have a builtin linked list type? I'll implement one real quick.

packages/editor/src/optimizer/optimizer.ts
griffinteller: It is being transformed, within `spinAndWaitForInit`.

PR #1810, opened 2024-07-01
build: remove monaco and upgrade to storybook 8

wodeni:
# Description

Now that #1798 is removing monaco, `@penrose/components` should be monaco-free as well. This PR replaces monaco with the codemirror in `Listing` component.

As a side, this PR also bumps storybook from 6 to 8 for better build configs.


KyleleeSea (approved):
W

Issue #1814, opened 2024-07-02
bug: Dangling domain identifier 


KyleleeSea:
**Describe the bug**
With the new Lezer parser, in domain there is ambiguity when a subtype statement follows a constructor or function statement that does not have a named return type. 

<img width="341" alt="Screenshot 2024-07-02 at 2 17 37 PM" src="https://github.com/penrose/penrose/assets/85892844/b9519712-25a5-4ed0-bfae-9aab247606d7">

This is because we ignore newlines and after the return type identifier, there is an optional return name identifier for functions and constructors. When a subtype statement follows, the next identifier could be either the start of a subtype or the end of a function/constructor output type.

It is ambiguous to Lezer which is the correct case and it ends up accepting it as the return name for the function/constructor, causing a parse error. 

**To Reproduce**
Go to domain
Type:
`type A
type B        
function f(A arg) -> B
A <: B`
Observe error on last line 

**Notes**
We've tried the following to fix this:
1. Adding ambiguity markers so Lezer uses GLR. This did not have any effect although we tried many different combinations of where ambiguity markers went. 
2. Looking into insertsemi to see if the insertedsemi on newline would hint to Lezer that the new identifier belonged to a new statement after a newline. For some reason, this insertedsemi would be placed on newline, but removed and moved to following the ambiguous identifier once the identifier has been typed. This seems to do with acceptToken in Codemirror library.
3. Making domain newline sensitive. This was decided to be not worth it because it leads to requiring newlines, having to make special rules about newlines at the beginning of programs, and potential extraneous edge cases.

PR #1815, opened 2024-07-02
test: run components tests in CI

wodeni:
# Description

Following up on #1798, this PR changes the "test" job in our CI to run the test suite in `@penrose/components`, which includes some editor-specific tests and parser tests (which are to be ported to `core`). 

The `components` test suite currently has two failing tests:

* "tree integrity": contains a case of #1814, which is refactored to not test this case
* "dangling output type conflict with subtype decls": explicitly tests for #1814, which is to be un-skipped after the issue is resolved.


KyleleeSea (approved):
W

PR #1820, opened 2024-07-03
fix: prohibiting inline subtyping on literal types

### LINKED ISSUE #1818: bug: inline subtyping on built-in types not checked
**Describe the bug**

For built-in literal types (`Number` and `String`), we do not allow these types to act as subtypes or supertypes.

If I write
```
type X
X <: Number
```
the compiler correctly reports an error.

But if I inline the subtyping, as in
```
type X <: Number
```
then the error is unreported.

liangyiliang:
# Description

Resolves #1818.

This PR moves the check on prohibiting literal types to whenever a subtyping relation is registered. That way, no matter where the subtyping relation appears (in `TypeDecl` or in `SubtypeDecl`), the checker will prohibit subtyping on literal types.

A test has been added.

# Checklist

### DIFF
--- a/packages/core/src/compiler/Domain.test.ts
+++ b/packages/core/src/compiler/Domain.test.ts
@@ -268,5 +268,9 @@ type Set
 function F(Set s) -> String
     `;
     expectErrorOf(prog4, "OutputLiteralTypeError");
+    const prog5 = `
+type Set <: Number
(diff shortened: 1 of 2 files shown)


wodeni (approved):
good catch! also thanks for adding the test case 💯

PR #1822, opened 2024-07-03
feat: added cntrl/cmd+enter and :w to compile diagram

KyleleeSea:
# Description
Feature existed prior to Codemirror migration. This PR adds it back in. 
1) In Vim mode, :w will compile diagram. 

https://github.com/penrose/penrose/assets/85892844/79b776c8-a747-4acd-b399-ae2704d62773

2) Cntrl/Cmd+Enter will compile diagram

**Notes**
Listing uses EditorPane and does not pass in an onWrite property. In this case onWrite will be undefined and thus no keymaps will be added.


wodeni (approved):
:W 🥇

PR #1823, opened 2024-07-03
feat: Layout stage autocomplete + corresponding tests

KyleleeSea:
# Description
Adds stage name autocomplete based on the names users defines after `layout = [...]`

Adds four tests to test suite both testing layout stage names obtained correctly and autocomplete fires for stage names. 

Screenshot 2024-07-03 at 4 40 16 PM


wodeni (approved):
Looks good to me! Thanks for the PR :D.

PR #1833, opened 2024-07-17
feat: added recover button to Settings

KyleleeSea:
# Description

Post cloud storage merge, users no longer see their old locally saved diagrams. This adds a button to settings which automatically traverses local storage and saves all locally saved diagrams to cloud storage and syncs the system.

https://github.com/user-attachments/assets/44dbc551-f3b7-47ce-817e-fff76141f9cd

Function loops over all locally saved file ids and saves diagrams whose ids are not in the cloud storage saved list. This required adding the package ["localForage-getItems"](https://github.com/localForage/localForage-getItems/tree/master) which traverses local storage with a driver. localForage did not have this capability built in.

This strategy has two implications:
1 - If a user deletes some of these diagrams from cloud storage, clicking this button a second time will restore the ones they deleted 
2 - If somehow the user created another diagram that miraculously has the same id as a legacy locally saved diagram, that legacy locally saved diagram won't be restored. Diagram id strings are very long, the probability of this is negligible. 

Additionally, this required refactoring useSaveNewWorkspace. I created a helper function saveNewLogic which is now used by both useSaveNewWorkspace and useRecoverAll.

Transparency:
1 - In Settings users is displayed a count of how many unrestored legacy diagrams they have
2 - User is given a confirmation box listing the names of all diagrams to restore upon clicking "Recover"

Regarding 1: This requires us calling a function to traverse local storage to get ids every time the savedFiles recoil object is updated (otherwise once the diagrams get restored, the number remains the same in this view, which is misleading). This doesn't seem to cause any noticeable performance decrease, but if performance suffers unexpectedly, come back to this useEffect

# Open questions
The help string I display in settings is: "We've migrated to cloud storage! Recover all legacy locally stored diagrams:" is this good?

### DIFF
--- a/packages/editor/package.json
+++ b/packages/editor/package.json
@@ -49,6 +49,7 @@
     "color-name-list": "^8.38.0",
     "flexlayout-react": "^0.7.3",
     "localforage": "^1.10.0",
+    "localforage-getitems": "^1.4.2",
     "lodash": "^4.17.21",
     "query-string": "^7.1.1",
     "react": "^18.2.0",
--- a/packages/editor/src/components/SavedBrowser.tsx
+++ b/packages/editor/src/components/SavedBrowser.tsx
@@ -47,14 +47,21 @@ export default function SavedFilesBrowser() {
             {(currentWorkspace.metadata.location.kind !== "stored" ||
               !currentWorkspace.metadata.location.saved) && (
               <BlueButton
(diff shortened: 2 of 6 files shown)


KyleleeSea:
https://github.com/user-attachments/assets/69d516ab-1c17-491d-bfaf-b1d1c439479c

PR #1837, opened 2024-07-24
feat: organize top bar and settings in `editor`

KyleleeSea:
# Description
Goals...
- Improving usability and intuitiveness of the IDE 
- Making the mobile view more aesthetic (this is not a complete mobile UX overhaul but focuses on fixing blatantly broken styles)

Old mobile top bar for reference:

Screenshot 2024-07-25 at 12 33 40 PM

Notice:
- Top bar overflow makes the content overflow and be horizontally scrollable
- Top bar buttons become too tall and make the top bar ugly

Changes...
**1: Removes debug mode & state sidebar and moves into Settings**
Previously:

Screenshot 2024-07-25 at 12 27 04 PM

Now:

Screenshot 2024-07-25 at 12 27 15 PM

Rationale:
- debug mode is misleading as a title, the options open are just Settings, they don't provide much debugging assistance
- Settings previously felt quite barren
- debug mode and state opening in the right hand side made the UI feel very cluttered, especially on mobile

**2: Collapses Save Penrose SVG into export**
Rationale: Save Penrose SVG was misleading to users and made new users initially miss the cloud storage save mechanics and made them instead think that saving had to be done via manual download and upload

**3: Moves export from top bar into side bar cards**

https://github.com/user-attachments/assets/54e01e13-ffa7-4ef5-916c-b5394a4ddad5

Rationale:
- Primarily to clear top bar space for mobile, although the additional top bar space on Desktop gives the UI more room to breathe and looks better
- Previously the "upload" side panel felt too barren

**4 Mobile "new workspace" retitled to "new"**

**5: Mobile "save" "share" "new" right aligned**

**6: Mobile "compile" and "resample" moved into diagram panel**

**7: Mobile title concatenates when too long**
Previously would overflow and look bad

Screenshot 2024-07-25 at 12 40 00 PM

In sum, mobile top bar now looks like:

Screenshot 2024-07-25 at 12 48 02 PM

**8: Syntax errors open to docs on click**

https://github.com/user-attachments/assets/971433f9-41cc-41d5-953a-1a6b69e3218d

All substance syntax errors open to substance overview docs, all style syntax errors open to style overview docs, all domain errors open to domain overview docs

**9: Open to docs button**

https://github.com/user-attachments/assets/3be194cb-a759-417c-a72f-daa382154240

Button disappears on mobile to save top bar space 

# Open Questions
- Mobile title concatenation looks premature when there's no save/share button up (i.e. when opening an example for the first time), I could fix this by adding an if condition checking if the diagram is "local" or "example", but I didn't do this because it felt like bad code

Screenshot 2024-07-25 at 12 48 11 PM

- (has nothing to do with this change) sometimes syntax error displays twice "Syntax Error Syntax Error". This is because there are cases where Lezer will create multiple error nodes (example: writing just "a" in substance will error the type node and the program node)

- Old IDE issue: Once examples is moved to the center tabset when user enters mobile size, it does not move back afterwards. I could not find an easy fix for this in flexlayout-react since the left sidebar has to be type "IJsonBorderNode" which for whatever reason does not accept id as a property. The API Action.moveNode targets the move location based on id. 

- Onboarding quick tour: Currently this is not reliable with the flexlayout framework, since if we want to highlight any specific elements we need to ensure that tab is open but the only action flexlayout API provides is "selectTab" which toggles whether the tab is open, meaning if the user started with that tab open, calling this would close the tab we want to highlight elements in rather than open it


KyleleeSea:
**Updates:**
- Removed `StyleWrapper` replaced by explicitly using styled components defined in `SettingElements.ts`
- Adjusted button alignment in Settings 
- Fixed checkbox alignment next to "vim mode"
- Changed size of dropdown text in Select box to be less loud
- Decreased padding of Select box
- Made alignment more consistent between Variation and General settings
- Improved alignment between grid size slider and the number marker 
- Changed variation seed input box to use same border style as Select
- Added margins for variation seed input box 
- Changed variation seed input box to change color like Select on focus
- Put "Misc" settings (right now just the restore diagrams thing) in a Section card 

**Settings Before**
Screenshot 2024-07-30 at 11 31 26 AM

Screenshot 2024-07-30 at 11 47 15 AM

**Settings Now**
https://github.com/user-attachments/assets/bd052178-ff49-4374-a189-bca0b71a3a96

**Open Questions:**
1) For the Select box, there's an issue when the option text is longer than the box. I could not figure out how to solve this
Screenshot 2024-07-30 at 1 00 46 PM
https://react-select.com/styles#inner-components

2) Did not add back in docs button
Agreed the ? button was too loud
A solid blue button in the top bar also seems too loud
Creating a button style just for docs seems to draw too much attention to it
Ideally, I'd like to put the docs button in the left sidebar and on click it opens docs in a new tab
However, it seems react-flexlayout does not support this behavior 
Since this PR adds syntax errors being clickable and redirecting to docs, the specific docs button does not seem as necessary

wodeni (approved):
pushed 2 commits to fix 1, 2, and 3 above. good to go!

PR #1840, opened 2024-07-26
fix: Better Hover + info box type names

KyleleeSea:
# Description

Uses valueTypeDesc object to map function parameter types to more intuitive names
Creates a map for shape prop type names with names similar to those used in valueTypeDesc


wodeni (approved):
based PR. the type names look a lot better now!

PR #1846, opened 2024-08-26
feat: Sublime syntax and Set potatoes example

Fulguritude:
# Description

The code for this pull request contains:

- the 3 sublime-syntax files for Penrose in `util/subl` (to install them, simply go to "Preferences -> Browse packages" in ST3, then copy-paste the 3 files into the `User` directory found there). 

- an extra example directory, for "classical" Bourbaki-style "set potatoes", and functions/relations between them.

Sublime: LexouDuck basically redid my shoddy work from scratch following the Penrose syntax specifications.

Set potatoes:

- I tried to keep things simple (only 4 basic type: Set, Element, Relation, Arrow).

- The main idea was to encourage the diagram to be neatly legible from left-to-right and use color, since there was some inevitable amount of overlap.

- I provided some predicates which can remove overlap in the case of more complex diagrams (like with "triangularly related set trios", see image below).

- Finally, I stored the relevant constant parameters in a `Global` namespace, and neatly commented my code throughout.

There are substance files for 5 examples, in increasing complexity:

- a relation which is not a function
    
- a diagram showing that an injection can have multiple post/left-inverses
    
- a diagram showing that a surjection can have multiple pre/right-inverses (sections)
    
- a diagram showing that a non-injection does not verify the monomorphism property
    
- a diagram showing that a non-surjection does not verify the epimorphism property 

An image of the last example for the heck of it:

Screenshot from 2024-08-26 16-09-17

# Checklist

# Open questions

   - syntax coloring does have some issues when a keyword is used somewhere it has a secondary meaning (e.g.: `string` or `above`)

   - not sure how to improve diagram color other than by just letting the colors be generated at random; if colors can be specified in the substance file like Labels are, then I couldn't figure out how to do it.
   
   - might want to allow a special "ellipsis" element for infinite sets in the future for the domain/style pair.


Fulguritude:
CI was failing for format (extra line returns in sublime-syntax files), so I ran yarn format and pushed that

liangyiliang:
Hi @Fulguritude , this look great, thanks for doing this!

Do you mind adding your new set-potatoes diagrams into the registry? [Here's how](https://github.com/penrose/penrose/blob/main/CONTRIBUTING.md#registry).

Fulguritude:
Update: added 5 trio files, changed the registry.json (and only put one of the 5 diagrams, the one screenshot above, as a new gallery member since it'd be redundant otherwise). Will this do the trick ?

liangyiliang:
Yes it does! Thanks for this - once the automatic checks are done I'll do a final check before merging it

liangyiliang (approved):

PR #1864, opened 2024-10-12
feat: standalone Bloom distribution

griffinteller:
# Description

Bundles Bloom into a `min.js` file, which can be accessed at `https://penrose.cs.cmu.edu/bloom.min.js` after deployment. 
Adds method "Diagram.prototype.getInteractiveElement", which produces an `HTMLDivElement` displaying an interactive Bloom diagram. This removes the soft requirement of using React with Bloom, and allows Bloom to be used easily from static HTML. 

Example usage:

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title>Title</title>
</head>
<body>
  <div
    id="diagram-parent"
    style="
      width: 50em;
      height: 50em;
      border: 3px solid black;
      margin: auto;
      margin-top: 5em;"
  ></div>
  <script type="module">
    import { DiagramBuilder, canvas } from "https://penrose.cs.cmu.edu/bloom.min.js";

    const db = new DiagramBuilder(canvas(400, 400), "");

    db.circle({
      r: 10,
      drag: true,
    });

    const diagram = await db.build();
    const interactiveElement = diagram.getInteractiveElement();

    document.getElementById("diagram-parent").appendChild(interactiveElement);
  </script>
</body>
</html>
```

Additionally removes the "site integration" bloom tutorial, since we now need a dedicated react integration section.

When `Diagram.prototype.getInteractiveElement` is called, an optimization and render loop is started behind the 
scenes, which continues until the diagram converges. Both loops are restarted on interaction.

The React implementation of the optimization/render loop has been replaced with a thin wrapper around `getInteractiveElement`.

# Open questions

- Should we move React integration into a separate npm package (e.g @penrose/bloom-react)? It would be a very small package, but would save many Bloom users from the React dependency.
- At some point, we should think about versioning these `min.js` files, and ensuring that we keep hosting older ones.
- Relatedly, we might want to host these files on `cdnjs`
- 2.6 MB is pretty chunky


Fil:
It would be great to host this file on npm or jsr, too. thanks!

PR #1953, opened 2026-07-13
fix: swapped arguments for contains() call in predicates tutorial

### LINKED ISSUE #1954: First Two Arguments Are Flipped in `contains()` Call in Predicates Tutorial Page
**Describe the bug**
The [predicates](https://penrose.cs.cmu.edu/docs/tutorial/predicates) tutorial page calls the `contains()` constraint to visualize a subset of a Set as a smaller circle drawn within a bigger circle. But the ordering of the arguments in the call to `contains()` is flipped, which draws the subset bigger than its superset.

`contains(s1, s2, padding)` requires that a shape s1 contains another shape s2, therefore within the condition Subset(A, B) where Set A is a subset of Set B, the shape of instance B should contain the shape of instance A.

**To Reproduce**
Follow the steps of the [predicates](https://penrose.cs.cmu.edu/docs/tutorial/predicates) tutorial page (excluding the exercises).

Then, draw the Sets' labels to see which circle corresponds to which Set. Replace the style file with the following text:
```
canvas {
  width = 800
  height = 700
}

forall Set A; Set B
where Subset(A, B) {
    ensure contains(A.icon, B.icon, 5.0)
    ensure disjoint(A.text, B.icon)
    A.icon above B.icon
}

forall Set x {
    x.icon = Circle {
        strokeWidth : 0.0
    }
    x.text = Equation {
        string : x.label -- Make sure to write `AutoLabel All` at the end of the substance file for this line to work!
        fontSize : "35px"
    }
    ensure x.icon.r > 25
    ensure contains(x.icon, x.text)
}
``` 

**Expected behavior**
The substance file dictates that Set B should be a subset of Set A, and Set C is a subset of Set B. Therefore the biggest circle should be Set A, with a smaller circle inside for Set B, and an even smaller circle by Set C inside of Set B.

**Screenshots**
Currently, Set C is the biggest circle and Set A is the smallest, but Set A should be the biggest:
Image

Galm007:
`contains(s1, s2, padding)` requires that a shape s1 contains another shape s2, therefore within the condition `Subset(A, B)` where **Set A is a subset of Set B, the shape of instance B should contain the shape of instance A**.

# Description

Resolves #1954.

Fixes an issue within the [predicates](https://penrose.cs.cmu.edu/docs/tutorial/predicates) tutorial page where a call to the contains() swaps the first two arguments, which visualizes a Subset as a Superset.

Substance:
```
/* This is the starter code for the substance program for tutorial 2,
 * which covers constraints in Penrose. Follow along with the write-up.
 * Good luck! :)
 */

-- Start your code below
Set a, b, c
Subset(b, a)
Subset(c, b)

AutoLabel All
-- End your code above
```

## Before fix:
image

## After fix:
image

### DIFF
--- a/packages/docs-site/docs/tutorial/predicates.md
+++ b/packages/docs-site/docs/tutorial/predicates.md
@@ -74,7 +74,7 @@ Now our selector is not just `forall Set A` since we only want to apply these st
 ```style
 forall Set A; Set B
 where Subset(A, B) {
-    ensure contains(A.icon, B.icon, 5.0)
+    ensure contains(B.icon, A.icon, 5.0)
     A.icon above B.icon
 }
 ```
@@ -101,7 +101,7 @@ So putting it all together, we have:
 ```style
 forall Set A; Set B
 where Subset(A, B) {
-    ensure contains(A.icon, B.icon, 5.0)
+    ensure contains(B.icon, A.icon, 5.0)
     A.icon above B.icon
 }
 


liangyiliang:
Thanks for catching it!

PR #1964, opened 2026-07-20
feat(bloom): add subtype selector matching

### LINKED ISSUE #1867: bloom: subtyping
Penrose has sub-typing:
```
type A
type B <: A
type C <: B
```
and selector-matching on `type A` (`forall A a`) gives us all matchings on `B` and `C`. This is currently not available in Bloom.

jiwonMe:
## Summary
- Extend `DiagramBuilder.type` to accept one or more previously declared supertypes
- Match `forall` and `forallWhere` selectors through reflexive and transitive subtype relationships, including multiple inheritance
- Add Bloom unit tests and tutorial documentation for subtype behavior

Fixes #1867

## Test plan
- [x] `yarn nx run bloom:test` (7 tests)
- [x] `yarn eslint packages/bloom/src/core/builder.ts packages/bloom/src/core/builder.test.ts --report-unused-disable-directives --max-warnings 0`
- [x] `yarn nx run bloom:build`

## Notes
- The package-wide `yarn nx run bloom:lint` still reports 7 pre-existing errors in `diagram.ts`, `types.ts`, and `react/hooks.ts`; the files changed by this PR pass ESLint.

### DIFF
--- a/packages/bloom/src/core/builder.ts
+++ b/packages/bloom/src/core/builder.ts
@@ -220,6 +220,8 @@ export class SharedInput {
  */
 export class DiagramBuilder {
   private substanceTypeMap: Map<Substance, Type> = new Map();
+  private declaredTypes: Set<Type> = new Set();
+  private typeSupertypes: Map<Type, Set<Type>> = new Map();
   private nextSubstanceId = 0;
   private substanceIdMap: Map<Substance, number> = new Map();
   private canvas: Canvas;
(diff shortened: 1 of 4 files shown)


packages/bloom/src/core/builder.ts
```diff
@@ -397,6 +417,25 @@ export class DiagramBuilder {
...
+        return true;
+      }
+      if (visited.has(current)) {
+        continue;
+      }
+      visited.add(current);
+      pending.push(...(this.typeSupertypes.get(current) ?? []));
+    }
+
+    return false;
```
liangyiliang: Small comment -- the library already has a Graph type https://github.com/penrose/penrose/blob/main/packages/core/src/utils/Graph.ts

why are we not just using Graph.descendants()? This is something that the Penrose Domain language compiler already does -- perhaps we can just reuse that part?

liangyiliang (commented):
Hi! Thanks for doing this! Left one small comment

packages/bloom/src/core/builder.ts
jiwonMe: Good call — updated to reuse `Graph` from `@penrose/core`.

The type hierarchy is now stored as a `Graph<Type>` with subtype → supertype edges (same orientation as Domain’s type graph), and `isSubtype` uses `typeGraph.descendants(actual).has(expected)`, matching `superTypesOf` / `isDeclaredSubtype`.

liangyiliang (approved):
Thanks for doing this! I will merge it.

PR #1961, opened 2026-07-19
fix(editor): eliminated memory leaks caused by detached SVG elements, mouse event listeners, and render state pileup during resample/compile

Galm007:
# Description

Resolves #1955.
           
This PR contains fixes within `@penrose/editor` for cleaning up unused SVG elements, mouse Event Listeners, and render state pileup in the diagramState Recoil atom during resample/compile, along with fixes to other minor memory leaks.

I've included a benchmark script at `mem-bench/leak-bench.mjs` for profiling leaked memory per resample/compile of select example diagrams. I've also made sure that the visual behavior of the editor in compilation and resampling of diagrams remains unchanged across all of the example diagrams.

## Memory Leak Benchmarks Before And After This PR

Mouse event listeners leaked per round: Completely eliminated.
| Diagram | Δlisteners/resample (before Fix) | Δlisteners/resample (after Fix) | Δlisteners/compile (before Fix) | Δlisteners/compile (after Fix) |
|-------------|--------------------------------------------|-----------------------------------------|-----------------------------------------|-------------------------------------------|
| caffeine | 46418.6 | 0.0 | 46002.5 | 0.0 |
| quaternion-cayley-graph | 6724.0 | 0.0 | 6930.9 | 0.0 |
| tree-euler | 6365.7 | 0.0 | 6601.1 | 0.0 |

Note: Examples `array-models/insertionSort` and `dinoshade/dinoshade` are excluded because they do not contain any interactable elements.

DOM nodes leaked per round: Completely eliminated in resamples, mostly eliminated in compiles.
| Diagram | Δnodes/resample (before Fix) | Δnodes/resample (after Fix) | Δnodes/compile (before Fix) | Δnodes/compile (after Fix) |
|-------------|--------------------------------------------|-----------------------------------------|-----------------------------------------|-------------------------------------------|
| caffeine | 9244.1 | 0.0 | 9248.9 | 12.2 |
| insertionSort | 1346.8 | 0.0 | 1786.5 | 25.1 |
| dinoshade | 721.0 | 0.0 | 406.9 | 2.1 |
| quaternion-cayley-graph | 244.0 | 0.0 | 367.9 | 12.0 |
| tree-euler | 104.4 | 0.0 | 191.4 | 10.0 |

Heap kB leaked per round: Completely eliminated in resamples, mostly eliminated in compiles.
[...]

Issue #1967, opened 2026-07-20
Memory leak: `mathjaxInit()` registers a new MathJax handler + document on every compile


Galm007:
**Describe the bug**

Every time a diagram is compiled, *CollectLabels.ts* in `@penrose/core` builds a brand-new MathJax pipeline (adaptor, HTML handler, and a full HTML document) and registers the handler into MathJax's process-global handler list which is never cleaned up and stays in the heap for the tab's lifetime. As a result, each compile permanently leaks a detached `HTMLDocument`, a `DOMParser`, a TeX/SVG pipeline, and several others. In the editor, recompiling in a loop linearly increases the tab's memory footprint, which can be inspected in JS heap snapshots.

Taking two snapshots of the JS heap where I repeatedly compiled the diagram 50 times in between the snapshots, comparing their memory footprints show a clear +50 delta on many HTML items:

<img width="2560" height="1433" alt="Image" src="https://github.com/user-attachments/assets/18edd7a3-d34d-47ea-b6ae-2fd57640cfd4" />

Whereas doing a similar test with 50 resamples instead does not exhibit the same memory leaks:
<img width="2560" height="1431" alt="Image" src="https://github.com/user-attachments/assets/24ba8859-e68b-46e2-81d7-d342d7f6a173" />

**Steps To Reproduce The Benchmarks Above**

1. Build and serve the editor:
   ```sh
   npx nx run core:build && yarn workspace @penrose/editor build
   cd packages/editor
   npx vite preview --port 4173 --strictPort
   ```
2. Open a diagram in google chrome/chromium that contains labels such as `set-theory-domain/tree-euler`, since the leak only fires whenever TeX labels are collected.
   ```sh
   google-chrome-stable "http://localhost:4173/try/?examples=set-theory-domain%2Ftree-euler"
   ```
3. Open inspect element (usually with Ctrl+Shift+I), go to the Memory tab, and take a Heap snapshot.

<img width="2560" height="1435" alt="Image" src="https://github.com/user-attachments/assets/bccfbe3f-c17a-42b9-8346-ba4ba221440c" />

5. Click **compile** on the penrose editor 50 times, then take another Heap snapshot.
6. Select Snapshot 2, switch the top-left dropdown from **Summary** to **Comparison** (relative to Snapshot 1), and sort by "# Delta".

<img width="2484" height="1389" alt="Image" src="https://github.com/user-attachments/assets/2e05a4b1-d458-4599-90b9-03965f15a88b" />

[...]

PR #1972, opened 2026-07-22
fix(editor): suppress InvalidDiagramIDError errors in the editor when repeatedly recompiling a diagram

Galm007:
# Description

Follow up to PR #1961.

Rapidly compiling a diagram that takes a while to converge shows an `InvalidDiagramIDError` on the editor, which is caused by a race condition where the previous diagram's ID is being used after it has been cleared.

This behavior is expected since the compute loop should be restarted to point to the current sequence, while the old diagram data is cleared. The diagram still converges properly, so I simply suppressed this InvalidDiagramIDError from showing up on the editor.

Go to an example diagram that takes some time to converge (e.g. persistent-homology), and rapidly click on the compile button, giving no time for the diagram to fully converge.

Screenshot_20260722_130357
Screenshot_20260722_125834
Screenshot_20260722_130209

# Checklist


Galm007:
Hi, I created a separate branch for the [PR](https://github.com/penrose/penrose/pull/1976) resolving #1967 so that the commits there wouldn't affect this PR.

wodeni (approved):

